PR #47, opened 2020-05-04
Added IrfanView x64

JudahGabriel:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/47)

### DIFF
--- a/manifests/IrfanSkiljan/IrfanView/4.54.yaml
+++ b/manifests/IrfanSkiljan/IrfanView/4.54.yaml
@@ -11,7 +11,7 @@ InstallerType: exe
 Tags: "images"
 Installers: 
     - Arch: x64
-      Url: https://www.fosshub.com/IrfanView.html?dwl=iview454_x64_setup.exe
+      Url: https://winget.b-cdn.net/iview454_setup.exe
       Sha256: A71A52136A5E3321EAF6E68F9F76AAF340FEE90ECA00CCFC65C32DC3141E9FA5
       Switches: 
         Silent: /silent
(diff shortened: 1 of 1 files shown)


chausner:
I think this currently fails due to multiple issues:

1. ZIP files are currently not supported so you should use the EXE installer instead.

2. The other problem is that there does not seem to exist any direct download link. The one that you specified that looks like a direct link still resolves to a landing page.

3. The validation bot currently does not seem to be compatible with websites that don't provide the file size in the response header (Content-Length), see https://dev.azure.com/ms/winget-pkgs/_build/results?buildId=79280&view=logs&jobId=c8020aa5-4aab-5881-691f-21986e02afd0&j=c8020aa5-4aab-5881-691f-21986e02afd0&t=bade86d6-d2be-5ddd-c1b9-7d8db9c4a60d:
   ````
   Error occurred during installer copy from https://www.irfanview.info/files/iview454_x64.zip.
    The Content-Length for the copy source is not specified.
   ````

superusercode:
there is an official mirror for the installer at http://m.majorgeeks.com/files/details/irfanview_64.html

JudahGabriel:
@superusercode is there a direct link to the installer? I don't see any on that page.

superusercode:
The link is http://files2.majorgeeks.com/6467afde578ca6d96a9a3f62a7f5a68d5dd8ff51/graphics/iview454_x64_setup.exe

JudahGabriel:
Thanks for the link there. Looks like that's not using any dynamic download tokens. I'll try using that link in the package.

JudahGabriel:
Looks like someone found a direct link to the installer and issued a working PR: https://github.com/microsoft/winget-pkgs/pull/1208

Closing this; we should use that PR instead.

PR #179, opened 2020-05-16
Added IntelliJ IDEA Community Edition

denelon:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/179)

### DIFF
--- a/manifests/JetBrains/IntelliJIDEA/Community/2020.1.1.yaml
+++ b/manifests/JetBrains/IntelliJIDEA/Community/2020.1.1.yaml
@@ -1,4 +1,4 @@
-Id: JetBrains.IntelliJIDEACommunity
+Id: JetBrains.IntelliJIDEA.Community
 Version: 2020.1.1
 Name: IntelliJ IDEA Community Edition
 Publisher: JetBrains
--- a/manifests/JetBrains/IntelliJIDEA/Community/2020.1.1.yaml
+++ b/manifests/JetBrains/IntelliJIDEA/Community/2020.1.1.yaml
@@ -1,0 +1,13 @@
+Id: JetBrains.IntelliJIDEACommunity
+Version: 2020.1.1
+Name: IntelliJ IDEA Community Edition
+Publisher: JetBrains
+Homepage: https://www.jetbrains.com/idea/
+License: Apache License 2.0
+LicenseUrl: https://github.com/JetBrains/intellij-community/blob/master/LICENSE.txt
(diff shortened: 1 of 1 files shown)


mimi89999:
In the case of Jetbrains installers only `mode=admin` is required. I tried to convince them to make it default (currently they have no default for this), but they didn't understand the issue:

lokinmodar:
> In the case of Jetbrains installers only `mode=admin` is required. I tried to convince them to make it default (currently they have no default for this), but they didn't understand the issue: https://youtrack.jetbrains.com/issue/IDEA-226645
> 
> Maybe the community could convinve them to have sane defaults and not require a config file for a very common install config.

This is an issue for all their installers. They refuse to acknowledge the problem.

mimi89999:
Two other issues have been opened (https://youtrack.jetbrains.com/issue/IDEA-242025, https://youtrack.jetbrains.com/issue/TBX-4949), but there still isn't any real feedback from them. I think that the problem here is that they don't quite understand what they have to do.

KevinLaMS:
as mentioned above this is blocked.
image

mimi89999:
I opened the PR with the required changes to IntelliJ two months ago, but is still hasn't been accepted. The dev that is following this PR asked about winger popularity in https://github.com/JetBrains/intellij-community/pull/1400#issuecomment-684702491. Maybe the community and users who would like to see JetBrains products in WinGet could speak up in that pull request?

denelon:
It looks like the PR at JetBrains is closed https://github.com/JetBrains/intellij-community/pull/1400#issuecomment-684702491 but there was a comment from @trespasserw indicating the change is in the main branch. The manifest here would need to be updated to the current supported version.

denelon:
@mimi89999 feel free to submit a new PR and I'll close this one, or let me know what changes need to be made.

mimi89999:
@denelon I will submit a PR when the new version with that change will be released. Now we basically only need to wait for the release and it should work as any other NSIS installer, so it will be easy to add.

mimi89999:
@denelon Just opened https://github.com/microsoft/winget-pkgs/pull/5212

mimi89999:
Also opened https://github.com/microsoft/winget-pkgs/pull/5213 for PyCharm

denelon:
JetBrains.IntelliJIDEA.Community was added in another PR.

PR #190, opened 2020-05-17
Docker

KevinLaMS:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/190)


ghost (approved):
Validation has completed

StefanScherer:
@KevinLaMS Thank you very much. I was surprised that a package for Docker Desktop already exists! 🎉 
But I see a problem using the dynamic link to the latest stable. I see a sha256 checksum. If `winget` also checks the downloads (which is good), it will break as soon as Docker releases a new stable version.

The Chocolatey package `docker-desktop` uses the deep link to the corresponding build to have a static link that wouldn't change after a release. https://github.com/riezebosch/BoxstarterPackages/blob/master/docker-desktop/tools/chocolateyinstall.ps1#L5

PR #232, opened 2020-05-17
Music

KevinLaMS:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/232)


ghost (approved):
Validation has completed

hjc:
@KevinLaMS - tried to install this through `winget` today. It downloads the `.exe` just fine (`SamuelAttard.GooglePlayMusicDesktopPlayer.4.7.1.exe`) but then it fails with "Failed to extract installer", however `winget` reports it as a success. Went to the URL in the manifest, downloaded the `.exe` myself and ran it and it worked fine, so looks like some weird `winget`tery. I ran this through Windows Terminal in Admin and not in Admin.

If this isn't the right place for a bug report for individual packages, let me know and I'll file elsewhere. Wasn't sure if this should be an issue here or in the source GitHub repo.

PR #256, opened 2020-05-18
Add kdenlive 20.04.1

### LINKED ISSUE #259: Support HTTP redirects better
We have encountered a few different kinds of issues with the various HTTP redirects used to identify installers.  We need better error messages for these validation scenarios, and a set of capabilities to correctly handle them when validating manifests and checking installers.

jazzdelightsme:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/256)


SoftCreatR:
@jazzdelightsme Until #259 has been solved, you may use the target URL which is

`http://mirror.cc.columbia.edu/pub/software/kde-applicationdata/kdenlive/release/kdenlive-20.04.1.exe`

PR #307, opened 2020-05-19
Add manifest for WinSCP v5.17.5

OkGoDoIt:
Adds [WinSCP](https://winscp.net) (v5.17.5)

WinSCP is an open source free SFTP client, FTP client, WebDAV client, S3 client and SCP client for Windows. Its main function is file transfer between a local and a remote computer. Beyond this, WinSCP offers scripting and basic file manager functionality. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/307)


OkGoDoIt:
To whoever looks at this from @microsoft, it would be worth updating the instructions.  Some issues I ran into:

* When running the manifest creation tool, it apparently set a weird encoding which caused Git to think my yaml file was binary.  I had to manually convert to UTF-8 via an editor.
* It seems the `silent` and `silentwithprogress` flags are required to get the validation to pass, alough I had to figure that out for myself because it's not documented or included in the generator tool.
* It seems that the installer pulled from the CDN was not downloading properly.  I'm not sure if this is an issue with not following redirects or maybe the host blocking the Azure pipeline IP as an anti-spam effort.  Some additional insight would be great for debugging.

I eventually got it all working but as you can see from my commit history it took several tries.  Hope this feedback can help others!

ghost (approved):
Validation has completed

PR #311, opened 2020-05-19
Fix VLC metadata

jbkempf:
Those modifications fix the version of VLC (64b vs 32b) and fixes some metadata errors, notably for the URL and MinOSVersions.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/311)


jbkempf:
Something is wrong here:
It refuses the new (fixed) URL because of 302 redirect detected. But the previous version was at the same location, with 302...

What is the proper fix?

Issue #277, opened 2020-05-19
VS Code's appmoniker is "vscode", not "code"?


Tyriar:
The CLI for VS Code is `code`/`code-insiders` and as a result the deb/snap/rpm packages are named `code` and `code-insiders`. The winget package should do it as well. I'd like to make this change if it sounds good.

Tyriar:
cc @KevinLaMS

NatoBoram:
I think `code` is not a good name because it's prone to conflicts. Non-Microsoft projects referencing it are using the moniker `vscode`. For example, [Scoop](https://github.com/lukesampson/scoop-extras/blob/master/bucket/vscode.json).

theterminator3000:
I have to agree with @NatoBoram on this. It's also known as vscode on chocolatey.

Tyriar:
It's only known as vscode on chocolatey because it's not an official package. If it were it would be consistent with [the rest of the packages](https://github.com/microsoft/winget-pkgs/issues/278#issuecomment-630952642) and the CLI.

NatoBoram:
It's still inherently a bad name for being so prone to conflict. For example, even [`code`](https://github.com/elementary/code) on Elementary OS changed their moniker to `io.elementary.code`. VSCode's repo is also [`vscode`](https://github.com/microsoft/vscode), obliterating your only argument.

Tyriar:
I don't know if "vscode is used somewhere" obliterates my argument. My argument is to align it with the cli for ux/discoverability reasons:

```
winget install code
code

apt install code
code
```

I brought up when I joined the team these potential conflicts with a generic name like `code` and maybe changing the cli to `vscode`, but the decision was to keep the cli as code, so packages have inherited that.

felipecassiors:
How about to support 2 monikers?

Tyriar:
@felipecassiors that's the preference

denelon:
The value "code" makes sense in the "Commands" key, but we've seen more preference for "vscode" as the value for the "AppMoniker" key.

Tyriar:
@denelon the docs say

> `# List of commands or aliases the user would use to run the package.`

Does `winget install code` work/redirect to it?

felipecassiors:
No, because of multiple matches.

Maybe we should invest in `winget install --comand` instead, like `--moniker`?

denelon:
[...]

PR #327, opened 2020-05-19
Add vscode-insiders

### LINKED ISSUE #320: [Request] Add VSCode Insiders

NatoBoram:
Closes #320 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/327)

### DIFF
--- a/manifests/Microsoft/VisualStudioCodeInsiders/1.46.0.yaml
+++ b/manifests/Microsoft/VisualStudioCodeInsiders/1.46.0.yaml
@@ -1,0 +1,13 @@
+Id: Microsoft.VisualStudioCodeInsiders
+Version: 1.46.0
+Name: Visual Studio Code Insiders
+Publisher: Microsoft Corporation
+License: Copyright (c) Microsoft Corporation
+AppMoniker: vscode-insiders
+Description: Visual Studio Code is a code editor redefined and optimized for building and debugging modern web and cloud applications. Visual Studio Code is free and available on your favorite platform - Linux, macOS, and Windows.
+Homepage: https://code.visualstudio.com/insiders/
+Installers:
+  - Arch: x64
+    Url: https://go.microsoft.com/fwlink/?Linkid=852155
+    Sha256: 434A9433B72692C0C579A33B1C20EB72BD05ED3BB533A34A76C9702482FA9A3E
+    InstallerType: inno


manifests/Microsoft/VisualStudioCodeInsiders/1.46.0.yaml
```diff
@@ -0,0 +1,13 @@
+Id: Microsoft.VisualStudioCodeInsiders
+Version: 1.46.0
```
Tyriar, line 2: This is technically not correct as the version is `x.y.z-insider`, but having the version number would require us to change it once a month. I was wondering in https://github.com/microsoft/winget-pkgs/issues/278 if we can use something like "Latest" or "Insider" here instead.

manifests/Microsoft/VisualStudioCodeInsiders/1.46.0.yaml
```diff
@@ -0,0 +1,13 @@
...
+Name: Visual Studio Code Insiders
+Publisher: Microsoft Corporation
+License: Copyright (c) Microsoft Corporation
+AppMoniker: vscode-insiders
```
Tyriar, line 6: I want to get https://github.com/microsoft/winget-pkgs/issues/277 sorted out before we do insiders, ideally we would use `code` and `code-insiders` to match the CLI and have `vscode` as an alias/search term.

manifests/Microsoft/VisualStudioCodeInsiders/1.46.0.yaml
```diff
@@ -0,0 +1,13 @@
...
+Description: Visual Studio Code is a code editor redefined and optimized for building and debugging modern web and cloud applications. Visual Studio Code is free and available on your favorite platform - Linux, macOS, and Windows.
+Homepage: https://code.visualstudio.com/insiders/
+Installers:
+  - Arch: x64
+    Url: https://go.microsoft.com/fwlink/?Linkid=852155
```
Tyriar, line 11: This would probably be better: https://update.code.visualstudio.com/latest/win32-x64-user/insider, the hash won't work though as it changes every day. While certainly possible, it would be preferable not to need a PR (or automated commit) everyday to do this. I also saw that sources (`winget source`) are supported, are all Microsoft packages meant to be maintained in this repo? It would be easier to automate if we owned the repo.

Tyriar (commented):
I started a conversation about how to do this in https://github.com/microsoft/winget-pkgs/issues/278. The comments are for @KevinLaMS

manifests/Microsoft/VisualStudioCodeInsiders/1.46.0.yaml
NatoBoram: Thanks, I added [my opinion](https://github.com/microsoft/winget-pkgs/issues/277#issuecomment-631440881) on it!

manifests/Microsoft/VisualStudioCodeInsiders/1.46.0.yaml
NatoBoram: Scoop also had to deal with this in https://github.com/lukesampson/scoop/issues/181, and they [added simplistic nightly support](https://github.com/lukesampson/scoop/pull/314). Might be worth to see that they did.

KevinLaMS:
Duplicate of #2056
https://github.com/microsoft/winget-pkgs/pull/2056
Would you be OK closing now that Vscode is owning updating this?

NatoBoram:
Sure!

PR #337, opened 2020-05-19
Add Epic Games Launcher 10.15.2

boarnoah:
For the Epic Games Store as well as Unreal Engine dev (its a dual purpose launcher).

I will note that URL's sub-domain doesn't look stable "https://launcher-public-service-prod06.ol.epicgames.com".

I checked under VPN from other regions and other settings and it seemed stable (same subdomain). It might not be the case over time, therefore this is something that might need addressing (from Epic's side or by updating manifest here as changes occur).

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/337)

### DIFF
--- a/manifests/EpicGames/EpicGamesLauncher/10.15.2.Yaml
+++ b/manifests/EpicGames/EpicGamesLauncher/10.15.2.Yaml
@@ -1,0 +1,15 @@
+Id: EpicGames.EpicGamesLauncher
+Version: 10.15.2
+Name: Epic Games Launcher
+Publisher: Epic Games Inc.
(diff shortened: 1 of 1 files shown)


boarnoah:
That was interesting,  "winget validate" was sure my manifest was fine, but the pipeline flagged an error with the ID ("it has to be an exact match folder name wise").

ghost (approved):
Validation has completed

PR #342, opened 2020-05-19
Use permalink for Yarn download

Daniel15:
`/latest.msi` redirects to the latest version of Yarn. This means that if a new version is released, it'll start downloading the new version rather than the expected version (1.22.4), and the hash won't match.

This PR updates it to use the permalink to the installer for 1.22.4 specifically.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/342)


denelon:
Great fix!

ghost (approved):
Validation has completed

PR #346, opened 2020-05-19
Add SumatraPDF v3.2

drproteus:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/346)


denelon:
@drproteus did you install Sumatra on your local machine with the manifest. I got an error that looks like a "remove:" failed. This may be the difference in user mode vs. admin mode. It looks like it did install, but I'd like to make sure we're providing the best experience we can for users.

drproteus:
@denelon I must be too used to `apt` land because I immediately thought I was doing something wrong when I got the error in a user terminal, so I ran it in an elevated prompt. It seems like `winget` is trying to delete the temp installer before its done being executed. Not sure if its at all related to https://github.com/microsoft/winget-cli/issues/189.

KevinLaMS:
This is blocked on Issue https://github.com/microsoft/winget-cli/issues/189

ghost (approved):
Validation has completed

GitHubRulesOK:
The command line is incorrect the fileNAME-install.exe is the trigger and the silent option was /s (small) it will likely be -s in the next release but there are other considerations such as allow the multi user administrator to specify /d (destination will become -d) the current default is LocalAppdata

PR #358, opened 2020-05-19
Create 1.18.0.yaml

yualan:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/358)


denelon:
@yualan Can you check for switches to support unattended / no UI?

yualan:
We have the same installer as VS Code, so if they have support for unattended UI, we can follow same pattern. If they do not, then we can proceed with the same experience as VS Code.

denelon:
Thanks for the clarification. Since the installer did complete without me having to engage with any other UX, I'll go ahead and mark this completed.

ghost (approved):
Validation has completed

PR #361, opened 2020-05-20
Add Sublime Merge build 1119

boarnoah:
- Sublime Text (by another user Sublime company's manifest) doesn't have any installs witches listed. I can confirm the same switches here are present there as well, didn't want to include them in this PR in case it breaks the bot's CI process.
- Both this and sublime text requires winget to be launched as admin, else the installer fails with an "access denied" dialog

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/361)


ghost (approved):
Validation has completed

PR #366, opened 2020-05-20
Add Tokei 11.2.0

XAMPPRocky:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/366)

### DIFF
--- a/manifests/XAMPPRocky/tokei/11.2.0.yaml
+++ b/manifests/XAMPPRocky/tokei/11.2.0.yaml
@@ -10,5 +10,5 @@ Homepage: https://github.com/XAMPPRocky/tokei
 Installers:
   - Arch: x64
     Url: https://github.com/XAMPPRocky/tokei/releases/download/v11.2.0/tokei-x86_64-pc-windows-msvc.exe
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
It has been submitted, so now just waiting for a response.

XAMPPRocky:
@KevinLaMS It has been confirmed as a false positive.

Screenshot 2020-05-22 at 14 08 48

KevinLaMS:
The binary looks good, but it does not seem to install.  It appears the binary is the CLI not the installer, so I do not see how a user can access it.

XAMPPRocky:
@KevinLaMS Ah okay, I must have misread the instructions as I thought exe was supposed to point to the application. Do you have to provide an installer or is there a way to provide binaries directly like in other package managers?

denelon:
@XAMPPRocky it looks like you are requesting support for https://github.com/microsoft/winget-cli/issues/182.

XAMPPRocky:
Thank you for your response! Yes I believe so.

PR #469, opened 2020-05-20
Add Logitech Options

xdrl1:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/469)

### DIFF
--- a/manifests/Logitech/Options/8.20.329.yaml
+++ b/manifests/Logitech/Options/8.20.329.yaml
@@ -1,0 +1,11 @@
+Id: Logitech.Options
+Publisher: Logitech
+Name: Logitech Options
+Version: 8.20.329
+License: Copyright (c) Logitech 2020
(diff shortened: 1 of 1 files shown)


mauriceluo69:
@KevinLaMS 

Hello, thank you for your reply!

I've tried several silent parameters like /S /Q, but none of these work.

I've search for some information about silent installation of Logitech Options, but it seems that this software cannot be installed silently as the info shown in links below:

[1] [chocolatey](https://chocolatey.org/packages/logitech-options)
[2] https://www.itninja.com/question/does-any-one-know-about-silent-installation-parameters-for-logitech-options

Only previous version seems able to install silently: 

https://silent-install.net/software/logitech/options/6.50.60

Therefore, I don't know what I can do next to contribute to this PR.

Thanks again for your patience

PR #478, opened 2020-05-20
Use an in-memory stream in YamlCreate.ps1 for hash computation

chausner:
This changes [Tools/YamlCreate.ps1](Tools/YamlCreate.ps1) to not download and store the installer on disk for computing the hash. Instead, the hash is computed directly on the download stream.

Before this change, downloaded installers would stay in the temporary directory and not get deleted by the script. Also, you could not run the script multiple times in parallel as the temporary file name was hardcoded and there would be a conflict.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/478)


KevinLaMS (dismissed):
Nice.  Thank you.

KevinLaMS (approved):

PR #485, opened 2020-05-20
Remove bogus special character

AMDmi3:
There's bogus special character in GitExtensions package which breaks yaml parsing:
`yaml.reader.ReaderError: unacceptable character #x0008: special characters are not allowed`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/485)


ghost (approved):
Validation has completed

PR #501, opened 2020-05-20
update spotify version

pivendren:
I am interested to know if this is the suggested approach for updates. 

The URL for this exe is not versioned which got me asking a few questions because before this PR anyone who uses this will get  v1.1.33.569 instead of v1.1.30.658

1. Do I just bump the version in the manifest and leave the name of the manifest file mismatched? 
2. Rename the file and update the version?
3. Add a new file with the new version?** 

** This one is more confusing because irrespective of the version the user selects they will always get the latest version of the exe.

I opted for 2 in this PR.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/501)

### DIFF
--- a/manifests/Spotify/Spotify/1.1.33.569.yaml
+++ b/manifests/Spotify/Spotify/1.1.33.569.yaml
@@ -11,7 +11,7 @@ InstallerType: Exe
 Installers:
   - Arch: x86
     Url: https://download.spotify.com/SpotifyFullSetup.exe
-    Sha256: 77c953863c5882ba78796c192044cbea3841dbdab02a4b2b9237793991e17285
+    Sha256: 82c591f3c046a66883d041ac26514ab0a4ebfe29d84247331b5bac950ca091b6
     Switches:
       Silent: /Silent
       SilentWithProgress: /Silent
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
You should submit a new manifest for a new version so users can install a specific version.

pivendren:
> 
> 
> pivendren The package manager bot determined there was an issue with one of the installers listed in the url field, and cannot continue.
> 
> Here are some common causes for this failure you should consider:
> 
>     1. The Sha256 HASH in the manifest does not match the HASH of the installer. Run winget hash  to generate the hash.
> 
>     2. The URL is not valid. Make sure the URL to the installer is publicly available and valid.
> 
>     3. The installer has been identified as malware. If the installer is detected as malware, you can submit the installer to the defender team for analysis as a potential false positive.


Yeah, pretty much the same situation we discussed in #511 Waiting on a suggested approach from the reviewer

pivendren:
Not sure what the Manifest Content Validation error is, AZDO isn't giving me much in the logs. @KevinLaMS can you see anything more than me?

PR #507, opened 2020-05-20
add citrix workspace app

pivendren:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/507)


pivendren:
I see now that the download link has an expiring token at the end. I don't have a workaround for now so I am going to drop this PR. Thanks for the time @KevinLaMS

PR #511, opened 2020-05-20
Update GitKraken to v7.0.0

pivendren:
Update to latest version 7.0.0 with new exe hash.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/511)

### DIFF
--- a/manifests/Axosoft/GitKraken/7.0.0.yaml
+++ b/manifests/Axosoft/GitKraken/7.0.0.yaml
@@ -1,6 +1,6 @@
 Id: Axosoft.GitKraken
 Name: GitKraken
-Version: 6.5.4
+Version: 7.0.0
 Publisher: Axosoft
 Homepage: https://www.gitkraken.com/
 Description: Git GUI Client.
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
You should submit a new manifest for a new version so users can install a specific version.

pivendren:
> 
> 
> You should submit a new manifest for a new version so users can install a specific version.

@VIKTORVAV99 that would be cool but all the download links point to the exe of the latest version. Is there a suggested work around for this?

VIKTORVAV99:
That's a tricky one, I'm worried that updating the manifest could break scripts that pin a specific version.

This probably isn't the case yet since winget is pretty new but I think submitting a new manifest might be the best way even if they point to the same end version.

pivendren:
Yeah, it is a tricky one. I have another PR(#501) with the same situation where I asked these questions with possible solutions. Waiting on a reply from the reviewers for the best approach.

VIKTORVAV99:
I just realized I commented the same thing on that one... Hehe...

But yeah let's wait for the reviewers.

ghost (approved):
Validation has completed

PR #524, opened 2020-05-20
Added DBeaver

LightDestory:
Free universal database tool!

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/524)


LightDestory:
How can I implement the silent mode correctly?
Right now the manifest just download and open up the install and the user must manually follow the setup wizard.
Which is the correct type of installation: all user or current user? (winget runs with administrator privileges? If not how can we use the "all user" install type?)

KevinLaMS:
Looking at help, it appears you could pass /currrentuser /S for silent current user.
Eventually the package manager will add support for the user to choose.

LightDestory:
@KevinLaMS I have updated the manifest, now it should do a proper silent installaiton.

ghost (approved):
Validation has completed

PR #530, opened 2020-05-21
Add Tencent QQ

hifocus:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/530)


manifests/Tencent/QQ/0.0.0.0.yaml
```diff
@@ -0,0 +1,19 @@
+Id: Tencent.QQ
+Name: QQ
+Version: 0.0.0.0
```
kidonng, line 3: I don't think it's a good idea to leave out the version.

```diff
- Version: 0.0.0.0
+ Version: 9.3.2.26869
```

kidonng (changes requested):

manifests/Tencent/QQ/0.0.0.0.yaml
```diff
@@ -0,0 +1,19 @@
...
+Publisher: hifocus
```
imba-tjd, line 4:
```suggestion
Publisher: Tencent Corporation
```

imba-tjd: see `https://github.com/microsoft/winget-pkgs/pull/564`

hifocus: > 

See https://apps.apple.com/us/app/qq/id444934666

I would think it's a better idea to use `Tencent Technology (Shenzhen) Company Limited`.

manifests/Tencent/QQ/0.0.0.0.yaml
hifocus: image

I thought 'universal installers' shouldn't specify its version, as in https://github.com/microsoft/winget-pkgs/pull/60; however the original author did changed the version to a proper value so I did the same just now  https://github.com/microsoft/winget-pkgs/pull/530/commits/124388d37d36dd2768bee899dd15b921a0425971

hifocus:
The checks just keep being in the state of 'in progress' forever :(

hifocus:
> The checks just keep being in the state of 'in progress' forever :(

Sorry for the disruption, but can you please have a look at my situation to see if  anywhere in either my configuration or the Azure Pipeline goes wrong? Thanks @KevinLaMS

denelon:
@hifocus let's see if we can get that CI Bot to cooperate this time.

hifocus:
> @hifocus let's see if we can get that CI Bot to cooperate this time.

It seems that it got stuck again... Shouldn't there be some improvements on the auto flow of CI?

> started 9h 52m 58s ago

One reason that I didn't add the `SilentWithProgress` or similar variable is that a lot of software that I am adding is from China, they often come up with a bunch of other irrelevant software for promotion purpose. There being said, the user experience won't be good if auto installation is allowed.

I also have a concern despite some software (like the one in this PR, QQ) provides a 'always newest' download link, which means whenever a new version of the software is released, their staff is going to update the installer file on their server but keep the same name. However this could lead to a faulty sha256 of the installer file, also the user cannot get the latest update if on GitHub another PR towards the latest version is not made.

I know that the sha256 is for integrity to avoid damaged package to be installed, but can't we prompt a warning on the commandline saying that 'this software uses a always-latest-link, so the integrity of the download will not be verify, use it at your own risk'?

But eventually, thanks for the help.

KevinLaMS:
Thanks for the feedback.   I agree we do not want to encourage a bunch of irrelevant software getting added.  
We are aware of the issue with the newest URLs and are investigating the possibility of different solution in the future.

hifocus:
> @hifocus we're working with an internal team to see if we can determine why our automation is failing to provide a more meaningful status. I'm hoping we get a response in the next day or two so we can make progress getting this PR merged.

Sure, many thanks for the hard work!

PR #535, opened 2020-05-21
Add Anvil Studio 2020.12.03

chausner:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/535)


KevinLaMS:
Installer failed with exit code 1000

chausner:
> Installer failed with exit code 1000

Yes, [this is to be expected](http://underpop.online.fr/w/winrar/helpsfxsetupcode.htm). See also the comment I added in the manifest file. The installer in this case is wrapped in a WinRAR self-extracting archive. After extracting, RARSFX runs the installer and returns with the installer exit code + 1000. That means, if the installer succeeded (exit code 0), RARSFX will return exit code 1000.

We need a way to specify a success exit code in the manifest file, as not all applications on Windows adhere to 0 = success.

PR #537, opened 2020-05-21
Fix .NET display strings

terrajobst:
I've changed some of the strings to match the official product names.

@MichaelSimons @richlander:

Do we want to change the package ids? I understand that this is a breaking change but I guess now is the time to do that if we aren't happy with them. These are the current ones:

Id                       | Name
-------------------------|----------------
Microsoft.dotnet         | .NET Core
Microsoft.dotNetFramework| .NET Framework

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/537)

### DIFF
--- a/manifests/Microsoft/dotnet/3.1.202.yaml
+++ b/manifests/Microsoft/dotnet/3.1.202.yaml
@@ -8,7 +8,7 @@ LicenseUrl: https://github.com/dotnet/core/blob/master/LICENSE.TXT
 MinOSVersion: 6.1.7601
 Description: .NET Core is an open-source, general-purpose development platform. You can create .NET Core apps for Windows, macOS, and Linux for x64, x86, ARM32, and ARM64 processors using multiple programming languages. Frameworks and APIs are provided for cloud, IoT, client UI, and machine learning.
 Homepage: https://docs.microsoft.com/dotnet/core/
-Tags: .NET, .NET Core, dotnet, C#, csharp, F#, fsharp, VB, Visual Basic
+Tags: .NET, .NET Core, dotnet, net, C#, csharp, F#, fsharp, VB, Visual Basic
 Commands: dotnet
 InstallerType: exe
 Installers:


richlander:
LGTM

MichaelSimons (dismissed):
LGTM

MichaelSimons:
@terrajobst, In order to get this PR green, I think you need to split is apart.  The repo readme states that only one manifest can be contained in a PR.

MichaelSimons:
> Do we want to change the package ids? I understand that this is a breaking change but I guess now is the time to do that if we aren't happy with them.

I don't care for the inconsistent casing of dotnet vs dotNet

I chose Microsoft.dotnet vs Microsoft.dotnetCore so that it will be long term applicable when 5.0 is released and we drop “core” from the product name.  I thought it would be acceptable to change the name once .NET 5.0 is released from .NET Core to .NET.  An alternative option would be to have a "core" package for <= 3.1 and a non-core package for >= 5.0.  

Outside of the casing, I have no objections to `Microsoft.dotNetFramework`.

terrajobst:
> @terrajobst, In order to get this PR green, I think you need to split is apart. The repo readme states that only one manifest can be contained in a PR.

Oopsy... probably should have read that. Submitted #551 for .NET Framework and kept this to .NET Core.

terrajobst:
> I chose Microsoft.dotnet vs Microsoft.dotnetCore so that it will be long term applicable when 5.0 is released and we drop “core” from the product name. I thought it would be acceptable to change the name once .NET 5.0 is released from .NET Core to .NET. An alternative option would be to have a "core" package for <= 3.1 and a non-core package for >= 5.0.

I figured and I'm OK with that reasoning. I just raised it because it seems we should make this decision deliberately.

> Outside of the casing, I have no objections to `Microsoft.dotNetFramework`.

Should I submit a PR to change the casing? I'm afraid this will break stuff.

MichaelSimons:
>Should I submit a PR to change the casing? I'm afraid this will break stuff.
Yes please, I think now is the time to change it.

MichaelSimons (approved):

ghost (approved):
Validation has completed

PR #279, opened 2020-05-19
Add Tencent Tim

YadominJinta:
Add Tencent Tim.

It's a popular IM in China.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/279)

PR #547, opened 2020-05-21
Add Tencent TIM

zhuohoudeputao:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/547)


manifests/Tencent/TIM/3.3.0.21315.yaml
```diff
@@ -0,0 +1,12 @@
+Id: Tencent.TIM
+Name: Tencent TIM
```
hifocus, line 2: I would suggest to simply use `TIM` here

imba-tjd:
See discussions in https://github.com/microsoft/winget-pkgs/pull/530#discussion_r428445262

hifocus (changes requested):

hifocus:
> See discussions in [#530 (comment)](https://github.com/microsoft/winget-pkgs/pull/530#discussion_r428445262)

I think using the installer's digital signature could be a better idea, though we do have the same outcome:

image

imba-tjd:
https://play.google.com/store/apps/details?id=com.tencent.mobileqq uses `Tencent Technology (Shenzhen) Company Ltd.`
I don't think it needs to be so rigorous. Anyway I will correspond with you.
As for the blank line, I prefer having. It's *nix style.

kidonng:
#279?

imba-tjd:
:joy: Add @YadominJinta in discussion. I don't know what to do. You and `zhuohoudeputao` decide.

YadominJinta:
I don't think it needs to be so rigorous. 🤣 
But I will follow you suggestions

imba-tjd:
One thing is that both YadominJinta and zhuohoudeputao are trying to add TIM. In theory we should keep the early one.

YadominJinta:
But my pull stuck in CI, I think it's maybe QQ or Tim's problem

hifocus:
> But my pull stuck in CI, I think it's maybe QQ or Tim's problem

Same as mine. I tried to add QQ.

zhuohoudeputao:
Does it seem that it will be stuck here forever?

PR #571, opened 2020-05-21
Add Krita. It's a digital painting application

shengurun:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/571)


shengurun:
This URL will be redirect Official KDE Mirrors, so an error has occurred:
"A redirected response (HTTP status code 302) from the copy source is not supported."
What should I do?

PR #583, opened 2020-05-21
Added /NORUN parameter to ShareX manifest

Jaex:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/583)


ghost (approved):
Validation has completed

superusercode:
https://github.com/microsoft/winget-cli/blob/e997f3e394e66f2389f4f33f36c58b1c2768e975/src/AppInstallerRepositoryCore/Manifest/ManifestInstaller.cpp#L137-L140 @Jaex /VERYSILENT and /SILENT are both included by default, but I don't think there are any ramifications to providing the flag twice.

Jaex:
> https://github.com/microsoft/winget-cli/blob/e997f3e394e66f2389f4f33f36c58b1c2768e975/src/AppInstallerRepositoryCore/Manifest/ManifestInstaller.cpp#L137-L140 @Jaex /VERYSILENT and /SILENT are both included by default for the inno installertype, but I don't think there are any ramifications to providing the flag twice.

Those default flags are not included when you override switch to add new flag.
But still I tested it now without add silent flag and indeed I was right, installer not run as silent.

superusercode:
@Jaex is the Custom switch field what you need then? seen in the spec here: https://github.com/microsoft/winget-cli/blob/master/doc/ManifestSpecv0.1.md

PR #308, opened 2020-05-19
Move CCleaner to Piriform and update yaml

NathanBnm:
Piriform Software is not appropriate. It should be Piriform only so I made the changes. I updated CCleaner details too.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/308)

PR #594, opened 2020-05-21
Move and update CCleaner

NathanBnm:
This PR overwrites #308 in which I had an issue with my GitHub fork.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/594)


NathanBnm:
@denelon @KevinLaMS checks are failing only because it appears to have two manifests files which is normal because I moved it from one folder to another so you have the deleted one and the moved one. Could you please review this manually?

superusercode:
@NathanBnm git does a weird thing where if the first commit moves the file and the second commit actually edits the file it will recognize that the file has moved instead of as two different files.

NathanBnm:
I thought it could be force validated by the reviewer team. But I agree je with you.

msftrubengu:
@NathanBnm was this a git rm + git add? Or was it a git mv? If the file is marked as renamed we should be able to handle it.

NathanBnm:
I just moved it and edited it in my editor

msftrubengu:
I believe that would do a git rm + git add. The easiest solution would be to revert your commit and do a `git add -A .` Then, do git status to validate the file is marked as renamed. 

Also, note that your submission will fail because the new file was `Version: 1.53.1087`. So you would need to rename the file as 1.53.1087.yaml

NathanBnm:
I'll do that asap thanks 😉

PR #647, opened 2020-05-21
Add kodi

louislac:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/647)


superusercode:
wouldn't the nullsoft installer type be better for this listing?

louislac:
> 
> 
> wouldn't the nullsoft installer type be better for this listing?

What do you mean about nullsoft installer ?

superusercode:
@louislac  If you change `InstallerType: exe` to `InstallerType: nullsoft` it will do the silencing for you and you can remove the flags. There are already a number of files in this repo that use it.

For some reason this type is only exposed via the yaml creator tool in this repo and not in the GUI tool or the docs.

superusercode:
All the possible types are listed https://github.com/microsoft/winget-cli/blob/master/doc/ManifestSpecv0.1.md

I would have done this as a suggestion to your file but your file is currently in a binary format.

louislac:
I didn't know about nullsoft, thank you for the tip.

I made a few test, isn't it better no keep the switches to allow a user through winget whether he want silent or interactive ?

superusercode:
@louislac switches seem to not be encouraged for any type not an exe. I assume they eventually want to show the install info in terminal (like msix apps). There are no explicit opinions on this in the docs but I would honestly just future proof it with the nullsoft type.

louislac:
@superusercode I made the change, Nullsoft seems to fit better with kodi since there is nothing to custom in the installation. Thank you for the advice.
But I feel a bit sceptical, Nullsoft don't allow personalization with winget's flags ?

superusercode:
@louislac Not at the moment, but since they've made a type specifically for nullsoft I feel like it is definitely possible for them to do so in the near future.  Not sure if a tracking issue would be set here or in [winget-cli](https://github.com/microsoft/winget-cli)

PR #655, opened 2020-05-21
Added Liberica JDK 8 (x64).

den1ska:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/655)

### DIFF
--- a/manifests/BellSoft/LibericaJDK/8.0.252.9.yaml
+++ b/manifests/BellSoft/LibericaJDK/8.0.252.9.yaml
@@ -12,4 +12,3 @@ Installers:
       Url: https://download.bell-sw.com/java/8u252+9/bellsoft-jdk8u252+9-windows-amd64.msi
       InstallerType: Msi
       Sha256: 034758E2536F77F92A5D31BBCD47E5E7CFFB40678462DE0418FD544A415D70B4
-ManifestVersion: 1.0.0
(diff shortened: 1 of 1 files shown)


manifests/BellSoft/LibericaJDK/8.0.252.9.yaml
```diff
@@ -0,0 +1,14 @@
+Id: BellSoft.LibericaJDK
+Name: Liberica JDK 8
+Version: 8.0.252.9
+Publisher: BellSoft
+Homepage: https://bell-sw.com
+License: GPLv2 with exceptions
```
jamierocks, line 6: This should be `GPL 2 with Classpath Exception`.

jamierocks (changes requested):

manifests/BellSoft/LibericaJDK/8.0.252.9.yaml
den1ska: Thank you. Fixed.

jamierocks:
When this installer is run, does it do so silently?

If not, the installer probably provides switches that allow such behaviour.

den1ska:
> When this installer is run, does it do so silently?
> 
> If not, the installer probably provides switches that allow such behaviour.

I've just added switches to run the installer in the non-interactive mode.

ghost (approved):
Validation has completed

PR #665, opened 2020-05-21
Add python to path

scaryrawr:
- Add custom switch to pass in [PrependPath switch](https://docs.python.org/3/using/windows.html#installing-without-ui)
> Adds install and Scripts directories to PATH and `.py` to PATHEXT

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/665)

### DIFF
--- a/manifests/Python/Python/3.8.3.yaml
+++ b/manifests/Python/Python/3.8.3.yaml
@@ -15,3 +15,4 @@ Installers:
       Switches:
         Silent: /Quiet
         SilentWithProgress: /Passive
+        Custom: PrependPath=1
--- a/manifests/Python/Python/3.7.7.yaml
+++ b/manifests/Python/Python/3.7.7.yaml
@@ -15,3 +15,4 @@ Installers:
       Switches:
         Silent: /Quiet
         SilentWithProgress: /Passive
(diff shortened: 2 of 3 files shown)


Reset816 (approved):

manifests/Python/Python/3.8.3.yaml
```diff
@@ -15,3 +15,4 @@ Installers:
       Switches:
         Silent: /Quiet
         SilentWithProgress: /Passive
+        Custom: PrependPath=1
```
Reset12138, line 18:
```suggestion
        Custom: PrependPath = 1
```
will it looked better?

Reset816 (changes requested):

manifests/Python/Python/3.8.3.yaml
scaryrawr: If you add spaces, python does not get added to the path.

Reset12138: got it

ghost (approved):
Validation has completed

PR #668, opened 2020-05-21
Submitting Sysinternals Process Monitor 3.53.yaml

jeffclevenger:
first commit for applying this. please let me know what i did wrong

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/668)


chausner:
This will run Process Monitor once, it will not install it. I think this would confuse users and we should wait until ZIP installs are properly supported by winget.

jeffclevenger:
thanks. I will keep the manifest for my team for the time being and keep on the lookout for that.

PR #679, opened 2020-05-21
Added Plexamp

owenmurr:
Added Plexamp

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/679)

### DIFF
--- a/manifests/Plex/Plexamp/3.0.3.yaml
+++ b/manifests/Plex/Plexamp/3.0.3.yaml
@@ -1,0 +1,18 @@
+Id: plex.plexamp
+Version: 3.0.3
+Name: Plexamp
+Publisher: Plex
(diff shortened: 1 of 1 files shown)


KevinLaMS:
I could not get this app to run.  I double clicked it, but no app showed up.  When run from command line i get:  C:\Users\kevin>
 Started Media Service: false
(node:7504) Electron: Loading non context-aware native modules in the renderer process is deprecated and will stop working at some point in the future, please see https://github.com/electron/electron/issues/18397 for more information
(node:7504) Electron: Loading non context-aware native modules in the renderer process is deprecated and will stop working at some point in the future, please see https://github.com/electron/electron/issues/18397 for more information

Issue #659, opened 2020-05-21
Amazon.AWSCLI Install has verification failed


ronorato:
I was attempting to install Amazon.AWSCLI today using winget and received a warning that the installer has failed verification.  Due to this mismatch I canceled the installer to prevent a possible security issue on my system. 
The following are the commands that I executed when installing:

    winget install -h -e Amazon.AWSCLI
    
    Found AWS Command Line Interface v2 [Amazon.AWSCLI]
    This application is licensed to you by its owner.
    Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
    Downloading https://awscli.amazonaws.com/AWSCLIV2.msi

    Installer hash verification failed. Continue? (Y|N)
    Canceled; Installer hash mismatch

System information:
- Hardware: SurfaceBook 2 with latest SurfacBook 2 patches
- OS: Windows 10 Pro, 2004, Build: 19041.264
- Windows Insider Program enabled for slow
- Developer settings
  - Developer Mode is on
  - Device Portal off
  - Device discovery off
  - PowerShell policy has been set to run all scripts

LukasNespor:
It seems that link https://awscli.amazonaws.com/AWSCLIV2.msi points to latest 2.x.x version so the hash does not match.

ronorato:
I did end up just installing the CLI via the Amazon site.  My goal is to create a PowerShell script to help rebuild my laptop after I complete an engagement.

jamierocks:
I've opened https://github.com/microsoft/winget-pkgs/pull/682 to remove the current awscli 2.0 manifest, as its clearly broken.

jamierocks:
This has been resolved with https://github.com/microsoft/winget-pkgs/commit/dc0deabf47897aeb7ee1e36529a08d82cc621bf6.

superusercode:
It is possible to get static links to the AWS CLI

i.e. https://awscli.amazonaws.com/AWSCLIV2-2.0.15.msi

superusercode:
See https://chocolatey.org/packages/awscli#files for examples on how to get static versions of the AWS CLI binary links

jamierocks:
Thanks @superusercode! I've opened another PR to re-add aws-cli v2 (https://github.com/microsoft/winget-pkgs/pull/820).

denelon:
We now have automation that will generate a new manifest if the installer has changed at the URL. Does that resolve this issue?

PR #682, opened 2020-05-21
GH-659 Remove Amazon AWSCLI

jamierocks:
The link that was used pointed to the **latest** 2.x release, rather
than a specific release. The current 2.x version is 2.0.15 (as of
yesterday), as such there is no benefit to continuing to provide this
manifest.

I took a cursory look to see if I could find any direct links, however
so far I've drawn a blank - I'll keep looking.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/682)


chausner:
I guess it will have to be decided how to handle websites that don't provide persistent download links for specific versions. In those cases, whenever the software gets updated, users will get a verification error until the manifest gets updated, as well. I still think it is better to have these in the repository than not at all. Maybe the SHA256 and version properties in the manifest could be made optional for those websites that have changing download versions. In those cases, the client could print a warning that the download will not be verified.

jamierocks:
Totally agree - however for the time being, I think its best to remove a (currently) erroneous manifest.

KevinLaMS:
thanks for raising the  issue

ghost (approved):
Validation has completed

PR #705, opened 2020-05-22
Super tux 0.6.2

latere-a-latere:
Part of https://github.com/MarNicGit/winget-pkgs/issues/4

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/705)


MarNicGit:
I've submitted the installer to Microsoft and they have confirmed that the installer does not contain malware.

image

KevinLaMS:
Goodnews, bad news.  :(
I worked with smartscreen to get the reputation up and it installs great.  The bad news is that I cannot get the app to run.  It just seems to crash.  Is there a dependency missing?

MarNicGit:
@KevinLaMS I'll investigate that later today - I remember the game installing fine on my regular desktop so I suspect it requires opengl which isn't available in virtualised environments

MarNicGit:
@KevinLaMS Hey there again! Just tried out some things and it appears that SuperTux simply requires OpenGL in order to run. My testing environment (the win10 dev env) experienced the silent crash you experienced, but after adding the (unsupported) RemoteFX adapter to my VM the game would start up. This also explains why it runs fine on my desktop.

denelon:
@MarNicGit I'll add a "Dependency" label to this issue. We're getting closer to starting work on dependencies. This manifest is quite old, so it would need to be updated to the new schema as well.

MarNicGit:
I'm converting this to draft, probably gonna redo the manifest later today. @denelon it looks like the PR was blocked because Super tux wouldn't run in the dev environment because it requires opengl, which sounds like it simply doesn't work in the dev environment. Osu! doesn't run in dev either because of the same issue but that PR did get passed, see #1067 and subsequent PR's referencing that one. 

Do you want to keep it blocked? I'll test it out again to see if anything changed, both with and without [RemoteFX](https://docs.microsoft.com/en-us/windows-server/virtualization/hyper-v/deploy/deploy-graphics-devices-using-remotefx-vgpu).

denelon:
@MarNicGit We've got dependencies in the v1.1 milestone. I'm not sure if that alone will fix things in the virtualized environment. If you don't mind, add your 👍 to https://github.com/microsoft/winget-pkgs/issues/12906.

If you are able to get the package to work, the "Blocking-Issue" label doesn't prevent the validation pipelines from running. We're using the label to prevent duplicate submissions for packages from different users. Once the support for the "Blocking-Issue" is in place, we will kick the builds off again to see if the manifests pass validation. 

You've already done the work to build a manifest, so we're trying to keep the value of your effort by keeping the PR open, but it's up to you 😊.

PR #738, opened 2020-05-22
Add LastPass 4.48.0

xiaoyang-sde:
The old version (4.43.0.7) is remoted because the download URL doesn't change even if a new version has been released. Thus, the installers of the older versions won't pass the SHA256 checksum verification.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/738)


samiranshah:
The installer still asks for which browsers to install LastPass on - since user input is required here - validation-unattend-failed

PR #761, opened 2020-05-22
Add Doxygen 1.8.18.0

### LINKED ISSUE #660: Add Doxygen package
# Description of the new feature/enhancement

Add Doxygen package for the tool. Download link for Windows available 
[here](https://sourceforge.net/projects/doxygen/files/latest/download)

Rover656:
Resolves #660 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/761)


swaldhoer:
In my opinion it would be better to use the zip-installer as this does not require elevated privileges and you gain nothing from using the exe-installer.

Furthermore I am not sure if using [doxygen.nl](http://www.doxygen.nl/download.html) as primary download source is a good idea. Older installers are relative early removed and then I guess the package is then broken (or did I miss something and Microsoft caches these packages?).

Rover656:
@swaldhoer I agree with the use of the zip-installer being better than the exe installer, I am happy to close the PR and await zip support (as it says in the spec it will be supported sometime) if people view this as the best way forward (or alternatively just update this once zip support is implemented).
I also agree with your point regarding the download source, I'm going to push a change which will make it use the sourceforge files instead.

ghost (approved):
Validation has completed

Rover656:
Just for conversation closure, I was unable to get the sourceforge download url to work, so I opted not to change it.

swaldhoer:
@KevinLaMS What is the repository policy on the availability of download links? I don't want to be nitpicking, but as the [doxygen.nl](http://www.doxygen.nl/download.html) download link is used this manifest will be broken a few doxygen releases later.

PR #776, opened 2020-05-22
Add Typora 0.9.86

latere-a-latere:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/776)


ousugo:
this PR is duplicated with #470 , 0.9.86 is already exist. and maybe your Publisher is wrong? Publisher should be Typora too.

PR #786, opened 2020-05-22
Add GNU Octave 5.2.0

superusercode:
"GNU Octave is a high-level language, primarily intended for numerical computations. It provides a convenient command line interface for solving linear and nonlinear problems numerically, and for performing other numerical experiments using a language that is mostly compatible with Matlab. It may also be used as a batch-oriented language."

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/786)

### DIFF
--- a/manifests/GNU/Octave/5.2.0.yaml
+++ b/manifests/GNU/Octave/5.2.0.yaml
@@ -1,0 +1,14 @@
+Id: GNU.Octave
+Version: 5.2.0
+Name: GNU Octave
+Publisher: GNU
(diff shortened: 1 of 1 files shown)


superusercode:
I should note that GNU Octave is a large application and it will take some time for your PC to install it even with an SSD.

ghost (approved):
Validation has completed

PR #805, opened 2020-05-22
WinScreenfetch added

sebagomez:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/805)


sebagomez:
The manifest was created using the Windows Store tool.  
Also, it is just a plain exe with no installation, do I still need to wrap it in an installer?

KevinLaMS:
This does not seem to install.
From what I can tell, this is an EXE that runs rather than installs.

sebagomez:
It is, it does not have an installer. Must it have one? Could it be a powershell script that does the "installation"?

denelon:
@sebagomez we don't have support for "stand alone installers" yet. https://github.com/microsoft/winget-cli/issues/182

denelon:
We need to implement support for standalone executables, and the schema needs to be updated.

PR #132, opened 2020-05-15
Androidstudio

KevinLaMS:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/132)

PR #827, opened 2020-05-22
Create SandboxTest script

felipecrs:
**PS:** check [this comment](https://github.com/microsoft/winget-pkgs/pull/827#issuecomment-656422758) if you want to use it before it gets merged.

# Description

A new PowerShell script to test the installation of any Manifest in Windows Sandbox.

# Usage

```sh-session
$ Get-Help .\Tools\SandboxTest.ps1
SandboxTest.ps1 [[-Manifest] <string>] [[-Script] <scriptblock>] [-MapFolder <string>] [<CommonParameters>]
```

## Examples

```sh-session
$ .\Tools\SandboxTest.ps1 .\Tools\SandboxTest.ps1 .\manifests\GitHub\cli\0.10.1.yaml { gh --version }
--> Validating Manifest
Manifest validation succeeded.

--> Downloading dependencies

--> Starting Windows Sandbox, and:
    - Mounting the following directories:
      - C:\Users\felip\AppData\Local\Temp\SandboxTest as read-only
      - C:\Users\felip\Repos\winget-pkgs as read-and-write
    - Installing WinGet
    - Installing the Manifest .\manifests\GitHub\cli\0.10.1.yaml
    - Refreshing environment variables
    - Running the following script:

{
 gh --version
}
```

A GIF is worth a thousand words.

U6ho8iV7OL

# To-do list

- [x] Download `winget` `.Appx` files on the fly, including the dependencies such as the `VCLibs` **(and remove them from git, of course)**

- [x] Validate Manifest file with `winget validate` before pushing to Sandbox test

- [x] Install `winget` in Sandbox 

- [x] Run `winget install -m` with the provided manifest file in Sandbox

- [x] Add ability to run the custom script as the last step in Sandbox

- [x] Automatically refresh the environment variables after installing the Manifest

- [x] Close Windows Sandbox automatically if needed

- [x] Provide a way to specify which folder to mount, default is the current

Fixes https://github.com/microsoft/winget-cli/issues/253, despite personally I would love to see this integrated into `winget-cli` itself, so at least users would not need to clone the repository before testing.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/827)

PR #843, opened 2020-05-23
Add Android Studio

felipecrs:
Android Studio installer is Nullsoft. 

It exits with status code 1223, and this is a known issue of many Nullsoft based installers when using the `/S` switch. However, the installation goes fine.

Either the `winget` should consider 1223 as a success code in case of Nullsoft, or the Android Studio team has to fix it.

I submitted an issue in Android Studio's issue tracker for that. See https://issuetracker.google.com/issues/157286802.

This pull request duplicates https://github.com/microsoft/winget-pkgs/pull/132.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/843)

### DIFF
--- a/manifests/Google/AndroidStudio/3.6.3.yaml
+++ b/manifests/Google/AndroidStudio/3.6.3.yaml
@@ -1,0 +1,14 @@
+Id: Google.AndroidStudio
+Version: 3.6.3
+Name: Android Studio
+Publisher: Google
+License: See in https://developer.android.com/studio/terms.
+LicenseUrl: https://developer.android.com/studio/terms
+Tags: android, ide, emulator, kotlin
+Description: Android Studio provides the fastest tools for building apps on every type of Android device.
+Homepage: https://developer.android.com/studio/
+Installers:
+  - Arch: x64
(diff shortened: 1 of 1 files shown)


felipecassiors:
The installation runs fine here, despite the status code issue explained in the PR description.

See it in action through this GIF with 2:30 of duration 😁.

VQhKGLsuF9

Test made with https://github.com/microsoft/winget-pkgs/pull/827

KevinLaMS:
this package seems to install correctly but we get an error 1223. I suspect this error would be OK to ignore.
microsoft/winget-cli#137

denelon:
This is an old PR so I'm not sure if the version of Android Studio is still current (maybe the older version is worth having in the repository). The schema needs to be updated, and if the error code referenced is still a success result, that can be added to the manifest so the client doesn't error out.

felipecrs:
Well, there is https://github.com/microsoft/winget-pkgs/tree/master/manifests/g/Google/AndroidStudio, so I'm closing this.

PR #847, opened 2020-05-23
Add FTB App manifest

palmtree5:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/847)

### DIFF
--- a/manifests/FTB/FTB/1.0.8.yaml
+++ b/manifests/FTB/FTB/1.0.8.yaml
@@ -1,4 +1,4 @@
-Id: FTB.FTBApp
+Id: FTB.FTB
 Version: 1.0.8
 Name: FTBApp
 Publisher: Feed The Beast
--- a/manifests/FTB/FTB/1.0.8.yaml
+++ b/manifests/FTB/FTB/1.0.8.yaml
@@ -1,0 +1,15 @@
+Id: FTB.FTBApp
+Version: 1.0.8
+Name: FTB App
+Publisher: Feed The Beast
+License: LGPL 2.1
+LicenseUrl: https://github.com/FTBTeam/FTB-App/blob/master/LICENSE.md
+AppMoniker: ftb
+Tags: game, minecraft, ftb
+Homepage: https://www.feed-the-beast.com/
+Installers:
+  - Arch: x64
+    Url: https://apps.modpacks.ch/FTBApp/release/1.0.8/FTBA_windows-x64_1_0_8.exe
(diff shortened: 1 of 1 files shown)


manifests/FTB/FTB/1.0.8.yaml
```diff
@@ -0,0 +1,15 @@
+Id: FTB.FTB
+Version: 1.0.8
+Name: FTBApp
+Publisher: Feed The Beast
+License: LGPL 2.1
```
jamierocks, line 5: While this is true for the frontend, its not true for the backend (https://github.com/CreeperHost/modpacklauncher) which the installer is built from.

jamierocks (changes requested):

manifests/FTB/FTB/1.0.8.yaml
palmtree5: Fair point, I updated it to mention the licenses for both the frontend and the backend

jamierocks (changes requested):
Are there any switches to do a silent install?

palmtree5:
not that I could find

manifests/FTB/FTB/1.0.8.yaml
```diff
@@ -0,0 +1,15 @@
...
+License: LGPL 2.1 (frontend), CC BY-NC-ND (backend)
+LicenseUrl: https://github.com/FTBTeam/FTB-App/blob/master/LICENSE.md
+AppMoniker: ftb
+Tags: game, minecraft, ftb
+Homepage: https://www.feed-the-beast.com/
+Installers:
+  - Arch: x64
+    Url: https://apps.modpacks.ch/FTBApp/release/1.0.8/FTBA_windows-x64_1_0_8.exe
+    Sha256: 79372BA75C585D1F4DBE288C71591735FD288E2C821CF94940A8CC2551C04948
+    InstallerType: exe
```
jamierocks, line 14: Do we know what installer they use? nullsoft, etc

jamierocks (changes requested):

manifests/FTB/FTB/1.0.8.yaml
palmtree5: After reading install4j's docs, I figured out that it's `-q`.

ghost (approved):
Validation has completed

PR #880, opened 2020-05-23
Update 4.16.0.yaml

noahgaertner:
The github urls for old versions of calibre break with every new release because they get deleted from the releases page, this change grabs the installer from the website old releases page instead so it keeps working.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/880)


ghost (approved):
Validation has completed

PR #895, opened 2020-05-23
Add WiXToolset 3.11.2

chausner:
Note: WiX Toolset requires the .NET Framework 3.5 Windows feature to be enabled.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/895)

### DIFF
--- a/manifests/WiXToolset/WiXToolset/3.11.2.yaml
+++ b/manifests/WiXToolset/WiXToolset/3.11.2.yaml
@@ -1,6 +1,6 @@
 Id: WiXToolset.WiXToolset
 Version: 3.11.2
-Name: WiXToolset
+Name: WiX Toolset
 Publisher: WiXToolset
 License: Copyright (c) .NET Foundation and contributors
 LicenseUrl: https://github.com/wixtoolset/wix3/blob/master/LICENSE.TXT
--- a/manifests/WiXToolset/WiXToolset/3.11.2.yaml
+++ b/manifests/WiXToolset/WiXToolset/3.11.2.yaml
@@ -1,7 +1,7 @@
 Id: WiXToolset.WiXToolset
 Version: 3.11.2
 Name: WiX Toolset
-Publisher: WiXToolset
+Publisher: WiX Toolset
 License: Copyright (c) .NET Foundation and contributors
 LicenseUrl: https://github.com/wixtoolset/wix3/blob/master/LICENSE.TXT
 Tags: msi, wix, installer
--- a/manifests/WiXToolset/WiXToolset/3.11.2.yaml
+++ b/manifests/WiXToolset/WiXToolset/3.11.2.yaml
@@ -1,0 +1,17 @@
+Id: WiXToolset.WiXToolset
+Version: 3.11.2
+Name: WiXToolset
+Publisher: WiXToolset
+License: Copyright (c) .NET Foundation and contributors
+LicenseUrl: https://github.com/wixtoolset/wix3/blob/master/LICENSE.TXT
+Tags: msi, wix, installer
+Description: The WiX toolset lets developers create installers for Windows Installer, the Windows installation engine.
+Homepage: https://wixtoolset.org/
+Installers:
+  - Arch: x86
+    Url: https://github.com/wixtoolset/wix3/releases/download/wix3112rtm/wix311.exe
+    Sha256: 32BB76C478FCB356671D4AAF006AD81CA93EEA32C22A9401B168FC7471FECCD2
+    InstallerType: exe
+    Switches:
+      Silent: /q
+      SilentWithProgress: /q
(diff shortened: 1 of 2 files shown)


manifests/WiXToolset/WiXToolset/3.11.2.yaml
```diff
@@ -0,0 +1,17 @@
...
+Description: The WiX toolset lets developers create installers for Windows Installer, the Windows installation engine.
+Homepage: https://wixtoolset.org/
+Installers:
+  - Arch: x86
+    Url: https://github.com/wixtoolset/wix3/releases/download/wix3112rtm/wix311.exe
+    Sha256: 32BB76C478FCB356671D4AAF006AD81CA93EEA32C22A9401B168FC7471FECCD2
+    InstallerType: exe
+    Switches:
+      Silent: /q
+      SilentWithProgress: /q
```
TheQuazman, line 17: Silent with progress is "/passive".  "/q" doesn't show progress

manifests/WiXToolset/WiXToolset/3.11.2.yaml
```diff
@@ -0,0 +1,17 @@
+Id: WiXToolset.WiXToolset
+Version: 3.11.2
+Name: WiX Toolset
+Publisher: WiX Toolset
```
TheQuazman, line 4: When installed, the publisher shows as ".NET Foundation"

manifests/WiXToolset/WiXToolset/3.11.2.yaml
```diff
@@ -0,0 +1,17 @@
+Id: WiXToolset.WiXToolset
```
TheQuazman, line 1: This would change with the publisher, but not sure if a '.' is allowed at the start of the Id.  When I was testing mine I did "dotnet.foundation.wix.toolset"

TheQuazman (changes requested):
Thanks for adding this, I was about to, but figured I should check the pull requests first.

manifests/WiXToolset/WiXToolset/3.11.2.yaml
chausner: Thanks, fixed it.

manifests/WiXToolset/WiXToolset/3.11.2.yaml
chausner: I was unsure about this one. I changed it.

manifests/WiXToolset/WiXToolset/3.11.2.yaml
chausner: It seems PascalCasing is used by the vast majority of packages, so I chose DotNetFramework.

KevinLaMS:
when trying to install it showed an error, and shows UI.

manifests/WiXToolset/WiXToolset/3.11.2.yaml
chausner: @TheQuazman "/passive" does not seem to work for this installer. The UI is displayed but it shows a generic error and installation does not start.

chausner:
> when trying to install it showed an error, and shows UI.

It should only show UI if running in SilentWithProgress mode but not in silent mode. As I wrote in the original post, the setup requires .NET Framework 3.5 to be installed. Can you please make sure you have it enabled in the Windows optional features and try again?

KevinLaMS:
image

Looks like there is a dependency that is missing.

KevinLaMS:
https://github.com/microsoft/winget-cli/issues/182

denelon:
This PR is a bit old :smile:

I'm not sure if we need support for dependencies and/or standalone .exe applications for this to work.

In either case, the manifest needs to be updated to the current schema.

chausner:
> I'm not sure if we need support for dependencies and/or standalone .exe applications for this to work.

I would say this is a very special case since the external dependency (.NET Framework 3.5) is not a regular program/package but has to be enabled as a Windows feature.

denelon:
@chausner we added support for Windows Features as a possible type of dependency in the v1.0 schema. We're still looking into the implications of making Windows Features as dependencies since it may require a reboot. I've seen one suggestion where they are created as "packages" so the package level dependency could be used. More analysis is required to understand edge cases and the best UX.

KevinLaMS:
Clearing assignees until unblocked.

ItzLevvie (approved):

Trenly:
I'm going to close this PR because it is so old and doesn't conform to any of the current schemas. Additionally, the base branch is (at the time of writing) 84,000 commits behind `master` and it would be better to create new manifests rather than rebasing or merging in master. I will leave #96043 open to track the progress against WiXToolset, as it is based on the latest updates to the manifest schema and is a newer branch from `master`.  Thank you for your PR!

Close with reason: See above;

PR #915, opened 2020-05-24
Adds HandBrake

gab9281:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/915)


superusercode:
HandBrake uses the "nullsoft" installertype. By using this type you won't need to include the switches for silencing since they are automatically included in that type.

I can't suggest this change because github recognizes your yml as a binary file.

ghost (approved):
Validation has completed

PR #917, opened 2020-05-24
Added Brave 1.9.72

SoftCreatR:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/917)

### DIFF
--- a/manifests/BraveSoftware/Brave/1.9.72.yaml
+++ b/manifests/BraveSoftware/Brave/1.9.72.yaml
@@ -1,0 +1,23 @@
+Id: brave.browser.stable
+Publisher: Brave Software, Inc.
+Name: Brave Browser
+AppMoniker: brave
+Version: 1.9.72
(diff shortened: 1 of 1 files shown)


SoftCreatR:
@denelon Should I just remove the fields

- SystemAppId
- Channel

even if they should be supported, according to the manifest specification?

VIKTORVAV99:
The app ID should be: Brave.BraveBrowser.

The first field is the company name and the second field should be the software name.

SoftCreatR:
> The app ID should be: Brave.BraveBrowser.
> 
> The first field is the company name and the second field should be the software name.

That will be corrected during my next PR. However, i'm waiting for feedback regarding the validation failures, before pushing it.

Ty anyways.

PR #930, opened 2020-05-24
Add INSTALLEVEL3 for options important to development to the adoptopenjdk14

superusercode:
This makes it so that the JAVA_HOME variable is updated and it can be actually utilized by the software that depends on it in order to detect the JDK. Documented on https://adoptopenjdk.net/installation.html?variant=openjdk14&jvmVariant=hotspot#x64_win-jdk

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/930)


ghost (approved):
Validation has completed

PR #934, opened 2020-05-24
Add Screenpresso

gpon64:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/934)


DocBox12:
Why was pipelines not started?

KevinLaMS:
The problem with this submission is that it is not using the right path.  The publisher folder should be in manifests.

Issue #897, opened 2020-05-23
Package parity with PortableApps


superusercode:
This is primarily a tracking issue for myself with the intent to add packages not yet included from PortableApps, make sure they use static installer links, ensure the right installer types are being used, and include as much information possible in the manifests (such as descriptions, tags, cli names, etc.). I will avoid unsupported or EOL releases.

## Accessibility
- [ ]  Balabolka(Freeware) - read text aloud or save to audio file (zip)
- [ ]  Dicom- word completion utility
- [ ]  DSpeech(Freeware) - read text aloud or save to audio file
- [ ]  On-Screen Keyboard- Easily access an on-screen keyboard
- [ ]  Virtual Magnifying Glass- A full-featured screen magnifier
- [x]  Firefox Accessibility Extension - Make Firefox more accessible
## Development
- [ ]  AkelPad- lightweight plaintext editor
- [x]  Cppcheck- C/C++ analysis tool
- [ ]  Database Browser(Freeware) - Database management utility
- [ ]  DB Browser for SQLite- Simple SQLite database management tool
- [ ]  Frhed- free hex editor
- [x]  Geany- Text editor and basic IDE
- [ ]  gVim- advanced, feature-rich text editor
- [ ]  HM NIS Edit- NSIS development environment
- [ ]  IniTranslator- INI translation tool
- [x]  Notepad++- A full-featured text editor with syntax highlighting
- [ ]  Notepad2- Simple, lightweight text editor
- [ ]  Notepad2-mod- Lightweight text editor
- [x]  NSIS- Windows installer and launcher creation
- [ ]  Nvu& KompoZer- The easy-to-use Nvu web editor
- [ ]  PortableApps.com Installer - the easy way to packageapps
- [ ]  PortableApps.com Launcher - make appswithout writing code
- [ ]  Sqliteman- SQLite databases made easy
- [ ]  SWI-Prolog- Prolog programming environment
- [ ]  Q-Eye(Freeware) - lightweight QlikView data file viewer
- [ ]  XAMPP - Apache, mySQL, PHP, phpMyAdmin, etc in one package
## Education
- [x]  Artha- offline thesaurus
- [ ]  BPBible- bible study tool
- [ ]  Celestia-space simulator
- [x]  FreeMat- scientific prototyping and technical computing
- [x]  GoldenDict- dictionary and encyclopedia lookups
- [ ]  Gramps- genealogical research software
- [ ]  Marble- virtual globe and atlas
[...]

PR #961, opened 2020-05-24
Add AkelPad Portable 4.9.8

LukasNespor:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/961)

### DIFF
--- a/manifests/AkelPad/AkelPad/4.9.8.yaml
+++ b/manifests/AkelPad/AkelPad/4.9.8.yaml
@@ -1,0 +1,13 @@
+Id: AkelPad.AkelPad
+Name: AkelPad Portable
+Description: Lightweight text editor
+Version: 4.9.8
+Publisher: AkelPad
+License: GNU General Public License
+Homepage: http://akelpad.sourceforge.net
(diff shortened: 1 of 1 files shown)


LukasNespor:
Ref #897

LukasNespor:
Ref #897

MarNicGit:
@LukasNespor I've noticed you're adding a lot of portable editions of applications. Wouldn't it be a better idea to append the package name with something like `Portable` to avoid future conflict with the default desktop versions of these applications?

LukasNespor:
> @LukasNespor I've noticed you're adding a lot of portable editions of applications. Wouldn't it be a better idea to append the package name with something like `Portable` to avoid future conflict with the default desktop versions of these applications?
> 
> In this case the package name would be `AkelPad.AkelPadPortable`.

Good point. Closing this PR and creating new #1069 because I am an id*ot :) and removed my repo so I am not able to create new commit with change.

PR #1001, opened 2020-05-24
Add IceChat

ath0mas:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1001)

### DIFF
--- a/manifests/IceChat/IceChat/9.22.yaml
+++ b/manifests/IceChat/IceChat/9.22.yaml
@@ -1,0 +1,14 @@
+Id: IceChat.IceChat
+Version: 9.22
+Name: IceChat
+Publisher: IceChat
+License: Copyright (c) IceChat Networks
+LicenseUrl: https://github.com/icechat/IceChat/blob/master/LICENSE.TXT
+Tags: irc client, mirc, icechat, chat
+Description: IceChat is a fully featured, simple to use IRC Client allowing you to chat online with your friends and for meeting new people as well.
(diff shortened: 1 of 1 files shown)


KevinLaMS:
This app has a dependency on the ,NET framework.  And though I try to install it, I cannot get it to successfully install.  Is there a way to add the framework to your installation package?

ath0mas:
@KevinLaMS my PR but not my package :/
so, any feedback on that _please_ @icechat ?

icechat:
There is no installer yet for IceChat 9.50

ath0mas:
trying to add version 9.22 here,
so you say it should be fixed for future 9.50 only?
nothing more to do than to close this PR about 9.22 for now?

icechat:
The installer for 9.22 will install .NET 3.5 automatically. 
9.50 has no installer as of yet, and it requires .NET 4.5.2 which comes by default with Windows 10.

ath0mas:
ok thanks @icechat !
@KevinLaMS so .NET 3.5 should be installed automatically it seems

icechat:
It is not an MSI installer, but an installer made by InnoSetup. Could this be a problem?

ath0mas:
Inno is supported by Winget, and used by many other packages here

ghost (approved):
Validation has completed

PR #518, opened 2020-05-20
Add foobar2000

boarnoah:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/518)

PR #1016, opened 2020-05-25
Add foobar2000 1.5.4.0

suasgn:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1016)

### DIFF
--- a/manifests/foobar2000/foobar2000/1.5.4.0.yaml
+++ b/manifests/foobar2000/foobar2000/1.5.4.0.yaml
@@ -2,7 +2,8 @@ Id: foobar2000.foobar2000
 Publisher: foobar2000
 Name: foobar2000
 Version: 1.5.4.0
-License: BSD
+License: Copyright © Peter Pawlowski
+LicenseUrl: https://www.foobar2000.org/license
 Tags: music player
 Description: foobar2000 is an advanced freeware audio player for the Windows platform.
 Homepage: https://www.foobar2000.org/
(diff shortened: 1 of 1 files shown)


boarnoah:
Hi this is going to run into the same issues as my prior PR for this program https://github.com/microsoft/winget-pkgs/pull/518#issuecomment-633364260

Namely the URLs are not permanent, they expire after a time period.

superusercode:
why not try a mirror like https://www.majorgeeks.com/files/details/foobar2000.html ?

suabahasa:
I'm update the url from videohelp. the hash is original from the official one

manifests/foobar2000/foobar2000/1.5.4.0.yaml
```diff
@@ -0,0 +1,13 @@
+Id: foobar2000.foobar2000
+Publisher: foobar2000
+Name: foobar2000
+Version: 1.5.4.0
+License: BSD
```
boarnoah, line 5: I'm fairly sure the foobar2000 app is not BSD licensed, the official website lists it as [foobar2000 license](https://www.foobar2000.org/license) which is a custom freeware thing.

suabahasa: what should I change it to? `Copyright © Peter Pawlowski` or `Freeware` ?

boarnoah: Copyright would be more accurate I think, you can also add the additional field LicenseUrl that points to the document. :)

KevinLaMS:
I am not sure if the settings are correct.  When  i ran the installer, nothing seemed to install.

superusercode:
@KevinLaMS according to https://silentinstallhq.com/foobar2000-silent-install-how-to-guide/ the silent install must be done in an elevated command prompt. Maybe that has to do with it not installing properly?

PR #1035, opened 2020-05-25
Add Google Chrome 83.0.4103.61

dmex:
This PR updates to the latest version of Google Chrome with the correct link and required arguments for the consumer version.

The previous winget package for Google Chrome downloads the enterprise version which includes a bundle of enterprise management components and deprecated legacy features for interoperability.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1035)

### DIFF
--- a/manifests/Google/Chrome/83.0.4103.61.yaml
+++ b/manifests/Google/Chrome/83.0.4103.61.yaml
@@ -1,0 +1,17 @@
+Id: Google.Chrome
+Name: Chrome
+Publisher: Google
+Version: 83.0.4103.61
+Homepage: https://www.google.com/chrome/
+InstallerType: exe
+AppMoniker: chrome
+License: Google Chrome and Chrome OS Additional Terms of Service
+LicenseUrl: https://www.google.com/chrome/terms/
+Installers:
+  - Arch: x64
+    Url: https://dl.google.com/edgedl/chrome/install/ChromeStandaloneSetup64.exe
+    Sha256: 4CA5DF9D36510D35082025FE2AF4F12D0CEBCA4221DE881C8F66530335F97AAD
+    Language: en-US
+    Switches:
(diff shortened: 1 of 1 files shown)


dmex:
The manifest is correct and the application installs correctly:

image

@KevinLaMS @RDMacLachlan

The pipeline successfully validated the package and the installation succeeds because we executed winget as administrator but the manual review was executed as standard user?

I need some advice about what should be done about it. 

1) Remove the /silent switch.
or
2) Change "needsadmin=prefers" to "needsadmin=false"

The /silent switch prevents the chrome installer (due to a winget bug) from showing a UAC prompt for a machine-wide installation. Removing /silent fixes standard user installation and the install is still automated but you'll see a GUI progress window while the install completes and the window still closes itself - just like /silent but without /silent.

Alternatively changing needsadmin to false forces the per-user installation and we don't have to remove /silent but it will install per-user instead of per-machine.
needsadmin=false will install per user.
needsadmin=true will install machine-wide.
needsadmin=prefers will install machine-wide if permissions allow.

I used 'prefers' and running winget as administrator succeeds the install but running winget as standard user causes an error because winget executes binaries with low integrity (https://github.com/microsoft/winget-cli/issues/254)

So do we change the arguments to install per-user or remove the /silent switch and install per-macine?

For reference the chrome standaline installer is opensource and you can view the installer flags here: https://github.com/google/omaha/blob/51cbcac4c00afdce0a3d72a235f1167e11762765/omaha/common/const_goopdate.h#L82-L99

Chrome installer switches are located here:
https://github.com/google/omaha/blob/master/omaha/common/const_cmd_line.h

dmex:
@KevinLaMS @RDMacLachlan

Still waiting for feedback about switching to the less secure per-user install for the manual review to succeed... That delay has now also caused a merge conflict.

denelon:
#1254 was merged for this version of Chrome.

dmex:
@denelon 

This PR also changed the installer from the enterprise version to the consumer version and not just the version?

PR #1061, opened 2020-05-25
Add 32-Bit support

BergischNRW:
Add 32-Bit Version of 7-Zip 19.00.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1061)


BergischNRW:
I guess this will have to wait then until multi-arch manifests are supported.

PR #1067, opened 2020-05-25
Add Osu! 2020.523.0

latere-a-latere:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1067)

### DIFF
--- a/manifests/Peppy/Osu!/2020.523.0.yaml
+++ b/manifests/Peppy/Osu!/2020.523.0.yaml
@@ -1,0 +1,17 @@
+Id: Peppy.Osu!
+Version: 2020.523.0
+Name: Osu!
+Publisher: Dean "peppy" Herbert
+License: MIT
+AppMoniker: osu
+Tags: game, rhythm, taiko, ddr, stepmania, beatmania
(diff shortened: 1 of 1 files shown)


KevinLaMS:
It appeared to install and uninstall correctly.  However running the app did nothing. What am I mising?  Thanks.

MarNicGit:
@KevinLaMS After doing some research and trying installing it myself in the Win10 Dev environment, it appears that the application fails to launch if the system either runs off an Intel GPU or if it runs off the GDI Generic driver. (which is the case of the win10 dev vm)

Do any of these two reasons apply to you? Launching the game on a system with a dedicated GPU (my desktop with an RX5700XT) doesn't seem to give any problems whatsoever.

Issues I've found referencing this problem:
https://github.com/ppy/osu/issues/5585
https://github.com/ppy/osu/issues/4235
https://github.com/ppy/osu/issues/3025
https://github.com/ppy/osu/issues/6585
https://github.com/ppy/osu/issues/6496

ghost (approved):
Validation has completed

PR #1070, opened 2020-05-25
Add Cppcheck Portable 2.0

LukasNespor:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1070)


LukasNespor:
Ref #897

LukasNespor:
@KevinLaMS every package has to be installable unattended? If so it means that any portable (from PortableApps.com) package can be used in winget because the package does not have any switches and requires user to select where to "install" the app.

denelon:
This PR is a bit old :smile:

I'm not sure if the package can now be installed "silently" or not, but the manifest will need to be updated in either case.

PR #1071, opened 2020-05-25
Add FreeFileSync 10.24

Teraskull:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1071)


Teraskull:
> Please check the HASH and submit again. Sorry for the inconvenience.

Same hash, I don't know where is the issue.

KevinLaMS:
From what I see, this installertype is inno
When I switch to that, it fails because it does not support silent.
IMO This app is blocked until we support an un-attend installer.

image

PR #1076, opened 2020-05-25
Adding DYMO Connect

miniksa:
This installer is a little weird in that it unwraps other sub-installers including an embedded MSI. `winget` loses track of this chain and says "installed successfully" while one of the chain installers is still operating in the background unpacking things. But eventually it will shake out and it will be installed.

I couldn't find a manifest directive to give `winget` any additional information on how to track the state.

The silent switch comes directly from running `DCDSetup1.3.1.exe /?` and copying out the silent instructions from the installer's built-in help pop-up.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1076)

### DIFF
--- a/manifests/SanfordLP/DYMOConnect/1.3.1.yaml
+++ b/manifests/SanfordLP/DYMOConnect/1.3.1.yaml
@@ -1,0 +1,17 @@
+Id: SanfordLP.DYMOConnect
+Version: 1.3.1
+Name: DYMO Connect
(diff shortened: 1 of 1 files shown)


ghost (approved):
Validation has completed

PR #1078, opened 2020-05-25
Adding KDEnLive app

ebmarquez:
Adding KDEnLive Video Editor app.

```powershell
> winget install --manifest 20.04.1.yaml
Found KDEnLive [kdenlive.Release.20.04.1]
Downloading https://files.kde.org/kdenlive/release/kdenlive-20.04.1.exe
  ██████████████████████████████  81.8 MB / 81.8 MB
Installing ...
Successfully installed!
```

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1078)


ebmarquez:
@RDMacLachlan is there anything pending for this PR?

ghost (approved):
Validation has completed

IlyaBizyaev:
Kdenlive is served through KDE's CDN:
https://files.kde.org/kdenlive/release/kdenlive-20.04.1.exe
Is it preferrable to hardcode a specific mirror here?

ebmarquez:
The primary location does a 302 redirect to a mirror.  winget seems to fail on the 302 redirect.

ebmarquez:
Is there a way to set up the YAML package so it can work with a 302 redirect?

nicolas17:
With my KDE sysadmin team hat on:
It would be preferred if you can use the files.kde.org URL and let it redirect to an appropriate mirror.

However, if that's not possible, I would suggest using https://cdn.files.kde.org/ instead of hardcoding a random mirror. At least unless you ask the mirror owners about it.

We usually use the CDN only as a fallback when no mirror has synced the file yet (files.kde.org 302-redirects to cdn.* if it has no other option), because it's a sponsored service and we don't want to go way overboard in bandwidth use (eg. we have no plans to replace the mirror network with it). But using it in winget should be fine :)

PR #1086, opened 2020-05-25
VcXsrv uses nullsoft, not a generic exe

superusercode:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1086)

### DIFF
--- a/manifests/marha/VcXsrv/1.20.8.1.yaml
+++ b/manifests/marha/VcXsrv/1.20.8.1.yaml
@@ -11,7 +11,4 @@ Installers:
   - Arch: x64
     Url: https://sourceforge.net/projects/vcxsrv/files/latest/download
     Sha256: BA42C20502ACABF6B0626261FAA6E1862DF2F1899B126BA69E72EF9EA90ADDE4
-    InstallerType: exe
-    Switches:
-      Silent: "/S"
-      SilentWithProgress: "/S"
+    InstallerType: nullsoft


ghost (approved):
Validation has completed

echoix:
The download URL uses a /latest shortcut (https://sourceforge.net/projects/vcxsrv/files/latest/download), instead of directly pinning a particular file. This will cause a problem if the version is changed. The file should be pinned specifically to the file downloaded on the page https://sourceforge.net/projects/vcxsrv/files/vcxsrv/1.20.8.1/vcxsrv-64.1.20.8.1.installer.exe/download . I don't know exactly how to use the sourceforge links though.

superusercode:
@echoix that is the right way to do it, sorry I missed that on my first PR. Feel free to make a PR amending this.

echoix:
I’m not confortable doing it, since I didn’t understand how to get a direct download link from Sourceforge. On my browser it hand the landing page with ads, and the direct link didn’t seem to work on another browser (took more than 15 sec to respond)

superusercode:
@echoix that link exactly works with winget

you can get the direct download link (using chrome/edge) doing this: 
image

i.e. https://newcontinuum.dl.sourceforge.net/project/vcxsrv/vcxsrv/1.20.8.1/vcxsrv-64.1.20.8.1.installer.exe

but the downside to this is that some users on winget may not be near these servers or these download servers could eventually be removed

if you do submit a pr, all you'll have to do is change the download url line.

PR #1159, opened 2020-05-26
Add silent switches prompt to YamlCreate.ps1

ainen:
Added prompts for optional silent switches in YamlCreate.ps1

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1159)


chausner:
I guess it makes sense to only ask for these if the installer type was specified as "exe" as for the others it is automatically handled by winget.

ainen:
> I guess it makes sense to only ask for these if the installer type was specified as "exe" as for the others it is automatically handled by winget.

This is true. I've added some basic logic to confirm that the installer type is exe before prompting for silent switches.

KevinLaMS (approved):

KevinLaMS:
Thanks @ainen

PR #1177, opened 2020-05-26
Add Notepads

0x7c13:
Add Notepads
Project Page: https://github.com/JasonStein/Notepads
Store link: https://www.microsoft.com/store/apps/9nhl4nsc67wm

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1177)

### DIFF
--- a/manifests/Notepads/Notepads/1.3.5.0.yaml
+++ b/manifests/Notepads/Notepads/1.3.5.0.yaml
@@ -2,7 +2,7 @@ Id: Notepads.Notepads
 Publisher: Notepads
 Name: Notepads
 Version: 1.3.5.0
-License: MIT
+License: MIT License - Copyright (c) Notepads
 InstallerType: MSIX
 LicenseUrl: https://github.com/JasonStein/Notepads/blob/master/LICENSE.md
 AppMoniker: 
--- a/manifests/Notepads/Notepads/1.3.5.0.yaml
+++ b/manifests/Notepads/Notepads/1.3.5.0.yaml
@@ -2,11 +2,14 @@ Id: Notepads.Notepads
 Publisher: Notepads
 Name: Notepads
 Version: 1.3.5.0
-License: MIT License - Copyright (c) Notepads
+License: MIT License
 InstallerType: MSIX
 LicenseUrl: https://github.com/JasonStein/Notepads/blob/master/LICENSE.md
-AppMoniker: 
-Tags: notepad, text editor, fluent design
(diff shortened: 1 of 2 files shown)


Teraskull:
The correct folder structure should be Publisher/Package, in your case it is Package/Package.
Change it to: JackieLiu/Notepads

Only one installer is allowed per manifest, so separate the x64, x86 and ARM64 in different manifests.

Do not include empty keys (AppMoniker: )

And also, for the future, one manifest is allowed per PR.

JasonStein:
> The correct folder structure should be Publisher/Package, in your case it is Package/Package.
> Change it to: JackieLiu/Notepads
> 
> Only one installer is allowed per manifest, so separate the x64, x86 and ARM64 in different manifests.
> 
> Do not include empty keys (AppMoniker: )
> 
> And also, for the future, one manifest is allowed per PR.

I am changing it to Notepads.Notepads and also remove all archs and replaced with Neutral.

Teraskull:
The correct folder format should be Publisher/Package, in your case it is Package/Package.

> > The correct folder structure should be Publisher/Package, in your case it is Package/Package.
> > Change it to: JackieLiu/Notepads
> > Only one installer is allowed per manifest, so separate the x64, x86 and ARM64 in different manifests.
> > Do not include empty keys (AppMoniker: )
> > And also, for the future, one manifest is allowed per PR.
> 
> I am changing it to Notepads.Notepads and also remove all archs and replaced with Neutral.

Great, should work now

KevinLaMS:
The app appears to be signed incorrectly.  
"Install failed: Error <number> the root certificate of the signature in the app package or bundle must be trusted....

JasonStein:
> The app appears to be signed incorrectly.
> "Install failed: Error the root certificate of the signature in the app package or bundle must be trusted....

I am always using the store certificate to sign my app. If I install it using the  powershell script, it will install the cert and then the app without any problem. But for winget, I can only provide the link to the msixbundle binary file itself. I guess there is some gap here or do I need to do something differently?

JasonStein:
> The app appears to be signed incorrectly.
> "Install failed: Error the root certificate of the signature in the app package or bundle must be trusted....

I uploaded a new one to replace it. Can you help install again? It should work now.

ghost (approved):
Validation has completed

PR #1195, opened 2020-05-26
Add Microsoft Comic Chat v25

latere-a-latere:
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1195)

### DIFF
--- a/manifests/Microsoft/MSChat/v25.yaml
+++ b/manifests/Microsoft/MSChat/v25.yaml
@@ -1,0 +1,17 @@
+Id: Microsoft.MSChat
+Version: v25
+Name: Comic chat
+Publisher: Microsoft
+License: Copyright 1996 (c) Microsoft Corporation
+AppMoniker: mschat
+Tags: chat, abandonware, irc, comic, david kurlander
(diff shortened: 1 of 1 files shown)


KevinLaMS:
Well you got me.  It does not pass.  The app has an error as soon as it runs.

MarNicGit:
@KevinLaMS what error do you get? It runs out of the box on my win10 machine

MarNicGit:
@KevinLaMS Just installed it on the Win10 Dev environment and I'm assuming the error you mentioned is the following one:

> Failed to update the system registry. 
> Please try using REGEDIT.

Unfortunately this error pops up because of MSChat's old age; The program doesn't trigger UAC so when it tries to launch and edit something in the registry, it fails, giving you this message. You can however ignore the message and use the program without a hitch, and running MSChat as admin once gets rid of the message entirely.

Do note that the Microsoft Chat Servers have long been shut down; connecting to a still running IRC server will work without a problem. Rizon and Freenode are good examples, though the latter slows down the channel list a whole lot due to it's abundance of channels.

PR #1204, opened 2020-05-27
LyX 2.3.4.4

bquistorff:
I guessed at a few parameters:
- installer type: I choose `exe` since that's the installer filename extension, but not sure if it could be `inno` or `nirsoft`
- arch: I choose `x86` since the installer puts files into `Program Files (x86)`.

Let me know if there are better ways.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1204)


ghost (approved):
Validation has completed

Issue #1161, opened 2020-05-26
Allow InstallShield pass validation


Reset12138:
#661 For example, the Adobe Acrobat Reader DC's installer type should be InstallShield.
![image](https://user-images.githubusercontent.com/40438607/82923174-2ceb5400-9fad-11ea-8d26-d7d247bb6088.png)
Although if i changed the type to inno in the yaml and the validation run successfully, winget failed the validation of type if the installertype is InstallShield.
![image](https://user-images.githubusercontent.com/40438607/82923292-51dfc700-9fad-11ea-82b3-e0e551173393.png)
![image](https://user-images.githubusercontent.com/40438607/82923512-abe08c80-9fad-11ea-9052-15a50be385af.png)
![image](https://user-images.githubusercontent.com/40438607/82923539-b569f480-9fad-11ea-8592-f7f75b167f08.png)

oxygen-dioxide:
"InstallShild" should be "InstallShield". However, "InstallShield" failed validation, too.

Reset12138:
@oxygen-dioxide oh thank you for catching this, I will correct the mistake right now

denelon:
Is this issue resolved?

oxygen-dioxide:
> Is this issue resolved?

No. Installshield needs config file #246

Trenly:
@denelon - Since this is duplicated over at CLI, can probably be closed here?

PR #1213, opened 2020-05-27
Add BingWallpaper

Reset816:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1213)

### DIFF
--- a/manifests/Microsoft/BingWallpaper/1.0.7.4.yaml
+++ b/manifests/Microsoft/BingWallpaper/1.0.7.4.yaml
@@ -1,0 +1,13 @@
+Id: Microsoft.BingWallpaper
+Name: Bing Wallpaper
+AppMoniker: bing
+Version: 1.0.7.4
+Publisher: Microsoft
+Author: Microsoft
+Homepage: https://www.microsoft.com/en-us/bing/bing-wallpaper
+Description: Bing Wallpaper includes a collection of beautiful images from around the world that have been featured on the Bing homepage.
+InstallerType: inno
+Installers:
+  - Arch: x86
+    Url: https://download.microsoft.com/download/e/8/f/e8ff8991-04f4-4a97-9339-871b30e12eea/Installer/1.0.7.4/Bing/W011/BingWallpaper.exe
+    Sha256: 5C37052C95CB9A8876EFBDDDD2906BC443C21BEC41A6638D1B1C7BFB61D3A288
--- a/manifests/Microsoft/BingWallpaper/1.0.7.4.yaml
+++ b/manifests/Microsoft/BingWallpaper/1.0.7.4.yaml
@@ -7,8 +7,11 @@ Publisher: Microsoft
 Author: Microsoft
 Homepage: https://www.microsoft.com/en-us/bing/bing-wallpaper
 Description: Bing Wallpaper includes a collection of beautiful images from around the world that have been featured on the Bing homepage.
-InstallerType: inno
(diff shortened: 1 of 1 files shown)


KevinLaMS:
syntax error, mail to yao

yao-msft:
> syntax error, mail to yao

The InstallerType in the manifest is incorrect, it’s not an inno installer. Since the manifest does not declare its own Silent and SilentWithProgress switches, the BingWallPaper.exe failed with our default Inno type SilentWithProgress parameter “/silent”.

Fix: Please use exe as InstallerType and add Silent and SilentWithProgress switches with proper values.

Reset12138:
> > syntax error, mail to yao
> 
> The InstallerType in the manifest is incorrect, it’s not an inno installer. Since the manifest does not declare its own Silent and SilentWithProgress switches, the BingWallPaper.exe failed with our default Inno type SilentWithProgress parameter “/silent”.
> 
> Fix: Please use exe as InstallerType and add Silent and SilentWithProgress switches with proper values.

Strange, the interface looked like inno or installshield, and if it is one of them, it should work.#1161
 
image

However, if i use exe as InstallerType add Silent and SilentWithProgress with the value "\Q"(which it tell me), it failed too.
image

image
what value should i add?

yao-msft:
> syntax error, mail to yao
> 
> The InstallerType in the manifest is incorrect, it’s not an inno installer. Since the manifest does not declare its own Silent and SilentWithProgress switches, the BingWallPaper.exe failed with our default Inno type SilentWithProgress parameter “/silent”.
> Fix: Please use exe as InstallerType and add Silent and SilentWithProgress switches with proper values.
> 
> Strange, the interface looked like inno or installshield, and if it is one of them, it should work.#1161
> 
> However, if i use exe as InstallerType add Silent and SilentWithProgress with the value "\Q"(which it tell me), it failed too.
> 
> 
> what value should i add?

Regarding InstallShield support, unfortunately we don't have full support for automatic handling of InstallShield type yet, as InstallShield can be built as both exe and msi. And we don't support installer config files like for .iss to be used with InstallShield yet. So currently, InstallShield installers need to declare exe or msi type in the manifest respectively.

But regardless of the above, I took a quick look at the BingWallpaper.exe and I could not relate it to inno or Installshield immediately. Here is what I found. The exe manifest says it's IExpress extraction tool. It indicates this exe is a self extract archive. So I renamed the exe to .cab and it can be successfully extracted. In the cab, there's a exe and msi. By examining the extracted installer exe, I did not find specific installer info in exe resources.

Anyway, the BingWallpaper.exe and the underlying installer does not accept default inno or installer switches, so it's appropriate to use exe type for it. The /Q switch is irrelevant here since it's a switch for IExpress archive, not the underlying installer. I tried a couple of commonly used silent switches and passed to the underlying BWInstaller.exe like _BingWallpaper.exe /Q /C:"BWInstaller.exe /quiet"_ but unfortunately none of them does unattended installation.

Currently we require all installers support unattended installation. If the BingWallPaper installer does not support it(or maybe we did not find correct args for it), we'll have to put this installer on hold for now(we can revisit this in the future when we can handle interactive-only packages appropriately).

Yao

KevinLaMS:
Unfortunately this installer with the /Q switch fails.  We are going to need to have the path for supporting non-silent installs.

PR #1267, opened 2020-05-28
32-bit Firefox MSI added

danijeljw:
Added support for 32-bit MSI to yml to be inclusive. Used generator to create manifest.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1267)


SoftCreatR:
Well, looks like the check pipeline broke again, because it would fail I guess. Although the Windows Package Manager YAML Generator can create YAML files with multiple installers, winget does not support more than one installer for now.

PR #1278, opened 2020-05-28
Removed x64 Version

danijeljw:
- Removed x64
- Provided support for x86 instead
- x86 provides application that can be installed on x86 or x64
- Will update package to support both when package manager can support both versions in single YML file

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1278)


WSLUser:
Please create a x64 variant for now. I would never install a 32 bit app on my 64 bit computer unless it's the only option for the software. In that case, it would behoove me to find a better comparable app.

ghost (approved):
Validation has completed

EwoutH:
Why did this get merged?! x86 applications on a x64 system carry serious performance penalties, especially when SIMD intrinsics or assembly are involved.

ocdtrekkie:
It looks like all the later versions are the x64 version. So while this PR should not have been merged, it's pointless in the long run?

PR #1294, opened 2020-05-28
Add TeX Live 2020

ghost:
Note that this installation can never be completely silent. Also, this installation would benefit from a default install location change, but, alas, the installer does not seem to have an option for that.  

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1294)

### DIFF
--- a/manifests/TUG/TeXLive/2020.yaml
+++ b/manifests/TUG/TeXLive/2020.yaml
@@ -1,0 +1,15 @@
+Id: TUG.TeXLive
+Version: 2020
+Name: TeX Live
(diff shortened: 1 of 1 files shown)


mattiasmarka:
It seems that they are still updating the 2020 installer and there is no archived version. That is to say, the hash is only good for a short amount of time. Also, the latest installer seems to be broken. Closing for now.

PR #1304, opened 2020-05-29
R project.base

andreassag:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1304)


GaaH:
Why not call this package simply "R", or "R-base"? I think "RProject R base" is overcomplicated.

cderv:
Yes I think we should have `RProject.R` or R.R as we have `Python.Python`. 
Will then can add `Rproject.Rtools` or `R.Rtools` too: https://cran.r-project.org/bin/windows/Rtools/

exTerEX:
I'll get around to change it. ```RProject.R``` would in my opinion fit best to the scheme of the package repository.

manifests/RProject/R/4.0.0.yaml
```diff
@@ -0,0 +1,12 @@
+Id: RProject.R
+Publisher: RProject
+Name: RProject R
```
cderv, line 3: I think the name of the application would just be R... I don't usually hear people say they look for the Rproject R. 
Usually it is just R for the one from CRAN, and other name like MRO for the distribution provided by microsoft. 

Seems a bit strange this naming. Curious of what other think... 🤔

manifests/RProject/R/4.0.0.yaml
```diff
@@ -0,0 +1,12 @@
...
+Version: 4.0.0
+License: GNU General Public License, version 2
+LicenseUrl: https://www.r-project.org/Licenses/
```
cderv, line 6: There is also this licence file available
https://www.r-project.org/COPYING
I don't know which one should be included in the manifest

manifests/RProject/R/4.0.0.yaml
```diff
@@ -0,0 +1,12 @@
...
+Description: The base binary distribution of R statistics language.
+InstallerType: exe
+Installers:
+  - Arch: x64
```
cderv, line 10: This installer also works for x86 architecture (32bit). Maybe worth adding also ?

manifests/RProject/R/4.0.0.yaml
exTerEX: I agreed. I'll change the ```name``` from ```RProject R``` to ```R```.

manifests/RProject/R/4.0.0.yaml
exTerEX: I'm looking into it. Not sure exactly how this can be done, considering that every manifest I can find either use ```x86``` or ```x64``` with the exception of Rufus that has the attribute ```Neutral```, but I'm not sure exactly what this mean.

manifests/RProject/R/4.0.0.yaml
exTerEX: I think the link you are referring is more suitable.

manifests/RProject/R/4.0.0.yaml
cderv: I though a manifest can have several link, one per architecture... 🤔 
Using `WinGetYamlGenerator` from the Store, I was offered to add several binaries and it generated this 
```yaml
Installers:
  - Arch: x64
    InstallerType: EXE
    Url: https://cloud.r-project.org/bin/windows/base/R-4.0.0-win.exe
    Sha256: CA0C28A4F3E0AE66F64F01024BCCF3F3477C6E1008698AB96CEB3162E24E6C87
  - Arch: x86
    InstallerType: EXE
    Url: https://cloud.r-project.org/bin/windows/base/R-4.0.0-win.exe
    Sha256: CA0C28A4F3E0AE66F64F01024BCCF3F3477C6E1008698AB96CEB3162E24E6C87
```
So I thought it was supported

cderv: From the specification, seems ok.
https://docs.microsoft.com/fr-fr/windows/package-manager/package/manifest?tabs=minschema%2Ccompschema

cderv:
Maybe we should also add some switches like with Python to customize the installation 
https://github.com/microsoft/winget-pkgs/blob/ab5c0b84b5fca2c4370e60f763655e9cd91b2415/manifests/Python/Python/3.8.3.yaml#L15-L18

R switches : https://cran.r-project.org/bin/windows/base/rw-FAQ.html#Can-I-customize-the-installation_003f

So something like 
```yaml
Installers:
  - Arch: x64
    InstallerType: EXE
    Url: https://cloud.r-project.org/bin/windows/base/R-4.0.0-win.exe
    Sha256: CA0C28A4F3E0AE66F64F01024BCCF3F3477C6E1008698AB96CEB3162E24E6C87
    Switches:
      Silent: /VERYSILENT
      SilentWithProgress: /SILENT
```
I tested and it work 
```cmd
# Silent with progress
winget install -m .\4.0.0.0.yaml
# Silent 
winget install --silent -m .\4.0.0.0.yaml
```

This comply with the [specification Best Practive](https://docs.microsoft.com/fr-fr/windows/package-manager/package/manifest?tabs=minexample%2Ccompschema#tips-and-best-practices)
> All packages should be submitted with a silent install if possible. If you have an executable that does not support a silent install, the user experience will be diminished.

manifests/RProject/R/4.0.0.yaml
cderv: In fact reading the [spectification](https://docs.microsoft.com/en-us/windows/package-manager/package/manifest?tabs=minschema%2Cbetter), maybe you can keep this name and use the shorter `AppMoniker` field.
In any case, we should add 
```yaml
AppMoniker: R
```

manifests/RProject/R/4.0.0.yaml
```diff
@@ -0,0 +1,12 @@
...
+    Url: https://cloud.r-project.org/bin/windows/base/R-4.0.0-win.exe
+    Sha256: CA0C28A4F3E0AE66F64F01024BCCF3F3477C6E1008698AB96CEB3162E24E6C87
```
cderv, line 12: I would add some switches as the R installer allows it
```suggestion
    Sha256: CA0C28A4F3E0AE66F64F01024BCCF3F3477C6E1008698AB96CEB3162E24E6C87
    Switches:
      Silent: /VERYSILENT
      SilentWithProgress: /SILENT
      InstallLocation: /DIR="<INSTALLPATH>"
```

manifests/RProject/R/4.0.0.yaml
cderv, line 3: I would add `AppMoniker` as the specification defines it by `# The common name someone may use to search for the package.`
```suggestion
Name: RProject R
AppMoniker: R
```

cderv (commented):
I would also add the missing Manifest version
```yaml
ManifestVersion: 0.1.0
```
It seems to be in the minimal expected schema from the specification

manifests/RProject/R/4.0.0.yaml
exTerEX: Tnx. I think this will work.

manifests/RProject/R/4.0.0.yaml
exTerEX: I think changing ```name``` from ```RProject R``` to just ```R``` might be a better idea, while also including ```AppMoniker: R```. I'll wait for some feedback on this before I update the code.

manifests/RProject/R/4.0.0.yaml
exTerEX: This might not be allowed at the moment. The local validation didn't indicate anything wrong, but the Azure tests give feedback indicating that this isn't possible currently. It might be that it is simply a better idea just having ```x64``` supported, and then sometime later add support for ```x86```.

cderv: Yes you're right. Now we know 😄 
I agree with you about having only x64 for now.

cderv: Let's note that `winget validate` did not catch that on my end...

cderv:
@exTerEX just to share with you that specification in winget-cli repo about manifests are better
https://github.com/microsoft/winget-cli/blob/master/doc/ManifestSpecv0.1.md#yaml-syntax
They mention the limitation from preview version (like multiple installer not supported yet)

cderv:
Thank for taking my inputs @exTerEX ! Everything seems fine now on the automated check. Great ! 👍

cderv:
While digging a bit more working on another package, I understand now that the switches are necessary only with `InstallerType: exe` - hence the warning at validation.

Here, the R installer exe is in fact of type InnoSetup. That means we should use 
```yaml
InstallerType: inno
```
With this, the switches are automatically guessed. Specs says: 
> # The winget command tool uses this value to assist in installing this application.
# If the value is an exe, you will need to provide the quiet switches.
See source:
https://github.com/microsoft/winget-cli/blob/e997f3e394e66f2389f4f33f36c58b1c2768e975/src/AppInstallerRepositoryCore/Manifest/ManifestInstaller.cpp#L134-L141

If we change the value, we can clean the switches. I don't know what the best practices is. Reviewers will tell us I guess. Anyway, just wanted to share my findings

exTerEX:
I will look into if it has any practical difference. If not, then using ```InstallerType: inno``` should perhaps be used, just to simplify the manifest.

ghost (approved):
Validation has completed

ItzLevvie (approved):

PR #1306, opened 2020-05-29
Changed name from "Ruby" to "Ruby with DevKit"

ghost:
There are two packages with the same name:
```
Ruby                    RubyInstallerTeam.Ruby         2.7.1
Ruby                    RubyInstallerTeam.RubyWithD... 2.7.1.1
```

From the output you can't see that it is Ruby with DevKit, so the name should be changed from "Ruby" to "Ruby with DevKit", because it is Ruby with DevKit and not only Ruby.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1306)


ghost (approved):
Validation has completed

ItzLevvie (approved):

PR #1373, opened 2020-05-31
Add Composer-Setup 5.1.0

SoftCreatR:
Auto Generated by softcreatr@winget.it on Sun, 31 May 2020 04:06:00 +0000

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1373)


KevinLaMS:
Install fails with the The File name you specified must be php.exe

SoftCreatR:
> Install fails with the The File name you specified must be php.exe

That's correct, because PHP needs to be installed on the target system and without it, you can't use Composer. Those who attempt to install it, should know this.

PR #1398, opened 2020-05-31
Add Sumatra PDF 3.2

ghost:
May throw an error if not installed from an elevated shell. The actual installation will work regardless.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1398)

### DIFF
--- a/manifests/SumatraPDF/SumatraPDF/3.2.yaml
+++ b/manifests/SumatraPDF/SumatraPDF/3.2.yaml
@@ -1,0 +1,19 @@
+Id: SumatraPDF.SumatraPDF
+Version: 3.2
+Name: Sumatra PDF
+Publisher: Sumatra PDF
+License: GPL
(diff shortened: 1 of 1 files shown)


mattiasmarka:
I just realized there is already an open PR #346 for this package, which is blocked by issue #189. What is curious though, is that this PR does not seem to be affected. All the tests are passing -- I'm not sure if that was true for PR #346 as well -- and it installs fine on my machine.

The hash is the same, so the installer has not been changed in the time between PR #346 and this PR. The only thing that is different are the switches supplied in the manifest.

I'll leave this PR open in the meantime as I'm not sure how to proceed, but feel free to close it if you see fit.

mattiasmarka:
Okay, here's an update. This PR seems still to have the same issue as the original PR #346. That being said, I'm not sure if it is the same as microsoft/winget-cli#189. Closing for now.

PR #1419, opened 2020-06-01
Add Atomic Tanks 6.5

latere-a-latere:
Requires Visual C++ 2010 Redist (x86) to run. Keeping as a draft until we can actually define dependencies, unless that's not a problem. Other than that, installs fine and runs in my dev environment.

Part of https://github.com/microsoft/winget-pkgs/issues/897

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1419)


KevinLaMS:
The app is failing to run due to missing MSVCR100.dll.

PR #1474, opened 2020-06-02
Add Remote Desktop Manager Free

Phyyl:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1474)


MarNicGit:
@Phyyl Validation is working as intended albeit a bit bad in it's feedback. The manifest is currently located in `manifests/Devolutions/2020.1.20.0.yaml` but the correct directory would be `manifests/Devolutions/RemoteDesktopManagerFree/2020.1.20.0.yaml`.

manifests/Devolutions/2020.1.20.0.yaml
```diff
@@ -0,0 +1,21 @@
+Id: Devolutions.RemoteDesktopManagerFree
+Name: Remote Desktop Manager Free
```
Berkeley-Reject, line 2:
```suggestion
Name: RemoteDesktopManagerFree
```
I think you should move the yaml file into the `manifests/Devolutions/RemoteDesktopManagerFree` folder and change the name field to `RemoteDesktopManagerFree` in order to pass the validation test.

xiaoyang-sde (changes requested):

manifests/Devolutions/2020.1.20.0.yaml
MarNicGit: The name and publisher fields are allowed to be different from the package name, the validation only checks if the ID matches the folder structure.

ghost (approved):
Validation has completed

PR #1505, opened 2020-06-02
Update ImageMagick 7.0.10 to -16

ath0mas:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1505)

### DIFF
--- a/manifests/ImageMagick/ImageMagick/7.0.10-16.yaml
+++ b/manifests/ImageMagick/ImageMagick/7.0.10-16.yaml
@@ -1,5 +1,5 @@
 Id: ImageMagick.ImageMagick
-Version: 7.0.10-14
+Version: 7.0.10-16
 Name: ImageMagick
 Publisher: ImageMagick
 License: Copyright (c) ImageMagick Studio LLC - derived Apache 2.0 license
@@ -10,6 +10,6 @@ Description: Use ImageMagick® to create, edit, compose, or convert bitmap image
 Homepage: https://imagemagick.org/
 Installers:
   - Arch: x64
-    Url: https://imagemagick.org/download/binaries/ImageMagick-7.0.10-14-Q16-x64-dll.exe
-    Sha256: 38C49C2420552A6F28CD67B164C8472CE0FE2BF44780E635B722383EEBAD07ED
+    Url: https://imagemagick.org/download/binaries/ImageMagick-7.0.10-16-Q16-x64-dll.exe
+    Sha256: DC96EE898DB8F0BF96A143FC5A6E777738B785FC8586416146EC9DEC71ABABE9
     InstallerType: Inno


MarNicGit:
@ath0mas You should add a new manifest for this new version instead of replacing the existing one! Winget will automatically grab the last version available in the repository, unless the user specifies that he wants an older version.

ath0mas:
@MarNicGit it's a real question for me, for an often updated package like ImageMagick, do we want to add any new sub-version or to update it?
from my experience sometimes maintaining it for **brew**, I am used to their way of doing = updating (see [here](https://github.com/Homebrew/homebrew-core/commits/master/Formula/imagemagick.rb))

MarNicGit:
@ath0mas All packages currently in the repo use separate manifests for each of their versions. It's not a whole lot of extra work either compared to updating existing manifests - Just copy the old one into a new file and adjust that file instead.

How does brew handle users that want to install older versions?

ath0mas:
@MarNicGit I don't know the exact logic, and I suppose it's an update oriented process: you can install ImageMagick (v7) or ImageMagick@6, and _brew update/upgrade_ manage updating for you after _install_ ; you can not install a previous "sub-version"
back to our question, and my suggestion for this kind of package, would be to have manifests for each of major.minor.patch version but not maintaining history of sub-versions ... ? 🤔

MarNicGit:
@ath0mas I'd just make a separate manifest for now, even for sub-versions, unless the maintainers disagree.

@samiranshah, what is MS's stance on this?

ath0mas:
@MarNicGit Thanks for your feedback! I'm going to do another PR with separate manifest and close this one ;)

As a followup, I think that having a large list of historical download links may be hard to maintain as those previous versions will possibly not be downloadable forever

MarNicGit:
> As a followup, I think that having a large list of historical download links may be hard to maintain as those previous versions will possibly not be downloadable forever

I fully agree! Considering that practically all but a select few applications in the repo aren't hosted by Microsoft themselves (unlike many other package managers) it's worth opening an issue over this to discuss how 'dead' versions can be discovered quickly.

ath0mas:
And in fact, ImageMagick does not keep much binaries other than the last one, so the update I suggest is needed: previous download url for -14 fails now

see https://imagemagick.org/download/binaries/
more versions are still available in releases/ or windows/releases/, but only archives (zip) not binaries (exe)

samiranshah:
As long as the previous versions work, you can add a new manifest for the latest version.
In this case since the previous version download fails, it is okay to replace the version and download link.
Ideally, we do want the yaml file name to reflect the complete version number.

@KevinLaMS do you want to add anything to this?

ghost (approved):
Validation has completed

PR #1544, opened 2020-06-03
Remove Mojang.Minecraft

jamierocks:
The package was broken as it used an always up-to-date URL, having
spoken to peterix (the developer of the launcher at mojang) - they
have no plans to have versioned URLs exist for Windows.

    [6:04 PM] peterix: no, and there will be no such thing
    [6:05 PM] peterix: the launcher doesn't need packaging on windows
    [6:05 PM] peterix: plus, there would be no point in doing that,
                       because it immediately updates
    ...
    [6:22 PM] peterix: that is not an official source of the launcher
    [6:22 PM] peterix: and it shouldn't be there
    [6:22 PM] peterix: under any circumstances

Therefore, as it will not be possible to maintain this package, this
commit removes it.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1544)


ghost (approved):
Validation has completed

PR #1556, opened 2020-06-03
Add Apache Tomcat

superusercode:
You can only verify that it is installed by checking the list of services in task manager 
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1556)


superusercode:
Looks like apache deletes outdated links, going to have to find a mirror.

KevinLaMS:
I cannot get this to install.  It gives error 2.

superusercode:
The installer works fine for me on my laptop and on my VM, but since the installer is not being consistent I will close this PR.

Issue #91, opened 2020-05-09
Range GET requests are not supported by the copy source


denelon:
It looks like we have a problem with Range GET requests. #85

sundhaug92:
Looks like an undocumented requirement. I get why it's there but I think it should be removed

stedaniels:
The winget cli supports range requests.
A manifest that specifies a Url that references a web server that doesn't support range requests/streams will correctly fail-over to download, as per https://github.com/microsoft/winget-cli/blame/cc7904cd23bf1cdb1178a9dd8e634edf82d51e68/src/AppInstallerCLICore/Workflows/InstallFlow.cpp#L124.

The [validation-pipeline.yaml](https://github.com/microsoft/winget-pkgs/blob/master/DevOpsPipelineDefinitions/validation-pipeline.yaml) is the pipeline that appears to be broken, as far as I can tell it's the '[Installers Scan](https://github.com/microsoft/winget-pkgs/blob/master/DevOpsPipelineDefinitions/validation-pipeline.yaml#L98)' task.  I'm not really familiar with Azure pipelines but it appears to be calling executables/functions that haven't been open sourced.  Otherwise I would have attempted to find and fix the problem and sent a PR.

An example of a manifest that passes local validation but fails the pipelines is HeidiSQL.

`PS C:\devel\winget\winget-pkgs> winget install -m manifests\AnsgarBecker\HeidiSQL\11.0.0.5919.yaml
Found HeidiSQL [AnsgarBecker.HeidiSQL]
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Downloading https://www.heidisql.com/installers/HeidiSQL_11.0.0.5919_Setup.exe
  ██████████████████████████████  22.1 MB / 22.1 MB
Successfully verified installer hash
Installing ...
Successfully installed!`

This can be found failing here: https://github.com/microsoft/winget-pkgs/pull/1914

The output from the 'Installers Scan' is here: https://dev.azure.com/ms/winget-pkgs/_build/results?buildId=90013&view=logs&j=c8020aa5-4aab-5881-691f-21986e02afd0&t=bade86d6-d2be-5ddd-c1b9-7d8db9c4a60d&l=41


I don't want to minimise the issue, obviously range requests have been around for quite a while, but we can't expect everyone to renovate their server infrastructure.  The winget cli supports failing over, so drop these failures down to warnings and move on?

floh96:
[...]

PR #1583, opened 2020-06-04
Add HeidiSQL

superusercode:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1583)

### DIFF
--- a/manifests/HeidiSQL/HeidiSQL/11.0.0.5919.yaml
+++ b/manifests/HeidiSQL/HeidiSQL/11.0.0.5919.yaml
@@ -1,5 +1,5 @@
 Id: HeidiSQL.HeidiSQL
-Version: 11.0
+Version: 11.0.0.5919
 Name: HeidiSQL
 Publisher: HeidiSQL
 License: GNU General Public License v2.0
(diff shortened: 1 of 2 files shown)


superusercode:
getting really sick of these copy source errors

SoftCreatR:
@superusercode Please use the full version number (11.0.0.5919) instead.

superusercode:
@SoftCreatR done, but I find it weird that they use different version number for their github releases and the actually executables. This wasn't me shortening it but rather trying to match the github tags.

SoftCreatR:
> @SoftCreatR done, but I find it weird that they use different version number for their github releases and the actual executables. This wasn't me shortening it but rather trying to match the github tags.

@superusercode Yeah, i know. Btw: Linking from Majorgeeks won't work. The hashes generated are temporary.

superusercode:
Until the copy error is fixed I will just use the official URL since I cannot find a reputable mirror

SoftCreatR:
@superusercode I've created another PR: https://github.com/microsoft/winget-pkgs/pull/1610

As you can see, it uses a proprietary download URL. This is just a proxy around #91. It downloads the given file from the given URL and returns it to the client in a "proper" way.

Looks like it's working as expected, so i'll try to take over some of the failing applications.

SoftCreatR:
Obsolete by https://github.com/microsoft/winget-pkgs/pull/1610

PR #1638, opened 2020-06-06
Update BraveBrowser to 1.10.86

xiaoyang-sde:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1638)


SoftCreatR (changes requested):
This is a pre-release, thus it shouldn't be provided as official version. As of now, the official version is 1.9.76.

PR #1640, opened 2020-06-06
Update Bandizip to 7.08

xiaoyang-sde:
Since Bandizip does not use seperate download URL for installers of different releases, the previous yaml file is not valid now.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1640)


ghost (approved):
Validation has completed

PR #1653, opened 2020-06-06
Add calibre 4.18.0

jessicarod7:
Added calibre 4.18.0.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1653)


KevinLaMS:
Greetings.
I could not change the manifest directly because it was uploaded as a binary.  :(   It looks like the Switches are conflicting with the InstallerType. 
The quick fix, is remove the switches in the manifest, as we handle MSI natively. 
I will open an issue on the dev that if MSI is the  InstallerType and the switches are added, they are conflicting.

ghost (approved):
Validation has completed

PR #1658, opened 2020-06-07
Add HeidiSQL Nightly 11.0.0.5999

SoftCreatR:
Auto Generated by softcreatr@winget.it on Sun, 07 Jun 2020 13:40:15 +0000

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1658)


SoftCreatR:
@KevinLaMS If you want, i can drop support for automatic submission of nightly builds, because i doubt, that any of them will pass the smartscreen scan.

KevinLaMS:
Greetings.  The Url path is not what I would expect.  It should come from the host server.
https://www.heidisql.com/installers/HeidiSQL_11.0.0.5919_Setup.exe

PR #1679, opened 2020-06-08
add JetBrains.Toolbox

VagyokC4:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1679)


manifests/JetBrains/Toolbox/1.17.7018.0.yaml
```diff
@@ -0,0 +1,16 @@
+Id: JetBrains.Toolbox
+Publisher: Jet Brains
```
splitt3r, line 2:
```suggestion
Publisher: JetBrains
```

They always use `JetBrains` (see https://www.jetbrains.com/company/).

splitt3r (changes requested):

SoftCreatR (changes requested):
Publisher Name needs to be fixed: `Jet Brains` => `JetBrains`

EDIT: Nvm. Too slow 😄

ghost (approved):
Validation has completed

MarNicGit:
@jamierocks All I'm reading is that you're making assumptions without actually testing out what you're claiming. A quick `winget install jetbrains.toolbox` shows that it installs perfectly fine without any user interaction whatsoever.

jamierocks:
I'm relying on https://github.com/KevinLaMS's original statement that this was an issue, this manifest included nothing new - so i would assume the same blocking-issue is still in place.

PR #1690, opened 2020-06-08
Add Android Studio 4.0.0.16

SoftCreatR:
Auto Generated by softcreatr@winget.it on Mon, 08 Jun 2020 12:10:21 +0000

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1690)


KevinLaMS:
The installer is failing with error 1223.

SoftCreatR:
> The installer is failing with error 1223.

@KevinLaMS I don't know, what's causing this. It seems to appear at the end of the installation, so the application itself installs fine.

I just found out, that the installer uses NSIS, However, changing the installerType to `nullsoft` doesn't fix the problem.

PR #1728, opened 2020-06-09
Add PostgreSQL 12.3

### LINKED ISSUE #1718: Add package for PostgreSQL
# Description of the new feature/enhancement

I'm not sure if this is the right format to submit this (the GitHub issue template didn't mention anything about "new package requests"), but here goes anyways:

Add a package for PostgreSQL such as [the one from Chocolatey](https://chocolatey.org/packages/postgresql).

# Proposed technical implementation details (optional)

Some details from the following link may be useful for implementation:

https://github.com/majkinetor/au-packages/blob/master/postgresql/tools/chocolateyInstall.ps1

latere-a-latere:
Resolves https://github.com/microsoft/winget-pkgs/issues/1718 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1728)

### DIFF
--- a/manifests/PostgreSQL/PostgreSQL/12.3.yaml
+++ b/manifests/PostgreSQL/PostgreSQL/12.3.yaml
@@ -1,0 +1,17 @@
+Id: PostgreSQL.PostgreSQL
+Version: 12.3
+Name: PostgreSQL
+Publisher: The PostgreSQL Global Development Group
+License: PostgreSQL License
+LicenseUrl: https://www.postgresql.org/about/licence/
+Tags: database
+Description:  PostgreSQL is a powerful, open source object-relational database system with over 30 years of active development that has earned it a strong reputation for reliability, feature robustness, and performance. 
+Homepage: https://www.postgresql.org/
+Installers:
+  - Arch: x64
+    Url: https://get.enterprisedb.com/postgresql/postgresql-12.3-1-windows-x64.exe
+    Sha256: 1555EAF89DEF457B87637CA2B82F8CACF81B3DA6A0862F5D2BCEE0B1091F4D5C
+    InstallerType: exe
+    Switches:
+      Silent: --mode unattended
+      SilentWithProgress: --mode unattended --unattendedmodeui minimal


luzfcb:
The PostgreSQL 12 packaged by EnterpriseDB not officially support Windows 10, only Windows Server 2019, Windows Server 2016 and Windows Server 2012 R2

https://www.enterprisedb.com/services-support/edb-supported-products-and-platforms#pg12
https://www.postgresql.org/download/windows/

Also, the installer has no option to add the command-line tools to the Windows PATH automatically. This fact makes it very difficult for newbies. On Linux and Mac, there is no need to worry about this because the installer already does it for you.

luzfcb:
The problem of environment variables can be partially solved by loading `c:\Program Files\PostgreSQL\12\pg_env.bat` however, `pg_env.bat` has an incompatibility with Windows 10 or Windows 10 has a bug when processing the `PATH` environment variable that contains quotes character.

#### With quotes:

image




#### Without quotes:
image


I asked EnterpriseDB about this "problem", because I don't know if Windows Server behaves differently, and the answer was that Windows 10 is not supported:


image

ghost (approved):
Validation has completed

karlhorky:
Thanks @MarNicGit @KevinLaMS!

karlhorky:
@luzfcb we've installed this [via Chocolatey](https://github.com/majkinetor/au-packages/blob/master/postgresql/tools/chocolateyInstall.ps1) on a number of Windows 10 systems for our (very often beginner-level) students at @upleveled.

The `$PATH` thing is not ideal, but run a few lines of shell scripting and it can be added to the PATH automatically:

https://github.com/upleveled/system-setup/blob/master/windows.md#user-content-postgresql

Aside from that, no compatibility issues with Windows 10.

However, if you have an unattended installer alternative that could work better, then maybe it's worth investigating.

I'm sure that @majkinetor would also be interested (https://github.com/majkinetor/au-packages).

PR #1770, opened 2020-06-10
Add foobar2000

Witchilich:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1770)


KevinLaMS:
Change of redirect must have changed the hash.

Restia666Ashdoll:
@KevinLaMS winget should just be able to pick Videohelp links like Scoop
https://github.com/lukesampson/scoop-extras/blob/master/bucket/foobar2000.json

denelon:
@EndlessBernkastel it seems they would prefer not to be included in the Windows Package Manager.

https://hydrogenaud.io/index.php?topic=119506.0

EndlessBernkastel:
@denelon I guess we should stick to trying to make Videohelp links work like Scoop. There does not seem to any comment from the dev, just an opinion from another dev.

PR #1771, opened 2020-06-10
Ghostscript : Change Installer from nullsoft to exe

Witchilich:
https://www.ghostscript.com/doc/current/Install.htm
> The installer is NSIS-based (see also Release.htm) and supports a few standard NSIS options: /NCRC disables the CRC check, /S runs the installer or uninstaller silently, /D sets the default installation directory (It must be the last parameter used in the command line and must not contain any quotes, even if the path contains spaces. Only absolute paths are supported).

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1771)


SoftCreatR (changes requested):
NSIS = Nullsoft

So no need to modify the installer type.

PR #318, opened 2020-05-19
Add JetBrains toolbox manifest

cfanoulis:
Adds a manifest for Jetbrain's [Toolbox](https://www.jetbrains.com/toolbox-app/) app.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/318)

PR #1786, opened 2020-06-10
Remove JetBrains Toolbox

jamierocks:
A JetBrains Toolbox package has been held up since GH-318, as it
doesn't seem to have an un-attend mode.

Subsequent prs continue to come up, due to a lack of research.
GH-1679 introduced nothing new, and frankly was a lesser manifest
than the one in GH-318.

My understanding, given that GH-1679 introduced no new toggles, is
that this issue is still present - and as such, the manifest should
be removed.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1786)


MarNicGit:
@jamierocks I just tried it out in a vm and it installs fine, completely unattended.

VagyokC4:
@jamierocks If you spent the 5 seconds to install it, you too would see it's fine.

jamierocks:
> @jamierocks If you spent the 5 seconds to install it, you too would see it's fine.

I rarely use a Windows machine, so this wouldn't be a 5 second job for me.

If @KevinLaMS was mistaken, then I guess I've been misinformed.

mimi89999:
@jamierocks AFAIK only JetBrains product installers require silent config files as explained in https://github.com/microsoft/winget-pkgs/pull/179 and https://github.com/just-install/registry/issues/115.

jamierocks:
Maybe this is an issue, maybe it isn't - there hasn't been a clear, or consistent, line from Microsoft.

But, they also don't seem to care - so in line with @kevinlams's lack of response, I'm closing this - not feeling the agro of keeping it open. Microsoft should review prs more stringently and consistently.

PR #1858, opened 2020-06-14
Add Tidal

OfficialEsco:
No direct version installer as far as i know..

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1858)


OfficialEsco:
no idea what's failing, cba troubleshooting since it works locally, also seems like Tidal have moved to the Microsoft Store because that version is 2.17

PR #1938, opened 2020-06-16
Fix/Update Intel Driver And Support Assistant 20.6.22.6

### LINKED ISSUE #1742: Conflicting Intel Driver & Support Assistant packages
There are two manifests in this repository for the Intel Driver & Support Assistant, that differ only by the capitalisation of "a" in the "and". This seems unnecessary and makes installing one or the other unnecessarily difficult, as you need to (a) spot that the distinction is the capitalisation and (b) go hunting to remind yourself what the syntax is for the case-sensitive selection. One or the other should probably be deleted, or possibly they should be merged.

As best I can tell, this is entirely accidental. One was added in #1165 by @NathanBnm, the other in #1214 by @Reset12138. 

I've no particular preference about which is preserved and which is maintained; the one from #1214 (IntelDriverandSupportAssistant, i.e. without the capital "A") specifies a more recent version, but as best I can tell they're both downloading the same installer. One has the argument `/silent`, the other `/S /NORESTART`, but that, the version and the installer hash seem to be the only substantive differences between the manifests.

NathanBnm:
This closes #1742 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1938)

### DIFF
--- a/manifests/Intel/IntelDriverandSupportAssistant/20.6.22.6.yaml
+++ b/manifests/Intel/IntelDriverandSupportAssistant/20.6.22.6.yaml
@@ -1,16 +1,19 @@
-Id: Intel.IntelDriverandSupportAssistant
-Version: 20.6.22.6
+Id: Intel.IntelDriverAndSupportAssistant
+Publisher: Intel Corporation
 Name: Intel® Driver & Support Assistant
-Publisher: Intel
-License: Copyright © 2019 Intel Corporation. All rights reserved.
-AppMoniker: Intel Driver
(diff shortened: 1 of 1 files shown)


OfficialEsco:
@KevinLaMS please approve this

denelon:
The hash has changed on this. Is there a direct URL pointing to a specific version of the installer?

OfficialEsco:
@denelon no there is not, which is why i created #2353 and #2354

OfficialEsco:
Manifest deleted in #2354

PR #1963, opened 2020-06-17
Fix JRE link to the correct one for x64

palmtree5:
The link in the current JRE manifest is for x86 but the manifest states the arch is x64. This changes the link and the SHA256 checksum to the correct values for the x64 version

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/1963)


ghost (approved):
Validation has completed

PR #2005, opened 2020-06-20
Update path

whoiskevinrich:
The URL for the package is no longer valid as the contents of the 'latest' folder at the target change each update. This change does not change the package or sha being downloaded, simply the path to correct target version.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/2005)


ghost (approved):
Validation has completed

PR #2046, opened 2020-06-22
EA Origin Client v10.5.72.41482

Vivacio:
Added the current Origin game launcher for installation

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/2046)

### DIFF
--- a/manifests/ElectronicArts/Origin/10.5.72.41482.Yaml
+++ b/manifests/ElectronicArts/Origin/10.5.72.41482.Yaml
@@ -1,0 +1,18 @@
+Id: ElectronicArts.Origin
+Publisher: Electronic Arts Inc.
+Name: Origin
+Version: 10.5.72.41482
+AppMoniker: origin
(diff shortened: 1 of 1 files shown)


KevinLaMS:
i am running into an error due to a dependency.  
The code execution cannot proceed because MSVCP140.dll was not found.  Reinstalling the program may fix this problem.

Vivacio:
Ah sorry there! Reminds me i need to work in setting up a redeployable workstation to do clean tests of installs so thanks for that! :) 

As i understand from the bot response, currently there is no option yet to mark dependencies or add them right? 

If so i'll close the pull request and look back into it when that is possible (since the launcher will most definitely have an update as well.)

Vivacio:
Closing the pull request and will update when the dependencies can be added to the manifest

PR #2100, opened 2020-06-24
Add CentBrowser versions  "4.2.10.171"

koumaza:
`winget install -m .\4.2.10.171.yaml` was Successfuly.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/2100)


koumaza:
Try to solve from here.
https://www.microsoft.com/en-us/wdsi/filesubmission

KevinLaMS:
while it mostly runs unattended, there is an initial dialog.  Any chance to suppress.

koumaza:
> while it mostly runs unattended, there is an initial dialog. Any chance to suppress.

Unfortunately there is no way to suppress that dialog.
\> When I check the exe, it seems that `--show-install-ui` is passed as an argument, but the reverse is not found.

ghost (approved):
Validation has completed

PR #2131, opened 2020-06-25
Add WSL2 Kernel Update 5.4.42

peter-dolkens:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/2131)

### DIFF
--- a/manifests/Microsoft/WSL2/Kernel/5.4.42.yaml
+++ b/manifests/Microsoft/WSL2/Kernel/5.4.42.yaml
@@ -1,4 +1,4 @@
-Id: Microsoft.WSL2Kernel
+Id: Microsoft.WSL2.Kernel
 Version: 5.4.42
 Name: WSL2 Kernel Update
 Publisher: Microsoft Corporation
--- a/manifests/Microsoft/WSL2/Kernel/5.4.42.yaml
+++ b/manifests/Microsoft/WSL2/Kernel/5.4.42.yaml
@@ -1,0 +1,12 @@
+Id: Microsoft.WSL2Kernel
+Version: 5.4.42
+Name: WSL2 Kernel Update
+Publisher: Microsoft Corporation
+Homepage: https://docs.microsoft.com/en-gb/windows/wsl/wsl2-kernel
+License: Copyright (C) Microsoft Corporation
+Description: The WSL 2 is a new version of the architecture in WSL that changes how Linux distributions interact with Windows. WSL 2 has the primary goals of increasing file system performance and adding full system call compatibility.
+Installers:
+  - Arch: x64
+    Url: https://wslstorestorage.blob.core.windows.net/wslblob/wsl_update_x64.msi
+    InstallerType: MSI
+    Sha256: f4833159d00077e160c4b87ca4c4066da0b20bec69331cd414a6cb62bd88de70


manifests/Microsoft/WSL2/Kernel/5.4.42.yaml
```diff
@@ -0,0 +1,12 @@
+Id: Microsoft.WSL2Kernel
+Version: 5.4.42
+Name: WSL2 Kernel Update
+Publisher: Microsoft Corporation
+Homepage: https://docs.microsoft.com/en-gb/windows/wsl/wsl2-kernel
+License: Copyright (C) Microsoft Corporation
+Description: The WSL 2 is a new version of the architecture in WSL that changes how Linux distributions interact with Windows. WSL 2 has the primary goals of increasing file system performance and adding full system call compatibility.
+Installers:
+  - Arch: x64
+    Url: https://wslstorestorage.blob.core.windows.net/wslblob/wsl_update_x64.msi
```
StefanScherer, line 10: Nice idea, I also would like to have an easier option to install the WSL 2 kernel.

But I doubt this would work over a longer time. This link is a floating link to the "latest" msi package, and soon the checksum will be invalid. Is there a static link to that specific kernel version?
That would make more sense to use here in a versioned package.

peter-dolkens: Unfortunately I couldn't find an official static link, but from the verbage on the official page (https://docs.microsoft.com/en-gb/windows/wsl/wsl2-kernel) it sounds like future updates are planned to be distributed via windows update.

I'm hoping that this is either:

* Short term fix until the initial kernel update is in windows update
* Relatively static, with future updates delivered exclusively via windows update

StefanScherer: Yes, I also looked around and couldn't find something. 
BTW: MSFT is also working on `wsl --install` command to improve the first time experience, adding the Windows features. I hope that we don't have to think of the kernel in near future.

peter-dolkens: @StefanScherer - Slightly off-topic, but while we're discussing checksum validation errors - is there a recommended remediation approach for this? (Google Beta packages suffer this for example)

StefanScherer: @peter-dolkens I agree, only the official sources make sense. I only can suggest to get in contact with the software vendor that a static link for each released package should be provided. 

A floating link (which is „static“ eg. user can bookmark it or docs team can use it in static docs website) is nice for some use-cases. 
But it‘s bad for automation like package managers. Vendors should provide both and every user is happy. 😊

craigloewen-msft:
I'm a Program Manager working on WSL and I'm jumping onto this PR to add some more context from the WSL side that may help. 

We're working on having the Linux kernel be delivered via Microsoft Update. If you're on the Windows Insiders Dev Channel (formerly known as the Fast Ring) then you can get these kernel updates right away. Our guidance will be for users to click `Check for Updates` in Windows Settings to get the latest kernel update. As of right now this is only on the dev channel, but within the next months this will be available to all Windows users. You can find more details [here](https://devblogs.microsoft.com/commandline/gpu-compute-wsl-install-and-wsl-update-arrive-in-the-windows-insiders-fast-ring-for-the-windows-subsystem-for-linux/). 

As well we're working on `wsl --install` as @StefanScherer mentioned which will make sure that the latest kernel is installed during the installation process. Having this as a package inside of the Windows Package manager could be confusing for some users (e.g: I thought I already had WSL installed?, etc.)

peter-dolkens:
> I thought I already had WSL installed?

Shouldn't be an issue, as winget doesn't know how to tell the difference between installed, and uninstalled packages as is right now ;)

I get what you're saying though, and am happy to keep this in our private fork of the pkg repository so we can reference the manifest manually until `wsl --install` arrives.

PR #605, opened 2020-05-21
Add Signal Desktop

filiptronicek:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/605)

PR #2155, opened 2020-06-27
Signal v1.34.3

OfficialEsco:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/2155)

### DIFF
--- a/manifests/Signal/Signal/1.34.3.yaml
+++ b/manifests/Signal/Signal/1.34.3.yaml
@@ -1,0 +1,27 @@
+# ================== General ==================
+Id: Signal.Signal
+Publisher: Signal
+Name: Signal
+Author: Signal
+Description: Signal is an encrypted communications application. It uses the Internet to send one-to-one and group messages, which can include files, voice notes, images and videos, and make one-to-one voice and video calls.
+AppMoniker: signal
+Tags: messaging, texting, encryption, security
+
+# ================== License ==================
+Homepage: https://signal.org/
+License: Copyright 2014-2020 Open Whisper Systems - GPLv3
+LicenseUrl: https://download.blender.org/release/GPL3-license.txt
+
+# ================= Installer =================
+MinOSVersion: 10.0.0.0
+#Channel: stable                # Channel is not supported in this preview (5/24/2020)
+Version: 1.34.3
+
+InstallerType: nullsoft
+Installers:
+  - Arch: x86
+    Url: https://updates.signal.org/desktop/signal-desktop-win-1.34.3.exe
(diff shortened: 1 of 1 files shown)


ghost (approved):
Validation has completed

bearmannl:
@OfficialEsco, is there a reason you went for the Signal.Signal package name instead of OpenWhisperSystems.Signal as per the Publisher.Package convention? Am considering deleting all 21 versions of the package and re-adding them with the correct naming convention, as I did for the Signal-Beta package, but would like to get your thoughts before I get started on 42 PR's to get it done...

OfficialEsco:
This is a oooooooooooold Package Identifier as you can see in https://github.com/microsoft/winget-pkgs/pull/605. Back then none mentioned Publisher as per Control Panel. So instead we used the website or simplified "Publisher". Publisher got fixed in https://github.com/microsoft/winget-pkgs/pull/12182 which is when we got the word that winget uses the control panel to match packages.

Just downloaded 1.0.35 to see if they actually did Publish under Signal but they never did.
image

`Signal.Signal` is sorta a "valid" PackageIdentifier as long as `Open Whisper Systems` do not publish multiple Softwares or are known as `Open Whisper Systems`.

Feel free to move the manifests (which i see you've stated with lmao), however recommend you to wait until i approve a Manifest in `OpenWhisperSystems.Signal` so you can easily make good manifests with `wingetcreate update "id" --version --url --submit`

bearmannl:
> This is a oooooooooooold Package Identifier as you can see in #605. Back then none mentioned Publisher as per Control Panel. So instead we used the website or simplified "Publisher". Publisher got fixed in #12182 which is when we got the word that winget uses the control panel to match packages.
> 
> Just downloaded 1.0.35 to see if they actually did Publish under Signal but they never did.
> image
> 
> `Signal.Signal` is sorta a "valid" PackageIdentifier as long as `Open Whisper Systems` do not publish multiple Softwares or are known as `Open Whisper Systems`.
> 
> Feel free to move the manifests (which i see you've stated with lmao), however recommend you to wait until i approve a Manifest in `OpenWhisperSystems.Signal` so you can easily make good manifests with `wingetcreate update "id" --version --url --submit`

Cool, that makes sense.

I get that this has some history and that it is 'sorta' valid, although I'm seeing mixed implementations in the repository at times, which is not only inconsistent but also confusing to newcomers who lack some of the history (case in point 👋🏻☺).

Call me crazy (for doing the work now) but it makes it easier to apply the naming convention in a uniform manner without exception. Imagine if Open Whisper Systems publishes an additional app a few years down the line, then you would have to do the same rework I just did, but with even more versions in the backlog...

Thanks for the reply Esco, very informative 👊🏻

PR #2301, opened 2020-07-04
Add Chocolatey 0.10.15

invertedburger:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/2301)


KevinLaMS:
Sorry for the delay.  I appreciate the work to make this installer.  But our goal is to ensure we are installing from the source.  I would like the installer to come from Chocolatey.  Would you be willing to share this installer wrapper with Chocolatey so they can own it?
Feel free to reach out to me so we can discuss.  kevinla@microsoft.com

Issue #1381, opened 2020-05-31
add gitlab runner


gocarlos:
add gitlab runner for windows

details here:
https://docs.gitlab.com/runner/install/windows.html

denelon:
Support for this seems to be blocked by https://github.com/microsoft/winget-cli/issues/194 or https://github.com/microsoft/winget-cli/issues/182.

PR #2594, opened 2020-07-23
Add gitlab-runner

zanseb:

-----
#1381

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/2594)


KevinLaMS:
There are 2 issues with this install.
1) you must be an admin.  by default it fails with exit code 1.
2) it seems to be an executable that runs, rather than a true installer.

zanseb:
@KevinLaMS, thanks for the feedback. Is it mandatory for the package to use a true installer, or is it planned to support such scenarios too?

denelon:
It should be covered under https://github.com/microsoft/winget-cli/issues/194 or https://github.com/microsoft/winget-cli/issues/182.

denelon:
@zanseb this PR is blocked by the .exe feature. It will also fail now as we have updated the manifest schema to v1.0. I'm not sure if you'd like to update the manifest or close this PR until we have support for standalone executables.

zanseb:
@denelon I am fine w/ both options. What would you prefer?

denelon:
I think you've already done some of the heavy lifting to get the meta-data for the installer. If you don't mind updating the schema (knowing that it will fail validation), the community would benefit. The issue will be properly tagged so we can validate it once we have support for standalone executables, and nobody would need to go through the effort to gather the meta-data and submit a duplicate entry. Ultimately, it's up to you :smile:

zanseb:
@denelon I rebased the PR and migrated the format. Thanks for the heads up :)

denelon:
@zanseb the executable is named "gitlab-runner-windows-386.exe". Would users execute "gitlab-runner-windows-386.exe" to run this program after it was installed?

I'm working on the specification for portable / standalone executables.
https://github.com/microsoft/winget-cli/issues/182

zanseb:
@denelon, yes that is possible. 

I am not sure whether this is in the scope of microsoft/winget-cli#182, but I think it would be nice to have a "rename"-option, as most users would probably like to run just `gitlab-runner.exe`. It is also stated in their documentation, that they suggest renaming the executable after the download. JM2C

denelon:
@zanseb we do have a section in the manifest for the "command" to execute. I'm thinking about the implications of renaming the executable. I don't know how often this would be the case. Can you share a URL where they suggest renaming the executable? I'm thinking about the moderation scenario where we need to have someone look at the manifest to verify it's accurate, and "makes sense". I'd like our moderators to chime in on this as well.

@ImJoakim, @ItzLevvie, @jedieaston, @KaranKad, @OfficialEsco, @quhxl

zanseb:
@denelon, sure thing. Here it is: https://docs.gitlab.com/runner/install/windows.html

> The following assumes you have renamed the binary to gitlab-runner.exe (optional).

I thought a bit about this problem and asked myself following questions:
* How should the consumer of this package know that the name is actually `gitlab-runner-windows-386.exe`?
* What happens if the actual exe name changes after a specific version?

KaranKad:
Being able to rename executable would definitely be helpful. Its convenient for cli apps, for example 32bit binary of [yt-dlp/yt-dlp](https://github.com/yt-dlp/yt-dlp) is called `yt-dlp_x86.exe`. Also, binary of [Martchus/tageditor](https://github.com/Martchus/tageditor) contains whole lot of information like `tageditor-qt6-3.6.0-x86_64-w64-mingw32.exe` which is inconvenient for its cli use. [godotengine/godot](https://github.com/godotengine/godot) is a similar case as well where executable is named `Godot_v3.4.2-stable_win64.exe`

ItzLevvie (dismissed):

denelon:
@zanseb we've just finished support for portable applications. This PR will need to be updated to the 1.2 schema, and it's possible another PR in flight may be adding this package.

manifests/g/Gitlab/gitlab-runner/13.2.0/Gitlab.gitlab-runner.yaml
```diff
@@ -0,0 +1,20 @@
...
+Publisher: Gitlab
+PackageName: gitlab-runner
+License: MIT
+LicenseUrl: https://gitlab.com/gitlab-org/gitlab-runner/-/blob/main/LICENSE
+PackageUrl: https://gitlab.com/gitlab-org/gitlab-runner
+Moniker: gitlab-runner
+ShortDescription: GitLab Runner is the open source project that is used to run your CI/CD jobs and send the results back to GitLab
+Installers:
+  - Architecture: x86
+    InstallerType: exe
```
Trenly, line 13:
```suggestion
    InstallerType: portable
```
Since this is just a binary, it should be installer type 'Portable'

manifests/g/Gitlab/gitlab-runner/13.2.0/Gitlab.gitlab-runner.yaml
```diff
@@ -0,0 +1,20 @@
...
+    InstallerUrl: https://gitlab-runner-downloads.s3.amazonaws.com/v13.2.0/binaries/gitlab-runner-windows-386.exe
+    InstallerSha256: 6E91FB9FBF0548BA45DDAEF64C1DFCD06CA98ADAAF7E261E26F1F2503CC88C6B
+    InstallerSwitches:
+        Silent: install
+        SilentWithProgress: install
+ManifestType: singleton
```
Trenly, line 19: Should be updated to the multi-manifest format

manifests/g/Gitlab/gitlab-runner/13.2.0/Gitlab.gitlab-runner.yaml
```diff
@@ -0,0 +1,20 @@
...
+ManifestVersion: 1.0.0
```
Trenly, line 20: Must be 1.2.0 to support loose binaries

Trenly (changes requested):

manifests/g/Gitlab/gitlab-runner/13.2.0/Gitlab.gitlab-runner.yaml
zanseb: I tired to split them. Have some trouble validating them locally (`winget validate`). Can you hint me a resource in which the multi-manifest format is described? I only found https://docs.microsoft.com/en-us/windows/package-manager/package/manifest?tabs=minschema%2Cinstaller-example#multiple-manifest-files

ItzLevvie:
Close with reason: Replaced by PR 70732;

Trenly:
Apologies for over-topping your PR; Given that portable applications are still so new, we are still trying to figure out best-practices for all of the metadata in them to control how the version and name of the package gets mapped into the Add & Remove Programs table. 

I've used the data you included here in #70733. Thank you for your contribution and help!

zanseb:
No worries!

PR #2652, opened 2020-07-28
Samsung USB Driver for Mobile Phones 1.7.31.0 is added.

moolwalk:

-----
Silent install is not supported.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/2652)


moolwalk:
Silent install is not supported. The failure that bot raised appears to be due to the installer needs to be run in Administrator privilege. Please move forward with this PR.

Jesse Ahn.

denelon:
@moolwalk it looks like there are a few challenges here. We currently don't accept PRs for packages without support for a silent install. Many users are attempting to automate their environments, and we don't have the ability to filter on packages that require an interactive install yet.
https://github.com/microsoft/winget-cli/issues/823

The feature to support the client requesting elevation for installers that don't do that on their own also hasn't been implemented.
microsoft/winget-cli#152

We've also updated to the manifest v1.0 schema.

Until all three of those issues have been addressed this package will not pass validation. I'm not sure if you would prefer to update the schema, and keep this PR open with the "Blocking-Issue" label so we can submit it once the issues are implemented, or close this PR.

PR #2946, opened 2020-08-07
Add xmediarecode

sarukiti:
XMediaRecode is useful encoder software.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/2946)


sarukiti:
Sorry,I don't know how to include a dependency in a package.
I would like to see if there is a package manifest that uses it, but if that doesn't work, I'll close it.

I used machine translation

PR #3231, opened 2020-08-25
Add manifest file for foobar2000 v1.5.5

Robin-Wils:
**I have read it. This is not my original work, but I used the publisher of foobar2000. It is possible that this is not allowed. Please reply something if this is the case.**
**The hash of the installer is incorrect according to winget. I don't know why.**
**The Emacs package also has this problem. I had to force install Emacs on my machine, because of that, maybe I should create a new issue for that.**

-----

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/3231)

### DIFF
--- a/manifests/Resolute/foobar2000/1.5.5.yaml
+++ b/manifests/Resolute/foobar2000/1.5.5.yaml
@@ -1,0 +1,17 @@
+Id: Resolute.foobar2000
+Version: 1.5.5
+Name: foobar2000
+Publisher: Resolute
+License: foobar2000 license
+LicenseUrl: https://www.foobar2000.org/license
+AppMoniker: foobar2000
+Tags: music, media player, audio, foobar, 2000
+Description: foobar2000 is an advanced freeware audio player for the Windows platform. Some of the basic features include full Unicode support, ReplayGain support and native support for several popular audio formats.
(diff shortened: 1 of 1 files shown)


Robin-Wils:
Oh, you have to sign with a Github account to accept the license. 

I don't know what to fill in though, since I don't have a company.

yigitemres:
@Robin-Wils 

> The hash of the installer is incorrect according to winget. I don't know why.

There is no way adding foobar2000 from official web site right now because of two reasons:

1) url prefix changes constantly gonna cause problem.
2) foobar2000 releases (not sure about this) new builds in very short time and this cause hash errors.

> The Emacs package also has this problem. I had to force install Emacs on my machine, because of that, maybe I should create a new issue for that.

Emacs official repo use same link for every build (for same version). Hash checks fails because of this. I'm gonna pr right now.

Robin-Wils:
Thanks for the reply, makes sense. 

I am gonna be honest, I moved to Chocolatey for now, because it supports more commands like update. I might give this another try when it gets into a Windows version (not insider build). It is a promising project, so I hope that it becomes a thing.

PR #3259, opened 2020-08-26
Update Chrome to 85

auanasgheps:
- [x] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/change?
- [x] Have you validated your manifest locally with `winget validate <manifest>`, where `<manifest>` is the name of the manifest you're submitting?
- [x] Have you tested your manifest locally with `winget install -m <manifest>`?

-----

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/3259)

PR #3350, opened 2020-08-31
Google Chrome version renamed to latest + update

OfficialEsco:
No reason to have version specific manifest

-----

Yes there is a conflict with #3259 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/3350)

### DIFF
--- a/manifests/Google/Chrome/latest.yaml
+++ b/manifests/Google/Chrome/latest.yaml
@@ -1,12 +1,13 @@
 # ================== General ==================
 Id: Google.Chrome
 Publisher: Google
-Name: Chrome
+Name: Google Chrome
 Author: Google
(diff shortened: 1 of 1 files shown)


KevinLaMS:
@OfficialEsco why are you putting latest rather than the version in this case?  unless we have daily updates, there is no guarantee for the app to be the latest?

OfficialEsco:
1. There is no reason for multiple manifest for a program that don't have specific version
2. Its easier to update one manifest rather than update and rename another manifest

The guarantee for the app to be the latest is that it will always download the latest, the hash will however not match unless someone keep an eye on it. You guys do not guarantee to download the version we specify so there is no difference.

PR #3373, opened 2020-09-01
Corsair iCUE v3.32.80

OfficialEsco:

-----

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/3373)

### DIFF
--- a/manifests/Corsair/iCUE/3.32.80.yaml
+++ b/manifests/Corsair/iCUE/3.32.80.yaml
@@ -1,0 +1,28 @@
+# ================== General ==================
+Id: Corsair.iCUE
+Publisher: Corsair
+Name: Corsair iCUE
+Author: Corsair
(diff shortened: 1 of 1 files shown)


KevinLaMS:
install seems to conflict with itself

image

OfficialEsco:
I need to investigate this later, i only open iCUE to check for updates, so i actually never run the installer.
This MIGHT be the same problem AnyDesk have, no idea if its isolated to the MSI installer,

The way Chocolatey fixes this is to uninstall AnyDesk MSI then run the installer, but the iCUE script just runs the installer with /qn, no idea if it works for them.

OfficialEsco:
Newer version submited, looks like you NEED to submit with silent switches?

PR #3513, opened 2020-09-12
add xmind 10.2.1

chwdy:
fix error

- [x ] Have you checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/change?
- [x ] Have you validated your manifest locally with `winget validate <manifest>`, where `<manifest>` is the name of the manifest you're submitting?
- [x ] Have you tested your manifest locally with `winget install -m <manifest>`?

-----

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/3513)

### DIFF
--- a/manifests/XMind/XMind2020/10.2.1.yaml
+++ b/manifests/XMind/XMind2020/10.2.1.yaml
@@ -1,0 +1,20 @@
+---
+
+Id: XMind.XMind2020
+Version: 10.2.1
+Name: XMind2020
+Publisher: XMind
+License: © 2020 XMind Ltd.
+AppMoniker: xmind
+Tags: xmind
+Description: A full-featured Swiss Army Knife for your brain.
+Homepage: https://www.xmind.net/xmind2020/
+Installers:
+  - Arch: x64
+    Url: https://dl3.xmind.net/XMind-2020-for-Windows-64bit-10.2.1-202007272308.exe
+    Sha256: 8FC326F4E39FDB7C82469E972BBA0D46790411A1F001D35C7096708C03EC9C04
+    InstallerType: exe
+    Switches:
+      Silent: /VERYSILENT /NORESTART
+      SilentWithProgress: /VERYSILENT /NORESTART
+


manifests/XMind/XMind 2020/10.2.1.yaml
```diff
@@ -0,0 +1,20 @@
+---
```
yigitemres, line 1:
```suggestion

```
When I copy paste this manifest, this part contains "TAB".  Either delete tab or delete whole line.

```
    Why does YAML forbid tabs?

    Tabs have been outlawed since they are treated differently by different editors and tools.
    And since indentation is so critical to proper interpretation of YAML, this issue is just too
    tricky to even attempt.  Indeed Guido van Rossum of Python has acknowledged that allowing TABs
    in Python source is a headache for many people and that were he to design Python again, he
    would forbid them.
```

manifests/XMind/XMind 2020/10.2.1.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+Id: XMind.XMind 2020
```
yigitemres, line 3:
```suggestion
Id: XMind.XMind2020
```

or 

```suggestion
Id: XMind.XMind-2020
```

or 

```suggestion
Id: XMind.XMind_2020
```

winget validate throws error. When you change this part, don't forget to change file path.

```
┌[yigitemres]
└[~\gitrepos]> winget validate --manifest .\xmind.yaml
Manifest validation failed.
Manifest Error: Invalid field value. Field: Id Value: XMind.XMind 2020 Line: 3, Column: 5
```

yigitemres (changes requested):

yigitemres:
@chwdy Thanks for you contribution and have a nice day! 🥂

DW-42 (dismissed):

manifests/XMind/XMind 2020/10.2.1.yaml
chwdy: Thanks for your suggestion, I am confusing about how to name it actually.
Long time ago, xmind company had a classic mind map called xmind8, which is the original "xmind"
Then they have an updated version, which was a big change, and they called it xmind zen. they use "xmind zen" for at least two years.
and now it becomes xmind 2020 on their website. However, on the start menu, it only shows "xmind".
I am thinking "Xmind 2020" is too long and it is different from the start menu. but "xmind" might make old users confused.
Any suggestion for the name? feel free to directly modify my commit.
I would prefer the third suggestion. Since the id must be the same as the folder, I might have to make another PR later

yigitemres: You don't have to do another pr. Just rename the file and push it with proper id.

> Any suggestion for the name? 

If we talk about `Name` part, I think that "XMind 2020" is correct name for the app. If we talk about `Id` part it has to be something differentiate it from `XMind8` to `XMind2020` (because XMind8 can be added by anyone). We have several options for this:

1) If we name XMind2020 as XMind
   1) We have to name XMind8 as XMind8

2) If we name XMind2020 as XMind2020 (or similar)
   1) We can name XMind8 as XMind
   2) We can name XMind8 as XMind8

I think that explicitly indicating the app name is good choice for every manifest. If I create this file, I'm gonna use `XMind2020` path.

PR #3841, opened 2020-10-01
​

ItzLevvie:
​


OfficialEsco (dismissed):
@KevinLaMS You need to fix your system.
Just tested because it looks correct, and it installed like it should. You need to leave a reason when the bot says it works.

ItzLevvie:
It installed fine on my machine.

Before:
image

After:
image

PR #4355, opened 2020-10-21
First request to add CIB pdf brewer (3.4.1)

pdfbrewer:

-----

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/4355)

### DIFF
--- a/manifests/CIBsoftwareGmbH/CIBpdfbrewer/3.4.1.yaml
+++ b/manifests/CIBsoftwareGmbH/CIBpdfbrewer/3.4.1.yaml
@@ -20,5 +20,3 @@ Installers:
     Switches:
       Language: /LANG=<LANGUAGE>
       Log: /LOG="<LOGPATH>"
-      Silent: /VERYSILENT
-      SilentWithProgress: /SILENT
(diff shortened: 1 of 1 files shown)


KevinLaMS:
This app will not install silently unless a license is available

pdfbrewer:
> 
> 
> This app will not install silently unless a license is available

@KevinLaMS thank you for your time. Would it be possible to submit it without silent installation? If not, would it be possible to only submit silent installation with progress? Current submission will not work, winget still feeds silent parameters to the setup.

JanPokorny:
@pdfbrewer `winget` requires that the installers finish autonomously (meaning that a window can appear, but it can't require any input), which is usually what "silent mode" means. Since you seem to be the author of the app, I advise you to fix your installer's silent mode so it actually installs without user interaction.

pdfbrewer:
> 
> 
> @pdfbrewer `winget` requires that the installers finish autonomously (meaning that a window can appear, but it can't require any input), which is usually what "silent mode" means. Since you seem to be the author of the app, I advise you to fix your installer's silent mode so it actually installs without user interaction.

OK I see, thank you for your reply.

Issue #4360, opened 2020-10-21
Node.js package versions with `Switches` field are all broken


saschanaz:
# Brief description of your issue

For example, installing Node.js 14.14 throws:

```
> winget install Node.js -v 14.14.0
Found Node.js [OpenJS.NodeJS]
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Downloading https://nodejs.org/dist/v14.14.0/node-v14.14.0-x64.msi
  ██████████████████████████████  29.1 MB / 29.1 MB
Successfully verified installer hash
Starting package install...
Installer failed with exit code: 1603
Installer log is available at: C:\Users\Kagami\AppData\Local\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\TempState\WinGet\OpenJS.NodeJS.14.14.0.log
```

... while installing 14.12 is okay:

```
> winget install Node.js -v 14.12.0
Found Node.js [OpenJS.NodeJS]
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Downloading https://nodejs.org/dist/v14.12.0/node-v14.12.0-x64.msi
  ██████████████████████████████  29.1 MB / 29.1 MB
Successfully verified installer hash
Starting package install...
Successfully installed
```

# Steps to reproduce

* `winget install Node.js -v 14.14.0`

# Expected behavior

It should succeed

# Actual behavior

`Installer failed with exit code: 1603`

# Environment

```none
> winget --info
Windows Package Manager v0.2.2521 Preview
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.19042.572
Package: Microsoft.DesktopAppInstaller v1.11.2521.0

Links
--------------------------------------------------------
Privacy Statement   https://aka.ms/winget-privacy
License Agreement   https://aka.ms/winget-license
Third Party Notices https://aka.ms/winget-3rdPartyNotice
Homepage            https://aka.ms/winget
```

Pinging contributors who add `Switches`: @jedieaston @ItzLevvie

jedieaston:
I'm happy to fix the manifests, but I'm unable to reproduce this...

```
> winget install Node.js -v 14.14.0                                             
Found Node.js [OpenJS.NodeJS]
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
[...]

PR #4376, opened 2020-10-21
​

ItzLevvie:
​

### DIFF
--- a/manifests/OpenJS/NodeJS/15.0.1.yaml
+++ b/manifests/OpenJS/NodeJS/15.0.1.yaml
@@ -1,0 +1,15 @@
+Id: OpenJS.NodeJS
+Name: Node.js
+AppMoniker: nodejs
+Version: 15.0.1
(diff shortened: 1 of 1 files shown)


jedieaston:
Would you mind removing the switches? They cause issues (see #4360) if you aren't running the installer as admin, and since it's an MSI anyway, WinGet already runs it silently by default.

KevinLaMS:
@ItzLevvie  What do you think about removing the switches on the PR?  By removing them the MSI will pop the UAC prompt for a better experience.  @jedieaston just tweaked the previously submitted PRs.  
https://github.com/microsoft/winget-pkgs/pull/4372

ItzLevvie:
> @ItzLevvie What do you think about removing the switches on the PR? By removing them the MSI will pop the UAC prompt for a better experience. @jedieaston just tweaked the previously submitted PRs.
> #4372

Done.

PR #4381, opened 2020-10-22
​

ItzLevvie:
​


KevinLaMS:
We have found a better user experience if we remove the switches on the MSI installertype.  When the switches are there, when run in non-admin command prompt it will error 1603.  If you remove the switches the UAC is not suppressed and the MSI can be installed.
Would you mind changing that?

ItzLevvie:
> @ItzLevvie We have found a better user experience if we remove the switches on the MSI installertype. When the switches are there, when run in non-admin command prompt it will error 1603. If you remove the switches the UAC is not suppressed and the MSI can be installed.
> Would you mind changing that?

Done.

Issue #3197, opened 2020-08-21
POLL - Please Respond


denelon:
# Question

If a package manifest has a mistake like missing keys or incorrect value(s) and we can automatically correct it, what should we do?

Please chose an answer and +1

denelon:
From twitter: Another option: use the github code review API to make a diff suggestion, mimicking a human reviewer. That suggestion can be applied and committed by the author.

PR #4518, opened 2020-10-29
Add PeaZip 7.4.2.0

philip-khor:

-----

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/4518)


manifests/GiorgioTani/PeaZip/7.4.2.0.yaml
```diff
@@ -0,0 +1,17 @@
...
+LicenseUrl: https://www.peazip.org/peazip-tos-privacy.html#software_copyright
+AppMoniker: peazip
+Tags: zip, compression, peazip, giorgio, tani, files, compressed
+Description: PeaZip is a free and open-source file manager and file archiver for Microsoft Windows, Linux and BSD made by Giorgio Tani. It supports its native PEA archive format and other mainstream formats, with special focus on handling open formats. It supports 210 file extensions.
+Homepage: https://www.peazip.org/
+Installers:
+  - Arch: x86
+    InstallerType: MSI
+    Url: https://github.com/peazip/PeaZip/releases/download/7.4.2/peazip-7.4.2.WIN64.msi
+    Sha256: 7950DBC21542142DB94F60B8AD96766047E42480558BAA6E470EF0D13AC40C8E
```
yigitemres, line 16:
```suggestion
    Sha256: 7950DBC21542142DB94F60B8AD96766047E42480558BAA6E470EF0D13AC40C8E
    Switches:
      Silent: /quiet
      SilentWithProgress: /passive
```

yigitemres (dismissed):
It is good practice to adding switches even for "msi" installers.

manifests/GiorgioTani/PeaZip/7.4.2.0.yaml
ItzLevvie: The switches for MSI installers are not needed (some references)
1) https://github.com/microsoft/winget-pkgs/pull/4376#issuecomment-714097188
2) https://github.com/microsoft/winget-pkgs/issues/4360#issue-726661418

yigitemres: I know its not needed but it prevents msfbot go crazy about Switches and it creates habit about checking them for every manifest with or without MSI.

ItzLevvie: >I know its not needed but it prevents msfbot go crazy about Switches

Kevin adds the "Validation-Unattend-Failed" label to a PR whenever an installer cannot run in unattended mode, hence the reason why the msftbot replied about the switches.

>it creates habit about checking them for every manifest with or without MSI.

winget will automatically install MSI installers silently, so the only headache there is for including switches in manifest for MSI installers is, that everyone would have to run their Command Prompt, or PowerShell, as an Administrator, which is not ideal for most people.

Anyways, PeaZip no longer allows their installers to be ran in unattended mode since their own 7.4.x releases, therefore, the switches will be completely ignored in the manifest, even if it were to be included.

yigitemres: > Kevin adds the "Validation-Unattend-Failed" label to a PR whenever an installer cannot run in unattended mode, hence the reason why the msftbot replied about the switches. 

I see the label and I think that this is false positive. I'm pretty sure that I can find false positives about how msfbot going crazy (about anything?) but no need for that. For this pr, I also think that msfbot going crazy and complains `Switches` even don't need for that but PeaZip itself is the reason.

> Anyways, PeaZip no longer allows their installers to be run in unattended mode, since their own 7.4.x releases, therefore, the switches will be completely ignored, even if it were to be included in the manifest.

If the submitters test their manifests with not only `validate` but also with `install`, this pr is never existed. Because of this reason adding `Switches` (even for MSI) and than do the checks before pr submit is good habit in order to detect installation problems. Yesterday, I tried to push nmap v7.90 but when I creating manfiest file, I find that only nmap 7.90 oem supports switches.

> winget will automatically install MSI installers silently,

As I said previously, I know that. I don't know that when winget starts to do that for MSI but I have a suggestion about it: https://github.com/microsoft/winget-pkgs/issues/3197#event-3720532886

> that everyone would have to run their Command Prompt, or PowerShell, as an Administrator to install x application, which is not ideal for most people out there.

If the user lacks the privileges in order to install apps, I believe that with or without the `Switches` is still gonna need the same exact privileges (otherwise this is security vulnerability) because nearly all of the switches is only about silent install (and silent install switches generally don't need any privileges if the installer not configured strangely). XInstaller.msi may need removal of the `Switches` but this is about how XInstaller.msi is configured.

Have a nice day! 🥂

yigitemres:
@philip-khor Hi there. Firstly, thanks for your contribution.

There is a problem (I think this is a problem maybe its intended) about Peazip MSI. When passing /passive or /quiet arguments usually installer should bypass the EULA agreements. But in this installer, it's continue to asking. I believe that this is not intentional because if you download exe installer (inno) /silent argument bypass the EULA.

This pr never pass under this conditions. It will be good if you commit exe installer version with new switches and hash.

Have a nice day! 🥂

philip-khor:
Thanks for the input @yigitemres, will open a new PR for 7.5.0 exe if I find the time. Have a nice day too!

PR #4834, opened 2020-11-16
Update 8.0.61001.yaml to be more consistent with other Microsoft .YAML files

DW-42:
- [✔] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)?
- [✔] Have you checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/change?
- [✔] Have you validated your manifest locally with `winget validate <manifest>`, where `<manifest>` is the name of the manifest you're submitting?
- [✔] Have you tested your manifest locally with `winget install -m <manifest>`?

-----

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/4834)

### DIFF
--- a/manifests/Microsoft/VC++2005Redist-x86/8.0.61001.yaml
+++ b/manifests/Microsoft/VC++2005Redist-x86/8.0.61001.yaml
@@ -1,8 +1,8 @@
 Id: Microsoft.VC++2005Redist-x86
 Version: 8.0.61001
 Name: Microsoft Visual C++ 2005 Redistributable (x86)
-Publisher: Microsoft
-Homepage: https://www.microsoft.com/en-au/download/details.aspx?id=26347
+Publisher: Microsoft Corporation
+Homepage: https://www.microsoft.com/en-us/download/details.aspx?id=26347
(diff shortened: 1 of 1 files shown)


KevinLaMS:
the URL should probably be en-us instead of en-au
 https://www.microsoft.com/en-au/download/details.aspx?id=29

Also, the publisher seems fine: Microsoft Corporation.

DW-42:
> the URL should probably be en-us instead of en-au
> https://www.microsoft.com/en-au/download/details.aspx?id=29
> 
> Also, the publisher seems fine: Microsoft Corporation.

I noticed that some of the manifest files had "Publisher: Microsoft Corporation" and some had "Publisher: Microsoft" and it looked to me like less work to change them all to "Publisher: Microsoft" - but KevinLaMS is right. It would easier for someone/script/robot to filter the data to get "Microsoft" than to extrapolate "Corporation", so "Microsoft Corporation" wins.

PR #5036, opened 2020-11-26
Added Blender 2.91.0

rmarquis:

-----

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/5036)

### DIFF
--- a/manifests/BlenderFoundation/Blender/2.91.0.yaml
+++ b/manifests/BlenderFoundation/Blender/2.91.0.yaml
@@ -1,0 +1,29 @@
+# ================== General ==================
+
+Id: BlenderFoundation.Blender
+Version: 2.91.0
+Name: Blender
+Publisher: The Blender Foundation
+Description: Blender is the free and open source 3D creation suite. It supports the entirety of the 3D pipeline—modeling, rigging, animation, simulation, rendering, compositing and motion tracking, video editing and 2D animation pipeline.
+Homepage: https://www.blender.org/
+
+# ================== Search ===================
+
+AppMoniker: blender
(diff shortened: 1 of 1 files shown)


manifests/BlenderFoundation/Blender/2.91.0.yaml
```diff
@@ -0,0 +1,29 @@
...
+# ================== License ==================
+
+License: GPL 3.0 or later
+LicenseUrl: https://download.blender.org/release/GPL3-license.txt
+
+# ================= Installer =================
+
+Installers:
+  - Arch: x64
+    Url: https://mirror.clarkson.edu/blender/release/Blender2.91/blender-2.91.0-windows64.msi
```
Marsfan, line 24: Why are you pointing to a mirror instead of the default installer URL of 
https://www.blender.org/download/Blender2.91/blender-2.91.0-windows64.msi/
which then auto selects a mirror?

marsfan (commented):
Quick question on mirror that is used

manifests/BlenderFoundation/Blender/2.91.0.yaml
rmarquis: Because it doesn't work :(
Since it's a redirector, the hashfile won't match.

As such, I'm using the same strategy of the initial manifest author, which is to select a mirror manually.

Marsfan: Ah. That would explain it.

PR #5815, opened 2020-12-31
​

ItzLevvie:
​


KevinLaMS:
Not sure that the URL is correct

image

ItzLevvie:
cc @KevinLaMS WeMod seems to be installing fine here:
image

PR #5816, opened 2020-12-31
Remove WireGuard temporarily

zx2c4:
@KevinLaMS - let's see if the CI complains about this. If it doesn't we'll submit it, I'll submit the fix, and then we'll be set. If it does complain, then I'll submit 5 PRs.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-pkgs/pull/5816)


zx2c4:
Nope, looks like it failed. 5 PRs here we come.

PR #6801, opened 2021-01-29
​

ItzLevvie:
​


KevinLaMS:
I do not know about this app, however I see it fail to launch in my VM

image

KevinLaMS:
PS. @ItzLevvie   I want to thank you for updating the URLs.
I think I am going to be a pain, and start rejecting any HTTP URLs.  That will help us going forward.  Then URLs that are not HTTPS will be the exception.

PR #7168, opened 2021-02-09
add SabakiHQ.Sabaki v0.51.1

apetresc:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/7168)


apetresc:
Whoops. I thought we did support multiple installers for different architectures. `winget validate` didn't notice the problem either, even though it would be trivial for a static analysis tool to notice this...

I've removed the x86 one and left the x64 one for now.

PR #7232, opened 2021-02-10
replayweb-page 1.3.8

oxygen-dioxide:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/7232)

### DIFF
--- a/manifests/webrecorder/replayweb-page/1.3.8.yaml
+++ b/manifests/webrecorder/replayweb-page/1.3.8.yaml
@@ -1,0 +1,27 @@
+# ================== General ==================
+Id: webrecorder.replayweb-page
+Publisher: webrecorder
+Name: replayweb-page
+Author: webrecorder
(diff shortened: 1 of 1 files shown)


KevinLaMS:
Appears to use a proxy: https://ghproxy.com/

KevinLaMS:
@oxygen-dioxide can you share why you were using the proxy?  Is \ was it providing some service I am not familiar with?
@denelon

oxygen-dioxide:
> @oxygen-dioxide can you share why you were using the proxy? Is \ was it providing some service I am not familiar with?
> @denelon

Because github downloads very slow in mainland China without this proxy.

KevinLaMS:
@oxygen-dioxide Thanks for clarifying.  I will not reject anymore of these and investigate if we can approve with the proxy.

PR #7420, opened 2021-02-13
​

ItzLevvie:
​

### DIFF
--- a/manifests/Golang/Go/1.15.7.yaml
+++ b/manifests/Golang/Go/1.15.7.yaml
@@ -10,6 +10,6 @@ Description: Go is an open source programming language that makes it easy to bui
 Homepage: https://golang.org
 Installers:
   - Arch: x64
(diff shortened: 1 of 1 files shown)


SeparateRecords:
Hi @ItzLevvie, I'm curious, what the reasoning was to change the download URLs back? The reason to change them to golang.org/dl/ was in [the PR to add 1.15.7](https://github.com/microsoft/winget-pkgs/pull/6596#issue-561472463). Was the redirect causing problems?
Thanks in advance! :)

ItzLevvie:
The PR was made because I made a script to fix some of the 302 Redirect, rather than fixing URLs that redirected multiple times, I decided to fix most of it. There were no issues with downloading those programs that had URLs which were redirecting.

though recently, winget is having some issues now with URL redirects (e.g. it doesn't like it when a redirect changes from HTTPS to HTTP).

PR #7566, opened 2021-02-17
​

ItzLevvie:
​


KevinLaMS:
Installation was blocked with dialog:  "i cannot find an installation of tux paint.  Would you like to install anyway?

PR #7664, opened 2021-02-18
​

ItzLevvie:
​


KevinLaMS:
the updated URL is a MSI.  but the switches are still included and this is causing an install issue.  I am seeing the MSIexec error dialog.

PR #7674, opened 2021-02-18
​

ItzLevvie:
​


KevinLaMS:
may just be a bad return code

image

jedieaston:
See plexinc/plex-media-player#992 (they haven't fixed it but know about it)

PR #7821, opened 2021-02-21
​

ItzLevvie:
​


chausner:
I think we should revert back to version 7.16.11. This is the officially recommended latest version for Windows on https://boinc.berkeley.edu/download.php. The slightly newer version 7.16.16 is not linked anywhere on the main site and there are reports that there are problems with this version: https://groups.google.com/a/ssl.berkeley.edu/g/boinc_alpha/c/JS5-Fw-mcC4.

PR #8147, opened 2021-02-28
​

ItzLevvie:
​


KevinLaMS:
Winget fails to install with 3010.  Automation fails to install and reports  -1978335226

PR #8338, opened 2021-03-03
​

ItzLevvie:
​


KevinLaMS:
Hey @ItzLevvie 
a bunch of these are showing up with ZERO bytes and no diff.
Is there an issue with your script?

kevin

ItzLevvie:
cc @KevinLaMS: It changes the encoding from `UTF-16LE` to `UTF-8`; but for some reason GitHub shows it's an "Empty file" even though it's not.

Clicking on the "View file" button on in https://github.com/microsoft/winget-pkgs/pull/8338/files shows that there are 595 bytes in this file.

PR #8822, opened 2021-03-16
​

ItzLevvie:
​


jedieaston:
It looks like this is a newer version of Microsoft.Teams (the URL scheme is the same), should it just go under that?

ItzLevvie:
> It looks like this is a newer version of Microsoft.Teams (the URL scheme is the same), should it just go under that?

It's a different ring, Production is still on 1.4.00.4167 (from https://raw.githubusercontent.com/ItzLevvie/MicrosoftTeams-msinternal/master/defconfig)

PR #9007, opened 2021-03-19
Yubico.Authenticator-5.0.4

JGoutin:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/9007)

### DIFF
--- a/manifests/Yubico/Authenticator/5.0.4.yaml
+++ b/manifests/Yubico/Authenticator/5.0.4.yaml
@@ -1,0 +1,17 @@
+Id: Yubico.Authenticator
+Version: 5.0.4
+Name: Yubico Authenticator
+AppMoniker: yubioath
+Tags: 2fa, security, auth, totp
+Description: Authenticator to generate 2-step verification codes using your YubiKey
+Homepage: https://www.yubico.com/products/yubico-authenticator/
+Publisher: Yubico
+License: BSD-2-Clause License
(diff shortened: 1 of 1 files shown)


JGoutin:
I noticed than some package seems to be automatically updated. But don't find documentation about this feature.

This software is also available trough the following URL that always redirect to the latest version: https://developers.yubico.com/yubioath-desktop/Releases/yubioath-desktop-latest-win64.msi

Setting this URL in the yaml will enable auto update ? or this is more complicated.

denelon:
JGoutin, in general we prefer explicit URLs to specific versions of packages. If the "vanity" URL is used, then the manifest will fail for users when the publisher changes it out from underneath us. We do have automation to check for this event, but it still results in failed installs for users when the SHA256 hash doesn't match until we detect it.

JGoutin:
Thanks for the answer.

This is reasonable to not break the installation.

To keep the software up to date without manual update, it may be interesting to have an extra parameter in the manifest that the bot may check to see if there is an update available and update the manifest accordingly. In the current case, this parameter value should be the "vanity" URL (Which is simply a web redirection to the versioned installer). But this is an easy case, some case may be more complex like relying on GitHub releases. And this is not possible at all for many software.

PR #9147, opened 2021-03-22
​

ItzLevvie:
​


KevinLaMS:
see 9146 for an image of the same type of error i am getting when trying to install this app.  seems like the URL is not accessible.

ItzLevvie:
cc @KevinLaMS: It seems to be downloading fine on my system:
image

PR #9353, opened 2021-03-29
Update SparkLabs.Viscosity to 1.9.2

trametheka:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/9353)

### DIFF
--- a/manifests/SparkLabs/Viscosity/1.9.2.yaml
+++ b/manifests/SparkLabs/Viscosity/1.9.2.yaml
@@ -1,0 +1,19 @@
+Id: SparkLabs.Viscosity
+Version: 1.9.2
+Name: Viscosity
+Publisher: SparkLabs Pty Ltd
+License: Copyright © SparkLabs Pty Ltd
+LicenseUrl: https://www.sparklabs.com/company/legal/
+AppMoniker: viscosity
+Tags: viscosity, openvpn, vpn
+Description: A first class OpenVPN client that lets you secure your network with ease & style.
+Homepage: https://sparklabs.com/viscosity/
+Installers:
+  - Arch: Neutral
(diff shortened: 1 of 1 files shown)


trametheka:
Silent and SilentWithProgress switches are in there and validated.

trametheka:
Just as a follow up to this as an FYI, unless the manifest changes have happened since I made this pull request, the manifest is correct.

I think the issue is the build server is missing either .NET 4.8 or the VC redistributable and our installer is not correctly running those silently, or, a reboot is required after installing one of those pre-requisites. I will make some changes to the installer and resubmit.

@Suirtimed , @denelon , are you able to confirm what the build/test servers have in the way of .NET and VC redist installed? Cheers.

denelon:
@trametheka the Windows Package Manager doesn't support dependencies yet. Does your installer include them? Our validation environment is essentially a base OS without any additional libraries.

trametheka:
@denelon It doesn't include them, but will download and install them if missing. I'll run on the assumption it's getting stuck because the dependency installs are throwing dialogs when the installer is run silently and make some changes. Thanks for the reply.

denelon:
Screenshot 2021-04-21 200324

trametheka:
@denelon Is there a way to force winget to run the install process in verysilent mode? I can't see anything in the help output of winget.

denelon:
I believe we're defaulting to "Silent" for the validation server. You can pass flags for that install mode.

It looks like the manifest has:

Silent: /VERYSILENT /SP- /NORUN

arjun-msft:
@trametheka try changing the SilentWithProgress to be VERYSILENT as well. I think it installs the dependencies with that flag.

PR #10091, opened 2021-04-05
Add mIRC latest

chausner:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/10091)

### DIFF
--- a/manifests/m/mIRC/mIRC/7.64/mIRC.mIRC.yaml
+++ b/manifests/m/mIRC/mIRC/7.64/mIRC.mIRC.yaml
@@ -1,0 +1,29 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.singleton.1.0.0.schema.json
+PackageIdentifier: mIRC.mIRC
+PackageVersion: "7.64"
+PackageName: mIRC
+PackageUrl: https://www.mirc.com/
+Publisher: mIRC Co. Ltd.
+PublisherUrl: https://www.mirc.com/
+PublisherSupportUrl: https://www.mirc.com/help.html
+PrivacyUrl: https://www.mirc.com/privacy.html
+License: Proprietary
+LicenseUrl: https://www.mirc.com/eula.html
+Copyright: Copyright © 1995-2020 mIRC Co. Ltd.
+ShortDescription: IRC (Internet Relay Chat) client
+Description: >
+  mIRC is a popular Internet Relay Chat client used by individuals and organizations to
+  communicate, share, play and work with each other on IRC networks around the world. Serving the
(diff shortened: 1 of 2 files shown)


KevinLaMS:
This was failing... restarting to new baseline

chausner:
I am not sure why installer validation fails in CI, there are no useful error messages in the logs. Installation succeeds when I test locally.

OfficialEsco:
Not sure if this is what Kevin is referering to but this is what my VM gets
image

Also the URL is somehow not static 🤦‍♂️(mIRC's problem)

OfficialEsco:
Ohhhh i figured it out, `https://www.mirc.com/downloads/mirc/32/mirc764.exe` got deleted because they don't keep old versions, for me it redirect to `https://www.mirc.com/get.html` which somehow downloads a file which is unreadable.

The new "temporary" link is `https://www.mirc.com/downloads/mirc/33/mirc765.exe` but it gives a exit code 1223. You can work around this by adding a valid exit code.

I see Chocolatey uses `https://www.mirc.com/get.php?version=765` which seems like a dummy link because it always leads to the newest version, for example you can do `https://www.mirc.com/get.php?version=` and it will lead to 765...

chausner:
@OfficialEsco Thanks for your investigation! I am not sure if this was the original issue that @KevinLaMS experienced but in any case I now updated the URL and hash, and also switched to the "latest" versioning scheme.

OfficialEsco:
You also have to add
```
InstallerSuccessCodes:
  - 1223
```
for it to be valid btw

chausner:
> You also have to add
> 
> ```
> InstallerSuccessCodes:
>   - 1223
> ```
> 
> for it to be valid btw

I just tested once again and it installs successfully with exit code 0 for me in Windows Sandbox. Maybe under certain conditions, the exit code is different but unfortunately we cannot add both:
```yaml
InstallerSuccessCodes:
  - 0
  - 1223
 ```
as the manifest scheme explicitly disallows the value 0 when InstallerSuccessCodes is overridden.

OfficialEsco:
0 is the default success, you don't have to add both :)
Example: https://github.com/microsoft/winget-pkgs/blob/master/manifests/p/plex/Plex/1.31.0/plex.Plex.installer.yaml

We can't see the error on this validation so my guess would be its getting a 1223 exit code which is why it failed

chausner:
> 0 is the default success, you don't have to add both :)

Thanks for the hint, it appears it was really failing due to the success code. 👍

PR #10163, opened 2021-04-06
​

ItzLevvie:
​


KevinLaMS:
@ItzLevvie why are we adding so many teams versions?  it seems unnecessary.
thanks
kevin

ItzLevvie:
cc @KevinLaMS: because it helps with finding client-side regressions much easily. I can close other PRs and only add the latest ones  (2 per day) if you would like me to do that?

PR #10185, opened 2021-04-06
Proper AnyDesk MSI Manifest

OfficialEsco:

-----
#10184

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/10185)


jedieaston:
Why does there need to be both an MSI and exe version of the installer? Wouldn’t we prefer everyone to use the MSI?

If one arch is only shipped as exe, just put an InstallerType (and InstallerSwitches) as necessary for each installer in the manifest. WinGet will know what to do. That’s the strategy I used for Firefox, as they don’t ship MSI files for arm64 but do for x86/x64.

OfficialEsco:
@jedieaston well the thing is, AnyDesk's Exe and MSI installer are 2 "different" programs for some wild reason...
Their EXE acts like a portable program but it can be installed, while their MSI is a bit weird, and i've noticed some problems updating that one...
I don't think it will hurt by having a AnyDesk and AnyDeskMSI manifest, Chocolatey does the same.

denelon:
@jedieaston and @OfficialEsco is there any reason not to put both installers in the same manifest? The "InstallerType" can be inside the specific node for each one. If they are different versions, that's a challenge, and we don't yet have a clear preference setting for the installer type, but that's coming.

For now, I don't see a reason not to have two packages, but in the future it seems simpler for users to just have the one "package".

jedieaston:
@denelon, if I understand correctly, the problem is that the EXE installs under user scope while the MSI installs as system scope. If the packages were merged, winget may upgrade a user to the wrong version (at least, until winget can do scope). Similar to VSCode, in that regard.

OfficialEsco:
@jedieaston nono, as long as the exe contains -install "path" it will install as a regular program under the name "AnyDesk", i'm unsure how upgrade works because i've never gotten it to do it, but i believe the Package Name needs to match whatever Control panel says? In that case the msi is installed as "AnyDesk MSI".
image
AnyDesk MSI also installs under C:\Program Files(86)\AnyDeskMSI.
And i've just realized normal AnyDesk have a different Version number..

denelon:
@jedieaston and @OfficialEsco given that behavior, it might make sense to have two packages. Upgrade would have a hard time in this situation until we have more comprehensive support for choosing an installer type consistently. We'd also likely need more custom logic in "list" so the matching between the ARP entry and the PackageIdentifier in the Open Windows Catalog here would work correctly.

PR #10234, opened 2021-04-07
Update Release Manifest To Latest-Release

ghost:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/10234)


denelon:
@MaybelYT are you planning to switch the installers underneath the https://github.com/MabelYT/MabelTerm/releases/download/Release/Update.exe URL?

That will cause users to get the hash mismatch error as you release upgrades. You should also be aware the client uses "latest" as a sentinel value that would be installed over specific versions of the client.

It is better to provide explicit versions of the package with explicit URLs to each version. 

We are building tooling that should help with CI/CD scenarios or manual updates to generate new versions of manifests and make the PRs.

PR #10291, opened 2021-04-08
Fixed Join-Path in YamlCreate.ps1.

jedieaston:
The Join-Path command used to create the full path to the manifest folder in YamlCreate.ps1 was not compatible with versions of PowerShell below 6.0. Since Windows (at least for now, hopefully not forever) only ships with PowerShell 5.0, I fixed it so it will be compatible with both versions. 

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/10291)

### DIFF
--- a/Tools/YamlCreate.ps1
+++ b/Tools/YamlCreate.ps1
@@ -80,7 +80,8 @@ if (Test-Path -Path "$PSScriptRoot\..\manifests") {
     $ManifestsFolder = (Resolve-Path ".\").Path
 }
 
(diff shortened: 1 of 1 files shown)


jedieaston:
@OfficialEsco can you review this as well? Just want to make sure I didn't unintentionally break anything (although it works on my machine (tm)).

OfficialEsco (approved):
Ahhh, silly me just tested with Powershell 7... Yeah still works as expected.

KevinLaMS (approved):

PR #10719, opened 2021-04-14
Add WireGuard 0.3.11

luislavena:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/10719)

### DIFF
--- a/manifests/w/WireGuard/WireGuard/0.3.11/WireGuard.WireGuard.installer.yaml
+++ b/manifests/w/WireGuard/WireGuard/0.3.11/WireGuard.WireGuard.installer.yaml
@@ -1,0 +1,21 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+PackageIdentifier: WireGuard.WireGuard
+PackageVersion: 0.3.11
+Installers:
+- Architecture: x64
+  InstallerUrl: https://download.wireguard.com/windows-client/wireguard-amd64-0.3.11.msi
+  InstallerSha256: 794eb63bf0171f3c16a44adda1a05f3edafd58d47a3617e30c893ec527a2307a
+  InstallerType: msi
+  ProductCode: "{1103D5AC-8385-48F5-A2AE-045146E2EDC9}"
+- Architecture: arm64
+  InstallerUrl: https://download.wireguard.com/windows-client/wireguard-arm64-0.3.11.msi
+  InstallerSha256: 477bd6e83596eeb3cbe1d569a83691580b89bb2ef009aab5af0a5d4a2b7ca0d8
+  InstallerType: msi
+  ProductCode: "{211CE439-E7B8-4053-AC59-0E3D727EF032}"
+- Architecture: x86
+  InstallerUrl: https://download.wireguard.com/windows-client/wireguard-x86-0.3.11.msi
+  InstallerSha256: 3651c9d628f04871890046bdcd930f2497eb1ba336112be8692cafe1e9f95232
+  InstallerType: msi
+  ProductCode: "{86E41F9F-4A97-46B4-B8BC-A1240CC4AF42}"
+ManifestType: installer
+ManifestVersion: 1.0.0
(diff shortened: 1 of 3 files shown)


OfficialEsco:
Hey! Just a tip for future updates, ProductCode is different for each version, either update the product code or leave it blank, otherwise WinGet Upgrade will get confused

luislavena:
Hello @OfficialEsco, thank you for the tip. I thought `ProductCode` was used consistently across each architecture package (amd64 0.3.5 will be the same for amd64 0.3.11).

Did you received any particular conflict when installing? Apologies if this caused any issue on your end.

I couldn't find any particular reference to the usage of `ProductCode` in the manifest spec, so left the previous reference. I will keep in that in mind and remove the reference in future updates.

Cheers.

OfficialEsco:
I am by no means a expert and i do not really understand `ProductCode`'s but i know the main purpose of the `ProductCode` is to detect the Current version and Outdated version, instead of comparing to `PackageName`, `Publisher` and `Version` which i do not believe works yet (or a lot of packages are wrong).

I'm not using WireGuard so i don't know if it works or not (Its in experimental anyways), i just came here from OpenVPN and noticed `jedieaston` didn't update this one so i took a look 🙃 

The only way i know of finding the `ProductCode` is to install the program and run 
`get-wmiobject Win32_Product | Sort-Object Name | Format-Table IdentifyingNumber, Name -AutoSize` in `PowerShell 5` 
or 
`Get-CimInstance win32_product | Sort-Object Name | Format-Table IdentifyingNumber, Name -AutoSize` in `Powershell 7+`

According to `jedieaston` the `ProductCode` is different in each version for each `Architecture`

luislavena:
Thanks for the details @OfficialEsco!!! I just run that command and found that for this package the `ProductCode` definitely is different (got `{CC68D89C-B50F-4A7E-81C4-AD09F0B5D9A8} WireGuard`)

The only problem with the _install and check_ approach is that I cannot install an arm package or the x86 one, so I cannot obtain these codes 😿.

I hope there is another mechanism, but until that is published or connected, I will remove that from future updates to this package.

Thank you again for the details and your time! ❤️ ❤️ ❤️

OfficialEsco:
Oke i just tested a bit since i also have this problem with ARM64, you can apparently read the msi information without installing it but its a messy progress...
I found this command which works ONLY in Powershell 5...

`Get-AppLockerFileInformation -Path ".\wireguard-amd64-0.3.11.msi" | select -ExpandProperty Publisher | Select BinaryName`

In Powershell 7 it returns blank because when you run `Get-AppLockerFileInformation -Path ".\wireguard-amd64-0.3.11.msi" | select -ExpandProperty Publisher` it returns 
`Publisher  : O=WIREGUARD LLC, L=BOULDER, S=COLORADO, C=US\WIREGUARD\{CC68D89C-B50F-4A7E-81C4-AD09F0B5D9A8},0.3.11.0`

instead of 

```
PublisherName    : O=WIREGUARD LLC, L=BOULDER, S=COLORADO, C=US
ProductName      : WIREGUARD
BinaryName       : {CC68D89C-B50F-4A7E-81C4-AD09F0B5D9A8}
BinaryVersion    : 0.3.11.0
HasPublisherName : True
HasProductName   : True
HasBinaryName    : True
```

You can also use the carbon module
http://get-carbon.org/Get-Msi.html

denelon:
Great feedback on this one!

I've created a specific documentation issue to address ProductCodes:

https://github.com/microsoft/winget-cli/issues/921

OfficialEsco:
Daaamn you found us in a 19 day old PR, thank you for keeping a watch ❤️ 
I feel like this should be a function in Powershell since its actually is a Windows Installer which is getting more and more used..
Also the validation process should include a Product Code check.

denelon:
I actually started an internal e-mail thread with a couple of teams based on the comments here. We're already looking to see what we can do to improve the new winget-create tool we're preparing to ship. We're also looking into what the validation pipelines will be able to do. The current thinking is "warn" if there is no product code, and "error" if we can determine it's the wrong one.

PR #11651, opened 2021-04-27
Add Safing.Portmaster

OfficialEsco:
Currently Portmaster don't have a direct download link or a version number displayed in Windows (Probably because it can't?)
Portmaster is currently in Alpha, but its a really interesting project which is a alternative to Simplewall so it deserves some love :)

@dhaavi can you clear up the version thing and look over the information?

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/11651)


dhaavi:
Hey @OfficialEsco, thanks for making the effort!

Because the Portmaster auto-updates, the installers we distribute are only minimal and usually only describe themselves, similar to all these "light or web installers" that only download the real deal when they are run.

Based on that, I have the following feedback:
- Depending on the usage and behavior of winget, possibly use an identifier specific to the installer: `Safing.Portmaster.Installer`
  - While this would be more technically correct, just using `Safing.Portmaster` probably makes more sense for all practical use cases. 
- Use the version of the installer: v0.5.2
- Use the direct link to the installer, not the "latest" redirect: https://updates.safing.io/windows_amd64/packages/portmaster-installer_v0-5-2.exe
  - Especially as you are providing a hash for it.
- Mention to the user that the Portmaster auto-updates when doing a silent install, if possible.

Also, we will be issuing an update to the installer within the next month that will change the installation experience: The Portmaster service will not be started automatically after the install finished, but the user will be prompted to restart the device to complete installation. It would be interesting how this could be communicated in the install process here. If the installer is run interactively, then this should not be a problem, but I wonder how the silent installs would then work.

I'm not all too familiar with winget, so take my advice with a pinch of salt.

OfficialEsco:
Thank you for the additional information @dhaavi

Point 1. In the WinGet ecosystem i don't think it really matters if its `Safing.Portmaster` or `Safing.Portmaster.Installer`, eitherway we can always change it. I know Chocolatey does this but in WinGet we can specify a bit more and it reads the Programs and Features data to look for current and new versions.

Point 2/3. Ahaaa so this is actually the installer version then https://github.com/safing/portmaster-packaging/releases, got a bit confused when i saw Version 0.6.12 in the Portmaster settings.. I couldn't find a direct download link anywhere, but now that we have it it should be easy enough to update :)
Its also not a huge problem if the installer don't have a static download link since we have Wingetbot which updates hashes which does not match.

Point 4. What do you mean? The installer just updates the software instead of installing over it? Or that Auto-Update is enabled when installing with silent switch? Keep in mind if Portmaster auto-updates WinGet won't get a chance to Upgrade/Update the package

Point 5. Uhm, from experience you will never get prompted to do a restart with silent/silent with progress switches, but programs which asks for restart tends to be working fine. Ill keep an eye on it :)

OfficialEsco:
@KevinLaMS you're too quick

dhaavi:
> Thank you for the additional information @dhaavi

No problem!

> Its also not a huge problem if the installer don't have a static download link since we have Wingetbot which updates hashes which does not match.

Ah. Interesting. I think I still prefer manually setting a version, as this will result in less unexpected behavior of the whole ecosystem. We don't update installers often anyway.

> Point 4. What do you mean? The installer just updates the software instead of installing over it? Or that Auto-Update is enabled when installing with silent switch? Keep in mind if Portmaster auto-updates WinGet won't get a chance to Upgrade/Update the package

This is mainly about notifying the user that the Portmaster auto-updates. Some very privacy conscious might be taken by surprise if they learn about that after the fact. We just want to be as transparent as possible about how the Portmaster works.

Installing "over" an existing installation would have no meaningful effect and should not pose a problem. If we release a new installer via winget, would that new version be somehow automatically installed?

> Point 5. Uhm, from experience you will never get prompted to do a restart with silent/silent with progress switches, but programs which asks for restart tends to be working fine. Ill keep an eye on it :)

Great, thanks. I'll also keep silent installs in mind for future installer versions.

OfficialEsco:
> Ah. Interesting. I think I still prefer manually setting a version, as this will result in less unexpected behavior of the whole ecosystem. We don't update installers often anyway.

Yes version Specific links are still preferred, its just a workaround for companies that are not that cooperative 😄 why don't you publish them on the GitHub Packages repo? 

> This is mainly about notifying the user that the Portmaster auto-updates. Some very privacy conscious might be taken by surprise if they learn about that after the fact. We just want to be as transparent as possible about how the Portmaster works.

Yeah no as of now there is no way to do this and probably never will be, i think however anyone that is privacy minded will read every piece of documentation on your website

> Installing "over" an existing installation would have no meaningful effect and should not pose a problem. If we release a new installer via winget, would that new version be somehow automatically installed?

No it will not, the user will choose update all or update app by app. WinGet Upgrade is in the experimental stages, but as of now the way it works is by checking for Product Code, i do believe later on it will check for Version, Name and Publisher.

This is an improvement over Chocolatey and Scoop which internally keeps track of last installed version. Which then gets messed up if the software auto-updates..

dhaavi:
> Why don't you publish them on the GitHub Packages repo?

Do you mean NuGet?

> No it will not, the user will choose update all or update app by app. WinGet Upgrade is in the experimental stages, but as of now the way it works is by checking for Product Code, i do believe later on it will check for Version, Name and Publisher.

I see. Thanks for explaining.

OfficialEsco:
> > Why don't you publish them on the GitHub Packages repo?
> 
> Do you mean in Github's NuGet repo?

Not sure how GitHub releases work, but i mean like this https://github.com/henrypp/simplewall/releases where you can simply download the exe from the assets
(Also what i meant by GitHub Packages repo was here https://github.com/safing/portmaster-packaging/releases 🙃 )

PR #11753, opened 2021-04-29
YamlCreate small tweaks

OfficialEsco:
+ WinGet validate will run after the manifest have been created, updated and new locale is created
+ Added Natural sorting so it actually picks the last version
+ Tags can now be Cross-Platform instead of having to be Cross Platform

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/11753)

### DIFF
--- a/Tools/YamlCreate.ps1
+++ b/Tools/YamlCreate.ps1
@@ -200,9 +200,9 @@ Function Read-WinGet-InstallerValues {
         $architecture = Read-Host -Prompt 'Architecture' | TrimString
     }
 
-    while ($InstallerType -notin @('exe', 'msi', 'msix', 'inno', 'nullsoft', 'appx', 'wix', 'zip')) {
+    while ($InstallerType -notin @('exe', 'msi', 'msix', 'inno', 'nullsoft', 'appx', 'wix', 'zip', 'burn', 'pwa')) {
         Write-Host
-        Write-Host -ForegroundColor 'Green' -Object '[Required] Enter the InstallerType. For example: exe, msi, msix, inno, nullsoft'
+        Write-Host -ForegroundColor 'Green' -Object '[Required] Enter the InstallerType. For example: exe, msi, msix, inno, nullsoft, appx, wix, burn, pwa, zip'
         $InstallerType = Read-Host -Prompt 'InstallerType' | TrimString
     }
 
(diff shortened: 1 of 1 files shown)


jedieaston:
This is unrelated to these changes specifically (I can open an issue if you want to deal with this in your next PR), but when I run the script under PowerShell 5.1, `Out-File -Path $LocaleManifestPath -Encoding 'UTF8'` fails with an error that -Path isn't a valid argument. 

Is there a good way to refactor the script to keep compatibility or shall we add a check to ensure the user is running it using PowerShell Core?

OfficialEsco:
> PowerShell 5.1, `Out-File -Path $LocaleManifestPath -Encoding 'UTF8'` fails with an error that -Path isn't a valid argument.

Oke... The inconsistencies between Powershell 5 and Powershell Core/7 is starting to piss me off..
I believe this can be solved by not having -Path?

Thank you for testing :))

OfficialEsco:
Why does
`New-Variable -Name "FileExtensions" -Value ($FileExtensions.Split("- ").Trim()[1..257] -join ", ") -Scope Script -Force` break it
and not
`New-Variable -Name "FileExtensions" -Value ($FileExtensions.Split("-").Trim()[1..257] -join ", ") -Scope Script -Force` ????

OfficialEsco:
Hey @jedieaston please quality check it now :)
APPARENTLY Powershell 5 `-split` and `.split` works differently.

Tools/YamlCreate.ps1
```diff
@@ -141,7 +148,8 @@ Function Read-PreviousWinGet-Manifest {
                     $regex = '(?ms)Tags:(.+?):'
                     $FetchTags = [regex]::Matches($OldManifestText,$regex) | foreach {$_.groups[1].value }
                     $Tags = $FetchTags.Substring(0, $FetchTags.LastIndexOf(' '))
-                    New-Variable -Name "Tags" -Value ($Tags.Split("-").Trim()[1..100] -join ", ") -Scope Script -Force
+                    $Tags = $Tags -Split '- '
+                    New-Variable -Name "Tags" -Value ($Tags.Trim()[1..100] -join ", ") -Scope Script -Force
```
palenshus, line 152: only 16 tags allowed

palenshus (dismissed):

Tools/YamlCreate.ps1
OfficialEsco: Alright alright, fair play

palenshus (approved):

denelon (approved):

SteveL-MSFT:
@OfficialEsco regarding `Out-File` differences, `-FilePath` is the original parameter in WinPS 5.1, but for PowerShell 7, we took a community PR to give it a `-Path` alias to the same parameter to make it consistent with other cmdlets that have a `-Path`.  We don't backport changes to WinPS5.1 except for security fixes.

As for `-split` vs `.split()` the latter is the .NET method which takes a string whereas `-split` the operator takes a regex.

PR #13613, opened 2021-05-21
Renamed 7zip.7zipAlpha 21.02 alpha to 21.02-alpha

OfficialEsco:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/13613)


JiiPee74:
```
Name          Id                  Version     Available       Source
--------------------------------------------------------------------
7-Zip - Alpha 7zip.7zipAlpha      21.02 alpha 21.02-alpha     winget
```

Why was version changed? Now it shows up all a time that there is new upgrade because version does not match.

OfficialEsco:
Because previous manifest versions was that way which made the real version not appear as the latest version, feel like it should indeed be changed back..

JiiPee74:
> Because previous manifest versions was that way which made the real version not appear as the latest version, feel like it should indeed be changed back..

Real version has different name and different id so I don't think that alpha should conflict with it? Issue is that version string in manifest is "21.02-alpha" when reported version string is "21.02-alpha"

PR #15174, opened 2021-05-31
PortugueseGovernment.AutenticacaoGov version 3.5.4272

Touratica:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/15174)


Touratica:
It worked for me...

Touratica:
Everything was working when I tried locally.

denelon:
@Touratica this error usually occurs either when there is some kind of UI required by the installer to complete, or the UI pops up due to dependencies.

Touratica:
I've checked on a Windows Sandbox, it seems to be because it prompts if I'd like to install a device software. It's impossible to get around these, right?
image

denelon:
Yes, that's the current state of the system. We either need to support dependencies so we could install the right ones, and the installer wouldn't prompt, or we would need to support the ability to "prefer" silent installers so we can start allowing "interactive only" installers in the community repository.

Touratica:
It's not possible to add this program to the repository yet, due to always prompting the user, even when using the silent installer.

PR #15270, opened 2021-05-31
Using the version as used by publisher Adobe

Master-Guy:
Moving from version 2021.001.20155 to 21.001.20155 as Adobe is also numbering their versions, so that winget installs the correct version. No actual update happened, so the file hashes are still the same. Just reflecting the correct version in this git.

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/15270)


Master-Guy:
The reported error seems incorrect?

``` One or more errors occurred. (The pull request contains more than one manifest. Either singleton or multi file manifest under one version directory.) The pull request contains more than one manifest. Either singleton or multi file manifest under one version directory.```

After the push, all "old" versions are removed, and only a single new version is uploaded. This new version consists of the already existing files where only the version number has been changed, and the directory name has been changed to match the version number.

Is this part of the "version 1.0" manifest change, where it all collapsed into a single file?

OfficialEsco:
Ahhhh yeeeah i noticed this and was going to fix it, but i think this is a job for @ItzLevvie

So are you trying to Update ONE version or Delete multiple and Update ONE version?
You can only change one Manifest at a time, GitHub shows that you deleted 8 different versions which is why it fails.

Master-Guy:
> So are you trying to Update ONE version or Delete multiple and Update ONE version?
> You can only change one Manifest at a time, GitHub shows that you deleted 8 different versions which is why it fails.


In order to prevent version issues (2021 is higher then 21, so it probably won't upgrade to the 21.x.x versions), I deleted all the 2021 versions indeed.

OfficialEsco:
To preserve the version's you would need to update everyone of the manifests, which is why i haven't gotten around to do it, we also need to pick between Adobe DC locales or Adobe DC MUI (which have a different PackageName)

Master-Guy:
Cleaning up my fork. @OfficialEsco I hope you are willing to pick this up the correct way? 👍 Gonna see if I can update some other (easier) packages.

PR #15665, opened 2021-06-01
FTB.FTB version 1.21.55.1814

coehlrich:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/15665)

### DIFF
--- a/manifests/f/FTB/FTB/1.21.55.1814/FTB.FTB.installer.yaml
+++ b/manifests/f/FTB/FTB/1.21.55.1814/FTB.FTB.installer.yaml
@@ -1,0 +1,14 @@
+# Created using wingetcreate
+
+PackageIdentifier: FTB.FTB
+PackageVersion: 1.21.55.1814
+Installers:
+- Architecture: x64
+  InstallerType: exe
+  InstallerUrl: https://apps.modpacks.ch/FTBApp/release/202105051818-5b1f30277f-release/FTBA_windows-x64_202105051818-5b1f30277f-release.exe
+  InstallerSha256: 63F8AB17AA03EAF6362631EA8F4CC15F5179D3D3D0019ED4D44F124A9FBFD77D
+  InstallerSwitches:
+    Silent: -q
+    SilentWithProgress: -q
+ManifestType: installer
+ManifestVersion: 1.0.0
--- a/manifests/f/FTB/FTB/1.21.55.1814/FTB.FTB.locale.en-US.yaml
+++ b/manifests/f/FTB/FTB/1.21.55.1814/FTB.FTB.locale.en-US.yaml
@@ -1,0 +1,18 @@
+# Created using wingetcreate
+
+PackageIdentifier: FTB.FTB
+PackageVersion: 1.21.55.1814
+PackageLocale: en-US
+Publisher: Feed The Beast
+PackageName: FTB App
+PackageUrl: https://www.feed-the-beast.com/
+License: LGPL 2.1 (frontend), CC BY-NC-ND (backend)
+LicenseUrl: https://github.com/FTBTeam/FTB-App/blob/master/LICENSE.md
+ShortDescription: FTB.FTB
+Moniker: ftb
+Tags:
+- game
+- minecraft
+- ftb
+ManifestType: defaultLocale
+ManifestVersion: 1.0.0
--- a/manifests/f/FTB/FTB/1.21.55.1814/FTB.FTB.yaml
+++ b/manifests/f/FTB/FTB/1.21.55.1814/FTB.FTB.yaml
@@ -1,0 +1,7 @@
+# Created using wingetcreate
+
+PackageIdentifier: FTB.FTB
+PackageVersion: 1.21.55.1814
+DefaultLocale: en-US
+ManifestType: version
+ManifestVersion: 1.0.0


jedieaston:
This app doesn't appear to install silently for me. It asks questions as to whether or not I want to install it (even though winget exits, it must just use the downloaded installer to bootstrap). Can you see if there are different silent flags for this release?

manifests/f/FTB/FTB/1.21.55.1814/FTB.FTB.installer.yaml
```diff
@@ -0,0 +1,14 @@
...
+PackageIdentifier: FTB.FTB
+PackageVersion: 1.21.55.1814
+Installers:
+- Architecture: x64
+  InstallerType: exe
+  InstallerUrl: https://apps.modpacks.ch/FTBApp/release/202105051818-5b1f30277f-release/FTBA_windows-x64_202105051818-5b1f30277f-release.exe
+  InstallerSha256: 63F8AB17AA03EAF6362631EA8F4CC15F5179D3D3D0019ED4D44F124A9FBFD77D
+  InstallerSwitches:
+    Silent: -q
+    SilentWithProgress: -q
```
jedieaston, line 12: This app doesn't appear to install silently for me. It asks questions as to whether or not I want to install it (even though winget exits, it must just use the downloaded installer to bootstrap). Can you see if there are different silent flags for this release?

jedieaston (changes requested):

manifests/f/FTB/FTB/1.21.55.1814/FTB.FTB.installer.yaml
coehlrich: After a quick look through the flags that install4j has there doesn't seem to be any new flags for unattended mode but when I tried it the only thing requiring user interaction was the popup that windows has that asks if you still want to run it when it hasn't been run enough times.
There is another installer with a different url that the main website uses that I haven't tried but doesn't seem to have anything relating to the version in the url and the installer that it is currently set to use installs the latest version even if the installer is for an older version.

jedieaston: This is the window that appears when I run the manifest. 
image

It's clear that nothing's installed until I click through this, and if I click out of it, there's no way to get to it outside of running the installer again. It may be worth opening a bug in FTBTeam/FTB-App if you can't get this to work, because I know it did at one point.

coehlrich: -q does work but Overwolf needs to have already been installed.

coehlrich: I have tried the official download link but Winget seems to have a problem downloading it correctly and it installs Overwolf if it isn't already installed.

OfficialEsco: Nvm it appears https://apps.modpacks.ch/FTBApp/release/ is their domain 🤦‍♂️
I've done some more testing with your URL, i am pretty sure it requires `Java Runtime` since that is the first thing it downloads AND it definitively requires Overwolf to be installed, it did actually install with -silent, but `Welcome to FTB app on Overwolf` is still on the screen even tho it installed.

I don't think the Overwolf app likes the unattended install since it wants to do it itself.

coehlrich: The installer I was referring to when I said the installer that the installer uses that this pr uses was https://apps.modpacks.ch/Butthunter/FTB%20AppInstaller.exe (which is referred to from https://github.com/CreeperHost/modpacklauncher/blob/master/installerstuff.install4j#L123 (which https://github.com/CreeperHost/modpacklauncher/ is referred to from https://github.com/FTBTeam/FTB-App#ftb-app) where if it is running on windows it downloads the exe then runs it then exits) which does download correctly and works properly with the -silent parameter but seems to be signed on an earlier date than the one hosted on https://download.overwolf.com/ while also being for an older version of Overwolf but FTB App still installs fine.

coehlrich: The installers on apps.modpacks.ch/FTBApp/releases weren't meant to have been generated and have been removed and the main installer are the ones from overwolf which wingetcreate and winget dont the ones on download.overwolf.com download correctly

coehlrich: It seems that when downloaded through most downloader's the output is gzip compressed

ItzLevvie (changes requested):
https://github.com/microsoft/winget-pkgs/pull/15665#pullrequestreview-673690159

coehlrich:
I have tell overwolf support about a month ago that the download subdomain always uses gzip even if the client doesn't support gzip and they plan to support winget sometime in the future.

ItzLevvie (changes requested):
https://github.com/microsoft/winget-pkgs/pull/15665#issuecomment-873475001

OfficialEsco:
@coehlrich don't reply until its fixed 😁 Levvie is hiding the PR from the PR list so we don't have to keep checking it

PR #15881, opened 2021-06-02
Delete Plex v1.14.0

JackW6809:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/15881)


KaranKad:
Why delete it? does it not work anymore?

JackW6809:
> Why delete it? does it not work anymore?

I do not see the reason in keeping a version of an application that would be outdated. Also Plex auto updates so installing that version would be useless as it would just auto update.

OfficialEsco (approved):
Yeeee you're right, unless we find a disable auto update switch this is useless

PR #16094, opened 2021-06-03
Fixed Winamp version

KaranKad:
image
Publisher is called Winamp SA might have to change that as well.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/16094)

### DIFF
--- a/manifests/r/Radionomy/Winamp/5.8/Radionomy.Winamp.yaml
+++ b/manifests/r/Radionomy/Winamp/5.8/Radionomy.Winamp.yaml
@@ -1,8 +1,9 @@
 PackageIdentifier: Radionomy.Winamp
-PackageVersion: 5.8.3660-Beta
+PackageVersion: 5.8
 PackageName: Winamp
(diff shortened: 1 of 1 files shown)


manifests/r/Radionomy/Winamp/5.8/Radionomy.Winamp.yaml
```diff
@@ -1,8 +1,9 @@
 PackageIdentifier: Radionomy.Winamp
-PackageVersion: 5.8.3660-Beta
+PackageVersion: 5.8
 PackageName: Winamp
 Publisher: Radionomy
```
jedieaston, line 4:
```suggestion
Publisher: Winamp SA
```

jedieaston (changes requested):
Can you change the publisher too, so it matches? Thanks!

OfficialEsco (approved):

Issue #9476, opened 2021-03-31
How to move from existing singleton to multi-file manifests?


dpprdan:
# Description 

Some existing manifests (e.g. for [pgAdmin](https://github.com/microsoft/winget-pkgs/tree/master/manifests/p/PostgreSQL)) are duplicated for different installers (x86 and x64 in this case). How should we proceed with these moving forward?

1. Should the existing manifests be consolidated into multi-file manifests in the pgAdmin directory? Or leave them as-is? (To further complicate things there aren't manifests for both installers for all recent pgAdmin versions).
2. Where should manifests for future versions live, i.e. where would I create the manifests for pgAdmin v5.1? I assume they should preferably be multi-file manifests?

3. Only partly related: What's the general policy for new versions? Is it okay to update existing manifests to the new versions (essentially erasing the manifests for the previous version - this [has happened for pgAdmin](https://github.com/microsoft/winget-pkgs/pull/8081/files)) or should a new manifest be created always (always assuming that the previous version is still available at the source).

denelon:
@dpprdan I've been looking at Visual Studio Code as another example.

I think we want a single (multi-file) manifest for packages like this. It may mean we need to create a new package for the program with the multiple installers. We're still working on [User vs. System](https://github.com/microsoft/winget-cli/issues/149), but the Architecture is auto-selected by the client to be the most optimal one for the machine.

For pgAdmin I would suggest creating the multi-file manifest in the /p/PostgreSQL/pgAdmin path to add support for both installers since that package name doesn't have the x86 architecture specified.

For a given version, I would delete the singleton manifest and add the multi-file manifest.

denelon:
For new versions of packages, I'd create a new manifest (assuming the URL is unique to a specific version). If the URL for the installer is the same for multiple versions, I'd delete the old manifest.

dpprdan:
Thanks for changing all the manifests to multi-file, @OfficialEsco! 🚀 

Tagging the PRs for deleting the existing x64 manifests for completeness: #9536, #9537, #9538, #9539, #9540.

> It may mean we need to create a new package for the program with the multiple installers.

Sorry, I don't follow, @denelon. Could you elaborate, please?

Re "What's the general policy for new versions?" I assume it's "create a new manifest" (as long as the previous version is still available at the source). Maybe this could be documented somewhere (if it isn't already and I missed it)?

denelon:
@dpprdan it looks like @OfficialEsco did what I had suggested.

[...]

PR #16352, opened 2021-06-05
add RProject.Rtools 4.0.1

dpprdan:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/16352)

### DIFF
--- a/manifests/r/RProject/Rtools/4.0.1/RProject.Rtools.installer.yaml
+++ b/manifests/r/RProject/Rtools/4.0.1/RProject.Rtools.installer.yaml
@@ -1,0 +1,23 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+PackageIdentifier: RProject.Rtools
+PackageVersion: 4.0.1
+MinimumOSVersion: 10.0.0.0
+InstallModes:
+  - interactive
+  - silent
+  - silentWithProgress
+Installers:
+  - Architecture: x64
+    InstallerType: inno
+    InstallerUrl: https://cloud.r-project.org/bin/windows/Rtools/rtools40v2-x86_64.exe
+    InstallerSha256: 3A41212B16523473E6343680D665C5DCC23D068D63C28F0D4CE91C5CE50960DD
+    InstallerSwitches:
+      Custom: /NORESTART
+      Silent: /VERYSILENT
+      SilentWithProgress: /SILENT
+#    ProductCode: 
+    Scope: machine
+    InstallerLocale: en-US
+    UpgradeBehavior: uninstallPrevious
+ManifestType: installer
+ManifestVersion: 1.0.0
--- a/manifests/r/RProject/Rtools/4.0.1/RProject.Rtools.locale.en-US.yaml
+++ b/manifests/r/RProject/Rtools/4.0.1/RProject.Rtools.locale.en-US.yaml
@@ -1,0 +1,22 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultlocale.1.0.0.schema.json
+PackageIdentifier: RProject.Rtools
+PackageVersion: 4.0.1
+PackageLocale: en-US
+Publisher: RProject
+PublisherUrl: https://www.r-project.org/
+PublisherSupportUrl: https://www.r-project.org/help.html
+#PrivacyUrl: 
+#Author: 
+PackageName: Rtools
+PackageUrl: https://cran.r-project.org/bin/windows/Rtools/
+License: BSD-3-Clause
+LicenseUrl: https://github.com/r-windows/rtools-base/blob/master/LICENSE
+#Copyright: 
+#CopyrightUrl: 
+ShortDescription: Rtools is the toolchain bundle that is used on Windows to build R base and R packages that contain compiled code. You can download it from CRAN.
+#Description: 
+Tags:
+  - R
+  - Rstats
+ManifestType: defaultLocale
+ManifestVersion: 1.0.0
--- a/manifests/r/RProject/Rtools/4.0.1/RProject.Rtools.yaml
+++ b/manifests/r/RProject/Rtools/4.0.1/RProject.Rtools.yaml
@@ -1,0 +1,6 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.version.1.0.0.schema.json
+PackageIdentifier: RProject.Rtools
+PackageVersion: 4.0.1
+DefaultLocale: en-US
+ManifestType: version
+ManifestVersion: 1.0.0


manifests/r/RProject/Rtools/4.0.1/RProject.Rtools.installer.yaml
```diff
@@ -0,0 +1,23 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+PackageIdentifier: RProject.Rtools
+PackageVersion: 4.0.1
```
ItzLevvie, line 3:
```suggestion
PackageVersion: 4.0
```
image

manifests/r/RProject/Rtools/4.0.1/RProject.Rtools.locale.en-US.yaml
```diff
@@ -0,0 +1,22 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultlocale.1.0.0.schema.json
+PackageIdentifier: RProject.Rtools
+PackageVersion: 4.0.1
```
ItzLevvie, line 3:
```suggestion
PackageVersion: 4.0
```
image

manifests/r/RProject/Rtools/4.0.1/RProject.Rtools.locale.en-US.yaml
```diff
@@ -0,0 +1,22 @@
...
+PackageLocale: en-US
+Publisher: RProject
+PublisherUrl: https://www.r-project.org/
+PublisherSupportUrl: https://www.r-project.org/help.html
+#PrivacyUrl: 
+#Author: 
+PackageName: Rtools
```
ItzLevvie, line 10:
```suggestion
PackageName: Rtools
```

manifests/r/RProject/Rtools/4.0.1/RProject.Rtools.yaml
```diff
@@ -0,0 +1,6 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.version.1.0.0.schema.json
+PackageIdentifier: RProject.Rtools
+PackageVersion: 4.0.1
```
ItzLevvie, line 3:
```suggestion
PackageVersion: 4.0
```
image

ItzLevvie (changes requested):
​

dpprdan:
@ItzLevvie I am not sure about the `PackageVersion: 4.0` as opposed to `PackageVersion: 4.0.1`, since this is a [new version of Rtools4](https://twitter.com/opencpu/status/1395692067884503044) and 4.0.0 [already exists](https://github.com/microsoft/winget-pkgs/blob/67ecd21017fc62b8e6ac6f836bb63a6b7ab96e61/manifests/r/RProject/Rtools/4.0.0/RProject.Rtools.yaml#L4) in winget-pkgs. So `4.0` seems confusing IMHO. What do you think?

dpprdan:
@jeroen Would it be possible to use semantic versioning for Rtools everywhere? On https://cran.r-project.org/bin/windows/Rtools/ the latest version's file name is `rtools40v2-x86_64.exe` and in Windows > Apps & Features it shows up as `Rtools 4.0 (4.0.1.0)`. This is a slightly confusing, see above.

screenshot with Rtools 4.0 (4.0.1.0)

IMHO it would be better to name it `rtools-4.0.1-x86_64.exe` and `Rtools 4.0.1`, respectively (cf `R-4.1.0-win.exe` for [base R](https://cran.r-project.org/bin/windows/base/)).

ItzLevvie:
The issue with keeping it `4.0.1` is that when someone does `winget upgrade` on their machine it'll always keep updating it even if you're on the latest version because 4.0.1 is newer than what's shown in Control Panel (and you'll end up in an endless loop of updating.)

jeroen:
I don't think calling it 4.0.1 is a good idea because that may be confused that this is related to the R version 4.0.1. This is not the case. Rtools 4.0 can be used with any R >= 4.0.

The `.1` is the built number postfix. Perhaps I should use a date there, just like msys2. So that it becomes `4.0.20210604`

ItzLevvie (changes requested):
​

jeroen:
@ItzLevvie thanks! An Rtools 4.0 package is indeed the proper solution

dpprdan:
Updating the manifest seems to be the right solution in this particular case since the version tag is not properly set in the installer. 

In general, though, I still think it would be better to have a manifest for every (proper) version. At least this is how I read @denelon's [advice](https://github.com/microsoft/winget-pkgs/issues/9476#issuecomment-811158348) to "create a new manifest (assuming the URL is unique to a specific version)" for new versions of packages and "to retain manifests for every valid version that may still be installed". 

Older versions of RTools are still available back to RTools26 [at the source](https://cran.r-project.org/bin/windows/Rtools/history.html), Chocolatey for example also [lists multiple versions](https://community.chocolatey.org/packages/rtools/3.5.0.4#releasenotes). Why should this be any different with winget? The default `winget install rtools` command (i.e. without the `--version` tag) or `winget upgrade` ensure that users get the recommended latest version. But if they need to get an old version for e.g. debugging they still can.

@ItzLevvie While I agree with the solution, I have to say that I am puzzled by your style. Maybe it's just me, but it feels quite insincere to have the moderator open another PR (#16740 by the way) without notice instead of waiting for the OP to react to and fixing the issue. Even more so when it's done with a second account with the stated goal "to merge my own PRs to make my changes available to production as soon as possible." Is this a race somehow?

grafik

This feels like volunteer contributions are not valued very much here. I am not sure whether this is the sentiment you want to convey.

ItzLevvie:
I'm sorry if that's what you think - anyone is free to contribute to the catalog and that's why this repository was open-source to begin with. Without anyone's help, it wouldn't be a "package manager" as it wouldn't have that many applications.

The second account wasn't made to "race" with anyone. It was created merely to create "own" PRs and get it merged faster into master without waiting for another moderator's approval as I put out hundreds of PRs to fix, update or remove manifests and it may take hours or days for it to be merged, not to mention that it also has to be tested to see if it shows up currently in `winget list`, etc.

I'm in the process of installing 1900+ applications on a virtual machine and fixing some of the versions, names, and publishers, so even if the account and this PR wasn't created, the problem with this manifest would've been noticed by me and fixed.

It was only several weeks ago that I realized how important it is to update values to match what's shown in Control Panel's ARP. Unfortunately, I don't use Chocolatey, so I don't know how their system works in terms of what they do to see if a particular application is installed, so it's not really my fault if application developers aren't showing a proper version or updating it, etc.

dpprdan:
> I'm sorry if you feel that way, and not notifying you about the PR I had created was a mistake on my end and I deeply apologize for that.

Thank you, that's all I needed to hear! 😃 

As for the second account: I understand the reasons for creating it. This workaround just looks weird from the outside and it sounds to me like processes are not optimal yet. But I am sure you and the winget team will find a good solution here.

Thanks for your work!

denelon:
@dpprdan we are working to improve all of the end to end scenarios. Moderation is a very recent necessity to help ensure a high quality catalog of packages in the community repository.  At present, there are only three community moderators fielding hundreds of submissions from the community, and trying to keep up with "Help-Wanted" asks. We have been balancing direct community support against fixing bugs, and building new features. These are exciting times.

We sincerely appreciate contributions!
* Feedback
* Feature Requests
* Package Submissions
* Manifest improvements
* Bug Reports
* Discussions
* PRs on documentation
* PRs on code

The last couple of weeks have been a bit chaotic with the number of new users and newly identified edge cases. Many of them were known, but many of them and several new ones have finally become reproducible. 

Feedback is critical to our success. We believe in transparency and we rely on support from the community of customers, publishers, and ISVs to help us build the best product possible.

PR #16735, opened 2021-06-08
Added ProductCode for REALiX.HWiNFO

KaranKad:
winget is not able to uninstall this application using moniker, will see if adding product code helps.
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/16735)


OfficialEsco (approved):

KaranKad:
It works!
image

OfficialEsco:
Oooo, i see whats wrong here, the ARP is failing to match, probably because the `PackageName` is `HWiNFO64`/`HWiNFO32` and the `Publisher` is `Martin Malik - REALiX`. I think how the matching works is `HWiNFO "wildcard"` and Publisher just fails is there is a `-` in it...
We should probably add a installer for x86 with `ProductCode: "HWiNFO32_is1"`

PR #16823, opened 2021-06-09
fixed LSoftTechnologies.ActiveDiskEditor

donid:
PackageVersion and PackagUrl

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/16823)

### DIFF
--- a/manifests/l/LSoftTechnologies/ActiveDiskEditor/7.0.19/LSoftTechnologies.ActiveDiskEditor.installer.yaml
+++ b/manifests/l/LSoftTechnologies/ActiveDiskEditor/7.0.19/LSoftTechnologies.ActiveDiskEditor.installer.yaml
@@ -1,0 +1,11 @@
+# Created using wingetcreate
+
+PackageIdentifier: LSoftTechnologies.ActiveDiskEditor
+PackageVersion: 7.0.19
+Installers:
+- Architecture: x86
+  InstallerType: inno
+  InstallerUrl: https://www.disk-editor.org/download/diskeditor-freeware.exe
+  InstallerSha256: 25A31F05CFFB1F96ED1F97BCF399F295DFC6A2DC3B6B793B24C807889F42BE0D
+ManifestType: installer
+ManifestVersion: 1.0.0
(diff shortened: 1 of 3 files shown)


ItzLevvie (changes requested):
image

Control Panel shows `7`; and winget upgrade (as well as winget list) will always be in a constant loop of updating the application.

donid:
> will always be in a constant loop of updating

Ouch, that's bad. So, should I create a new PR where the PackageVersion is still 7 and I put the '7.0.19' in the ShortDescription? (And of course add the PackagUrl)

ItzLevvie:
> Ouch, that's bad. So, should I create a new PR where the PackageVersion is still 7 and I put the '7.0.19' in the ShortDescription? (And of course add the PackagUrl)

I think it's fine as is, because even on the website, https://www.disk-editor.org/index.html,  I see version 7 rather than 7.0.19, and if it's added into the ShortDescription, those that are using `wingetcreate update` may not notice that they'll have to update that too.

donid:
>I see version 7 shown rather than 7.0.19

Well, the Download-Link on that page actually (and wrongly) states "7.0.15" (the light-blue rectangle labeled "FREEWARE for WINDOWS" on the top/right) - that's a bit of a mess. I am wondering how a future update would be handled by winget when the publisher releases another bugfix-update (7.0.xx) that is still called "7".

OfficialEsco (changes requested):
It doesn't matter, its not a static link so wingetbot will update the hash once it dosn't match.
Just change the PackageVersion and Folder name as suggested.

richardsondev:
This is a duplicate of the existing 7 manifest. The URL is not different and hash is the same. Why aren't you editing the existing one? The version in this manifest is wrong. You have to use the control panel verison. I do not believe this should be approved.

richardsondev:
In addition, you did not accept the policies of the repro and confirm that the installation works. If this manifest was tested and verified you would have found the version shows as 7 in winget even with this manifest.

OfficialEsco:
Ahh you're right, @donid just close the PR

PR #16867, opened 2021-06-09
Microsoft.Git: update to 2.32.0.vfs.0.0

ldennington:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/16867)


ItzLevvie (changes requested):
The Control Panel displays the version as `2.32.0.0.0` and as Publisher as `The Git Development Community`:
image

Windows Package Manager requires the entires to match what's shown in Control Panel (for this particular manifest - that will exclude PackageName as updating it will conflict with the `Git.Git` package) so that other people can upgrade the application without being on a constant upgrade loop.

ldennington:
> The Control Panel displays the version as `2.32.0.0.0` and as Publisher as `The Git Development Community`:
> image
> 
> Windows Package Manager requires the entires to match what's shown in Control Panel (for this particular manifest - that will exclude PackageName as updating it will conflict with the `Git.Git` package) so that other people can upgrade the application without being on a constant upgrade loop.

Sure, I'm fine with making the necessary changes. To be clear - are you just asking for an update of the `PackageVersion` field to align with what's shown in Control Panel, or are there additional fields that should be updated?

ItzLevvie:
I'm asking for an update to the `PackageVersion` and `Publisher` fields to match what's shown in Control Panel so others can easily upgrade via `winget upgrade`.

ItzLevvie (approved):

manifests/m/Microsoft/Git/2.32.0.0.0/Microsoft.Git.yaml
```diff
@@ -0,0 +1,22 @@
...
+  Git distribution to support monorepo scenarios.
+  Note: This is not Git for Windows. Unless you are working in a monorepo and require
+  specific Git modifications, please run `winget install git` to start using Git for Windows.
+Installers:
+- Architecture: x64
+  InstallerUrl: https://github.com/microsoft/git/releases/download/v2.32.0.vfs.0.0/Git-2.32.0.vfs.0.0-64-bit.exe
+  InstallerType: inno
+  InstallerSha256: 73fe3e86ae419b7f63f87635850dc5d3c8285379e28b765e67501e502d4588b1
+  InstallerSwitches:
+    Custom: /COMPONENTS="AUTOUPDATE"
```
derrickstolee, line 19: @ldennington We should remove these `InstallerSwitches` now that we've adjusted the installer defaults.

ldennington: Will do - I'm going to do a larger manifest update once https://github.com/mjcheetham/update-winget/pull/124 merges and a new release of the update-winget task is cut. I will be sure to include at that time, and will manually make the update for now.

ldennington: Oh never mind - thought this was the new PR against one of the new versions. Will not manually update :).

derrickstolee:
> The Control Panel displays the version as `2.32.0.0.0` and as Publisher as `The Git Development Community`:
> image
> 
> Windows Package Manager requires the entires to match what's shown in Control Panel (for this particular manifest - that will exclude PackageName as updating it will conflict with the `Git.Git` package) so that other people can upgrade the application without being on a constant upgrade loop.

@itzlevvie: We want to be very careful to differentiate this package from the "standard" one that is Git for Windows. This version includes custom changes created by Microsoft and GitHub. Can we change the publisher to "Microsoft" instead of "The Git Development Community"?

I also notice that the license is listed as "Copyright (C) Microsoft Corporation" but at least the Git code is LGPLv2. Perhaps some other pieces that we ship in the package are not (@dscho might know a few of these components).

I look forward to your advice here.

dscho:
> the Git code is LGPLv2

It's actually GPLv2. And yes, different components shipped together with Git have different licenses (`git.exe` does not link to them).

ItzLevvie:
cc @derrickstolee: I believe you may have to update the publisher in the application to be `Microsoft` rather than changing it in the manifest itself.

This is what the documentation shows `Note: With the 1.0 release of the Windows Package Manager, this name affects how packages from a source are mapped to Apps installed in Windows 10 via Add / Remove Programs (ARP). The best practice is to ensure this matches the ARP entry for the package when it has been installed. The impact is associated with winget upgrade and winget list.` - from https://github.com/microsoft/winget-pkgs/blob/master/doc/manifest/schema/1.0.0/defaultLocale.md#publisher

Which is why I suggested to change it.

The license can be changed to `LGPLv2`. It was `Copyright (C) Microsoft Corporation` because there wasn't a `Copyright` and `CopyrightUrl` field before, so it was put in `License` and `LicenseUrl` instead.

derrickstolee:
> cc @derrickstolee: I believe you may have to update the publisher in the application to be `Microsoft` or `Microsoft Corporation` or something along those lines rather than changing it in the manifest itself.

Ok, so there is something we are doing wrong in our bundling of the product, probably because of how it shares a lot of components with `git-for-windows/git`. We will look into that part on our side.

PR #16994, opened 2021-06-10
Docs installers

denelon:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/16994)


doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,444 @@
...
+</details>
+
+### ManifestVersion
+<details>
+ <summary>The manifest syntax version</summary>
+
+ #### Required Field
+
+ This key must have the value "1.0.0". The Microsoft community package repository validation pipelines also use this value to determine appropriate validation rules when evaluating this file.
+</details>
```
ashpatil-msft, line 444: Do you think we should provide sample manifests (or links to them if they already exist) so the user can reference?

denelon: I'm leaning away from doing that as they tend to change and might get deleted.

doc/manifest/schema/1.0.0/defaultLocale.md
```diff
@@ -0,0 +1,208 @@
...
+ The current best practice is to use the value reported in Add / Remove Programs when this version of the package is installed. In some cases, packages do not report a version resulting in an upgrade loop or other unwanted behavior.
+</details>
+
+#### PackageLocale
+<details>
+  <summary>The package meta-data locale</summary>
+
+  #### Required Field
+  
+  This key represents the locale for package meta-data. The format is BCP-47. This value identifies the language for meta-data to be displayed to a user when no locale file matching their preferences is available. The Microsoft community package repository validation pipelines also use this value to determine appropriate validation rules for this file.
```
yao-msft, line 36: Also this is considered the key for one localization during winget client localization selection logic.

doc/manifest/schema/1.0.0/defaultLocale.md
```diff
@@ -0,0 +1,208 @@
...
+  >Note: This was included for future integration with the Microsoft Store source to provide the ability to display the full package description.
+ </details>
+ 
+ #### Moniker
+<details>
+  <summary>The most common package term</summary>
+
+  #### Optional Field
+
+  This key represents the most common term users would search for when installing or upgrading a package. If only one package uses this moniker, then the install, list and upgrade command will match with this package.
```
yao-msft, line 177: nit: IIRC the moniker is only 3rd in priority when we do package matching. This statement seems like over simplifying our matching logic.

doc/manifest/schema/1.0.0/defaultLocale.md
```diff
@@ -0,0 +1,208 @@
...
+ 
+ #### Tags
+<details>
+  <summary>List of additional package search terms</summary>
+
+  #### Optional Field
+
+  This key represents other common term users would search for when looking for packages. 
+
+  >Note: The best practice is to present these terms in all lower case with hyphens rather than spaces.
```
yao-msft, line 188: I don't think we would recommend hyphens over spaces since we haven't implemented fuzzy matching yet

doc/manifest/schema/1.0.0/defaultLocale.md
```diff
@@ -0,0 +1,208 @@
...
+ </details>
+ 
+
+### ManifestType
+<details>
+ <summary>The manifest type</summary>
+
+ #### Required Field
+
+ This key must have the value "version". The Microsoft community package repository validation pipelines also use this value to determine appropriate validation rules when evaluating this file.
```
yao-msft, line 198:
```suggestion
 This key must have the value "defaultLocale". The Microsoft community package repository validation pipelines also use this value to determine appropriate validation rules when evaluating this file.
```

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,444 @@
+# Windows Package Manager
+## Manifest Schema v1.0.0 "installer" File
```
yao-msft, line 2: nit: Manifest Schema v1.0.0 Installer File to match the pattern in above default locale doc?

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,444 @@
...
+
+</details>
+
+### Platform
+<details>
+ <summary>The installer supported operating system</summary>
+
+ #### Optional Field
+
+ This key represents the Windows platform targeted by the installer. The Windows Package Manager currently supports "Windows.Desktop" and "Windows.Universal".
```
yao-msft, line 51: I thought this field is more for the Store source to check applicability. Currently winget client does not use this field

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,444 @@
...
+<details>
+ <summary>List of supported installer modes</summary>
+
+ #### Optional Field
+
+ This key represents the install modes supported by the installer. The Microsoft community package repository requires a package support "silent" and "silent with progress". The Windows Package Manager also supports "interactive" installers.
+
+ >Note: Some installers will attempt to install missing dependencies. If these dependencies require user interaction, the package will not be allowed into the Microsoft community package repository.
+
+ >Note: This key may be present in the root of the manifest as the default value for all installer nodes. This key may also be present in an individual installer node as well. If this key is in the manifest root and in an installer node, the value in the installer node will apply.
```
yao-msft, line 107: Note: not used by winget client yet

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,444 @@
...
+</details>
+
+### InstallerSwitches
+<details>
+ <summary>Silent is the value that should be passed to the installer when user chooses a silent or quiet install</summary>
+
+ #### Optional Field
+
+ This key represents switches passed to the installer to provide a silent install experience.
```
yao-msft, line 116:
```suggestion
 This key represents switches passed to the installer to provide a specific install experience.
```

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,444 @@
...
+
+ >Note: When the Windows Package Manager installs a package using the "silent" install mode, any custom switches will also be passed to the installer. If a user applies override switches via command line via the Windows Package Manager, none of the switches from the manifest will be passed to the installer.
```
yao-msft, line 118: Custom switch is always passed to installer regardless of "silent" or "non-silent" install, unless user provides Override using cmdline args which overrides all switches.

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,444 @@
...
+
+</details>
+
+### SilentWithProgress
+<details>
+ <summary>SilentWithProgress is the value that should be passed to the installer when user chooses a non-interactive install</summary>
+
+ #### Optional Field
+
+ This key represents switches passed to the installer to provide a silent with progress install experience. This is intended to allow a progress indication to the user, and the indication may come from an installer UI dialogue, but it must not require user interaction to complete.
```
yao-msft, line 130: nit: we may mention this is the default install behavior

doc/manifest/schema/1.0.0/installer.md
yao-msft: Never mind, I misread the notes.

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,444 @@
...
+
+</details>
+
+### InstallLocation
+<details>
+ <summary>InstallLocation is the value passed to the installer for custom install location. </summary>
+
+ #### Optional Field
+
+ This key represents the path to install the package if the installer supports installing the package in a user configurable location. The &lt;INSTALLPATH&gt; token can be included in the switch value so the Windows Package Manager will replace the token with user provided path.
```
yao-msft, line 153: shall we hightlight the <INSTALLPATH> ?

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,444 @@
...
+
+</details>
+
+### Log
+<details>
+ <summary>Log is the value passed to the installer for custom log file path.</summary>
+
+ #### Optional Field
+
+  This key represents the path logs will be directed to if the installer supports specifying the log path in a user configurable location. The &lt;LOGPATH&gt; token can be included in the switch value so the Windows Package Manager will replace the token with user provided path.
```
yao-msft, line 163: I suggest highlight the "LOGPATH" part too

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,444 @@
...
+
+</details>
+
+### Upgrade
+<details>
+ <summary>Upgrade is the value that should be passed to the installer when user chooses an upgrade</summary>
+
+ #### Optional Field
+
+ This key represents the switches to be passed to the installer during an upgrade.
```
yao-msft, line 173: This switch will be passed toinstaller only if UpgradeBehavior is Install

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,444 @@
...
+
+</details>
+
+### SignatureSha256
+<details>
+ <summary>SignatureSha256 is recommended for appx or msix. It is the sha256 of signature file inside appx or msix. Could be used during streaming install if applicable</summary>
+
+ #### Optional Field
+
+ This key represents the signature file inside an MSIX installer. It is used to provide streaming install for MSIX packages.
```
yao-msft, line 413: nit: we may specifically call out it's the sha256 hash of the AppxSignature.p7x file under package or bundle

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,444 @@
...
+ This key must be present for each installer for this version of the package. There may be multiple installer nodes to support different architectures, locales, install scopes (User vs. Machine)
+</details>
+
+### ManifestType
+<details>
+ <summary>The manifest type</summary>
+
+ #### Required Field
+
+ This key must have the value "version". The Microsoft community package repository validation pipelines also use this value to determine appropriate validation rules when evaluating this file.
```
yao-msft, line 434:
```suggestion
 This key must have the value "installer". The Microsoft community package repository validation pipelines also use this value to determine appropriate validation rules when evaluating this file.
```

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,444 @@
...
+
+ This key represents the install modes supported by the installer. The Microsoft community package repository requires a package support "silent" and "silent with progress". The Windows Package Manager also supports "interactive" installers.
+
+ >Note: Some installers will attempt to install missing dependencies. If these dependencies require user interaction, the package will not be allowed into the Microsoft community package repository.
+
+ >Note: This key may be present in the root of the manifest as the default value for all installer nodes. This key may also be present in an individual installer node as well. If this key is in the manifest root and in an installer node, the value in the installer node will apply.
+</details>
+
+### InstallerSwitches
+<details>
```
yao-msft, line 111: Missing ### Silent header

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,444 @@
...
+
+</details>
+
+### PackageVersion
+<details>
+ <summary>The package version</summary>
+
+ #### Required Field
+
+ This key represents the version of the package. It is related to the specific release this manifests targets. In some cases you will see a perfectly formed [semantic](https://semver.org) version number, and in other cases you might see something different. These may be date driven, or they might have other characters with some package specific meaning for example.
```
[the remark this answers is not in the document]
denelon: I do want to reference semver.org "as much as reasonable" to help others understand we are aware of semantic versioning, and it being a "best practice". The other following sentences should clarify we might see "anything". Should I try to make that more clear?

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,444 @@
...
+<details>
+ <summary>The distribution channel</summary>
+
+ #### Optional Field
+
+ This key represents the distribution channel for a package. Examples may include "stable" or "beta".
+
+ >Note: This key is included for future use. The Windows Package Manager currently does not have any behavior associated with this key. The intent behind this key is to help disambiguate the different channels for packages lacking support for side by side installation. Some packages support having more than one package channel available on a system simultaneously. This key is intended to ensure the proper channel for a package is used during install and upgrade scenarios.
+ 
+ >Note: This key may be present in the root of the manifest as the default value for all installer nodes. This key may also be present in an individual installer node as well. If this key is in the manifest root and in an installer node, the value in the installer node will apply.
```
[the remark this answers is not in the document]
denelon: I would "think" we would want to have a single "package" like "Docker.Desktop" with manifests containing channels rather than two distinct packages since they cannot be installed side by side. Two packages makes sense when we can have side by side install like "Microsoft.VisualStudioCode" and "Microsoft.VisualStudioCodeInsider". We would still need a way to determine if a user had installed a particular channel so we don't do the "wrong" upgrade.

doc/manifest/schema/1.0.0/defaultLocale.md
denelon: We're essentially following the stack-overflow conventions. I think it helps avoid sentences for tags. I'll think about this one when I get closer to a merge.

doc/manifest/schema/1.0.0/installer.md
denelon: the other file would be something like "locale.en-US" so I didn't quote the other one, but when the title matches the name of the file I quoted it. I could remove quotes...

doc/manifest/schema/1.0.0/installer.md
denelon: what happens if I try to specify one of the modes not supported by the installer?

doc/manifest/schema/1.0.0/installer.md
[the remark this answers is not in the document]
yao-msft: I mean, we have following sentence saying the "best practice" is to use the Version string in arp entry(which most likely is not a semver). Seems like a conflict with this one.

doc/manifest/schema/1.0.0/installer.md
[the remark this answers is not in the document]
yao-msft: I mean, one manifest can have only one Channel, just like one manifest can have only one Version. They still belong to same package. A different Channel most likely mean the Version is also different.

And in implementation, we already couple the Version and Channel as one concept, we operate on {Version, Channel} pair as one package version. Changing a Version to map to multiple Channels seems a lot of code rewriting involved.

doc/manifest/schema/1.0.0/installer.md
yao-msft: Regarding switches selection, we have fallback logic to pick the most appropriate one. We don't force the installation to quit if desired behavior cannot be met, for now.

doc/manifest/schema/1.0.0/installer.md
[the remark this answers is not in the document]
denelon: That makes sense. I've added more context for clarity.

doc/manifest/schema/1.0.0/installer.md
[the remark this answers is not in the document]
denelon: OK, so we will need some guidance on authoring manifests when channels are present **when they are supported**. We will also need some kind of identifier to identify the "default" or "stable" channel, and we will likely need some kind of description for any other channels. When we perform search, and we receive results for a package with multiple channels, we will need a way to help users identify channels. Maybe show could display channels and descriptions. We will also need a way to ensure the upgrade scenario selects the correct channel, and we will need guidance or other flags to help with migrating between channels.

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,397 @@
...
+</details>
+
+### Channel
+<details>
+ <summary>The distribution channel</summary>
+
+ #### Optional Field
+ This key represents the distribution channel for a package. Examples may include "stable" or "beta".
+
+ >Note: This key is included for future use. The Windows Package Manager currently does not have any behavior associated with this key. The intent behind this key is to help disambiguate the different channels for packages lacking support for side by side installation. Some packages support having more than one package channel available on a system simultaneously. This key is intended to ensure the proper channel for a package is used during install and upgrade scenarios.
```
JohnMcPMS, line 34:
```suggestion
 >Note: This key is included for future use. The Windows Package Manager currently does not have any behavior associated with this key. The intent behind this key is to help disambiguate the different channels for packages lacking support for side by side installation. Some packages support having more than one package channel available on a system simultaneously; in this case it is better to use unique packages rather than channels. This key is intended to ensure the proper channel for a package is used during install and upgrade scenarios.
```

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,397 @@
...
+</details>
+
+### SignatureSha256
+<details>
+ <summary>SignatureSha256 is recommended for appx or msix. It is the sha256 of signature file inside appx or msix. Could be used during streaming install if applicable</summary>
+
+ #### Optional Field
+ This key represents the signature file (AppxSignature.p7x) inside an MSIX installer. It is used to provide streaming install for MSIX packages.
+
+ >Note: MSIX installers must be signed to be included in the Microsoft community package repository. If the installer is an MSIX this signature should be included in the manifest. The [Windows Package Manager Manifest Creator](https://github.com/microsoft/winget-create) can be used to determine the signature SHA 256. The `winget hash &lt;pathToInstaller&gt; --msix` command can also be used to determine the signature SHA 256.
```
JohnMcPMS, line 372:
```suggestion
 >Note: MSIX installers must be signed to be included in the Microsoft community package repository. If the installer is an MSIX this signature should be included in the manifest. The [Windows Package Manager Manifest Creator](https://github.com/microsoft/winget-create) can be used to determine the signature SHA 256. The `winget hash <pathToInstaller> --msix` command can also be used to determine the signature SHA 256.
```

JohnMcPMS (dismissed):
A few minor suggestions.

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,397 @@
...
+ 
+ >Note: This key may be present in the root of the manifest as the default value for all installer nodes. This key may also be present in an individual installer node as well. If this key is in the manifest root and in an installer node, the value in the installer node will apply.
+</details>
+
+### InstallerType
+<details>
+ <summary>Enumeration of supported installer types. InstallerType is required in either root level or individual Installer level</summary>
+
+ #### Required Field
+ This key represents the installer type for the package. The Windows Package Manager supports MSIX, MSI, and executable installers. Some well known formats provide standard sets of installer switches to provide different installer experiences.
```
KevinLaMS, line 62: Feels like we should include the list of well known formats.

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,397 @@
...
+ 
+ >Note: This key may be present in the root of the manifest as the default value for all installer nodes. This key may also be present in an individual installer node as well. If this key is in the manifest root and in an installer node, the value in the installer node will apply.
+</details>
+
+### Scope
+<details>
+ <summary>Scope indicates if the installer is per user or per machine</summary>
+
+ #### Optional Field
+ This key represents the scope the package is installed under. The two configurations are "user" and "machine". Some installers support only one of these scopes while others support both via arguments passed to the installer using "InstallerSwitches".
```
KevinLaMS, line 76: should InstallerSwitches be a hyperlink?

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,397 @@
...
+
+ >Note: When the Windows Package Manager installs a package using the "interactive" install mode, any custom switches will also be passed to the installer. If a user applies override switches via command line via the Windows Package Manager, none of the switches from the manifest will be passed to the installer.
+
+</details>
+
+### InstallLocation
+<details>
+ <summary>InstallLocation is the value passed to the installer for custom install location. </summary>
+
+ #### Optional Field
```
KevinLaMS, line 140: We do not have many samples of the command line with these options.  I wonder if we need to add them.

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,397 @@
...
+ This key represents any switches the Windows Package Manager will pass to the installer in addition to "Silent", "SilentWithProgress", and "Interactive".
+ 
+ >Note: If a user applies override switches via command line via the Windows Package Manager, none of the switches from the manifest will be passed to the installer.
+</details>
+
+### InstallerSuccessCodes
+<details>
+ <summary>List of additional non-zero installer success exit codes other than known default values by winget</summary>
```
KevinLaMS, line 174: what is the delimiter if we have multiple values?  1, 5, 205

doc/manifest/schema/1.0.0/installer.md
```diff
@@ -0,0 +1,397 @@
...
+<details>
+ <summary>The upgrade method</summary>
+
+ #### Optional Field
+ This key represents what the Windows Package Manager should do regarding the currently installed package during a package upgrade. If the package should be uninstalled first, the "uninstallPrevious" value should be specified.
+
+ >Note: This key may be present in the root of the manifest as the default value for all installer nodes. This key may also be present in an individual installer node as well. If this key is in the manifest root and in an installer node, the value in the installer node will apply.
+</details>
+
+### Commands
```
KevinLaMS, line 194: We should have the delimiter listed.

JohnMcPMS (approved):

PR #16997, opened 2021-06-10
Microsoft.VisualStudioCode_v1.57.0.b4c1bd0a9b

lszomoru:
Microsoft.VisualStudioCode_v1.57.0.b4c1bd0a9b

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/16997)


OfficialEsco (changes requested):
Package added with correct version. If you're even reading this, please remove the 4th version identifier as it causes issues..

denelon:
@lszomoru we need to do some work in the manifest to support a "display" package version, and have one match the Add/Remove Programs entry. The fourth part of the version causes `winget upgrade` to land in an infinite upgrade loop. Only the first three parts of the version are recorded in Add / Remove Programs, so the user will always see an update available.

PR #17324, opened 2021-06-12
TabibitoTechnology.Icaros version 3.2.1.0

musm:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/17324)

### DIFF
--- a/manifests/t/TabibitoTechnology/Icaros/3.2.1.0/TabibitoTechnology.Icaros.installer.yaml
+++ b/manifests/t/TabibitoTechnology/Icaros/3.2.1.0/TabibitoTechnology.Icaros.installer.yaml
@@ -1,0 +1,11 @@
+# Created using wingetcreate
+
+PackageIdentifier: TabibitoTechnology.Icaros
+PackageVersion: 3.2.1.0
+Installers:
+- Architecture: x86
+  InstallerType: inno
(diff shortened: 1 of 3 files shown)


musm:
I don't understand why we have a hash mismatch. I verified the has is correct.

musm:
Any help resolving the has mismatch would be appreciated. I verified that they match so it is confusing why I'm getting the error.

musm:
Any help resolving the has mismatch would be appreciated. I verified that they match so it is confusing why I'm getting the error.

musm:
Any help resolving the has mismatch would be appreciated. I verified that they match so it is confusing why I'm getting the error.

denelon:
@musm the installer URL took me to a page where a download key was generated. I suspect that's part of the problem here. It looks like there isn't a "static" URL directly to the package.

musm:
@denelon Yes that looks correct. Unfortunately I am not sure how we can resolve this.

musm:
In the meantime I will inquire if the package author is able to upload the installer to a stable url.

musm:
Huh, it would be good to find an good solution

ItzLevvie:
Oops, I may accidently  added your PR to the list of PRs that needed to be closed for being invalid, but the only good solution here would be to talk to the application developer about it since the download link can only come from the publisher's or ISV's website rather than uploading it to MediaFire, Google Drive, or OneDrive.

PR #17382, opened 2021-06-13
Users/bearmannl/update odt camelcasing

bearmannl:
First step to get the package renamed with proper CamelCasing.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/17382)


bearmannl:
This is going to take more effort than I thought. I'll delete all the versions and the root folder. Add the versions back with the proper CamelCasing one by one, etc.

PR #17565, opened 2021-06-15
update KiCad.KiCad to 5.1.10_1

donid:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/17565)

### DIFF
--- a/manifests/k/KiCad/KiCad/5.1.10_1/KiCad.KiCad.yaml
+++ b/manifests/k/KiCad/KiCad/5.1.10_1/KiCad.KiCad.yaml
@@ -1,5 +1,5 @@
 PackageIdentifier: KiCad.KiCad
-PackageName: KiCad EDA
+PackageName: KiCad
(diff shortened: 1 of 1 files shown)


manifests/k/KiCad/KiCad/5.1.10_1/KiCad.KiCad.yaml
```diff
@@ -0,0 +1,45 @@
+PackageIdentifier: KiCad.KiCad
+PackageName: KiCad EDA
```
OfficialEsco, line 2:
```suggestion
PackageName: KiCad
```

OfficialEsco (changes requested):
Expected the version to be wrong, but its actually right, the PackageName is wrong tho
image

OfficialEsco (approved):

Issue #16041, opened 2021-06-03
Does YamlCreate.ps1 store installers somewhere?


KaranKad:
I would like to know if `YamlCreate.ps1` keeps installer somewhere after generating manifest for an application. This would be helpful for me as I would be able to install the application afterwards to get additional metadata about the application like `ProductCode`, accurate `Publisher`, `PackageName` and `PackageVersion`

OfficialEsco:
This is the code part:

```
    while ([string]::IsNullOrWhiteSpace($InstallerUrl)) {
        Write-Host
        Write-Host -ForegroundColor 'Green' -Object '[Required] Enter the download url to the installer.'
        $InstallerUrl = Read-Host -Prompt 'Url' | TrimString
    }
    Write-Host $NewLine
    Write-Host 'Downloading URL. This will take awhile...' -ForegroundColor Blue
    $WebClient = New-Object System.Net.WebClient

    try {
        $Stream = $WebClient.OpenRead($InstallerUrl)
        $InstallerSha256 = (Get-FileHash -InputStream $Stream -Algorithm SHA256).Hash
    }
    catch {
        Write-Host 'Error downloading file. Please run the script again.' -ForegroundColor Red
        exit 1
    }
    finally {
        $Stream.Close()
    }
```

I don't actually know, but i don't think it saves it somewhere since it will be gone once you do `$Stream.Close`

KaranKad:
Is it possible to add it as a opt in feature, I really don't know much about PowerShell scripting.

OfficialEsco:
> Is it possible to add it as a opt in feature, I really don't know much about PowerShell scripting.

I would gladly add tons of stuff to YamlCreate, however it seems like the WinGet Team is keen on creating `wingetcreate` instead which slows down the merge process and i won't get any suggestions/help with parts of code...

Kinda disappointed `wingetcreate` didn't have the features of YamlCreate + more on release..

But yeah it would be possible to save the file to temp with a few lines

denelon:
@OfficialEsco, I've been talking with the wingetcreate team about extending behavior to include things that speed up the merge process. They are heads down fixing bugs and LOVING your suggestions. I don't know if you are following discussions or not, but https://github.com/microsoft/winget-pkgs/discussions/15951 is an ask to help identify things that you, @jedieaston, and @ItzLevvie would like us to focus on to help with "Catalog-Health".

OfficialEsco:
[...]

PR #17571, opened 2021-06-15
Small YamlCreate.ps1 update

OfficialEsco:
* Rearranged the installer metadata
* Removed spaces
* Added # Created using YamlCreate.ps1
* Added local save
* Added fetch ProductCode

-----

#16041 @KaranKad

I might add `Create simple Manifest` and `Create detailed Manifest` since there is "_too much_" information...

Also i wish i knew how to get ALL the Metadata from the previous Manifests (which is not included in this) so it could support Dependency etc.. Kinda tricky when the Manifests can be different..

https://github.com/microsoft/winget-pkgs/pull/11753#issuecomment-832348007 reminder to self

@palenshus

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/17571)

### DIFF
--- a/Tools/YamlCreate.ps1
+++ b/Tools/YamlCreate.ps1
@@ -1,4 +1,5 @@
 #Requires -Version 5
+$PSVersion = (Get-Host).Version.Major
 
 <#
 .SYNOPSIS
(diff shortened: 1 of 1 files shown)


Tools/YamlCreate.ps1
```diff
@@ -160,38 +161,68 @@ Function Read-PreviousWinGet-Manifest {
...
+            $WebClient.DownloadFile($InstallerUrl, $dest)
+        }
+        catch {
+            Write-Host 'Error downloading file. Please run the script again.' -ForegroundColor Red
+            exit 1
+        }
+        finally {
+            Write-Host "Time taken: $((Get-Date).Subtract($start_time).Seconds) second(s)" -ForegroundColor Green
+            $InstallerSha256 = (Get-FileHash -Path $dest -Algorithm SHA256).Hash
+            if ($PSVersion -eq '5') {$FileInformation = Get-AppLockerFileInformation -Path $dest | Select-Object -ExpandProperty Publisher}
```
palenshus, line 223: Seems like you'd want to do this whether or not the user picked the Save option. Perhaps you always call DownloadFile, and if they said no to save, then you just delete the file after hashing and extracting metadata

Tools/YamlCreate.ps1
```diff
@@ -160,38 +161,68 @@ Function Read-PreviousWinGet-Manifest {
...
         Write-Host
         Write-Host -ForegroundColor 'Green' -Object '[Required] Enter the download url to the installer.'
         $InstallerUrl = Read-Host -Prompt 'Url' | TrimString
     }
+
+    while ([string]::IsNullOrWhiteSpace($SaveOption)) {
+        Write-Host
+        Write-Host -ForegroundColor 'Yellow' -Object "Do you want to save the files to the Temp folder?"
+        $SaveOption = Read-Host -Prompt 'y or n' | TrimString
+    }
```
palenshus, line 190: This functionality feels like a good candidate for a function, something like:

$SaveOption = PromptForValue('Yellow', "Do you want to save the files to the Temp folder? (y/n)")

Then you could use it in various places

Tools/YamlCreate.ps1
```diff
@@ -739,7 +775,9 @@ Write-Host "Yaml file created: $InstallerManifestPath"
 
 Function Write-WinGet-LocaleManifest {
 $LocaleManifest = @(
+'# Created using YamlCreate.ps1'
```
palenshus, line 778: ># Created using YamlCreate.ps1 [](http://example.com/codeflow?start=1&length=30)

Might want to make this a variable instead of hardcoding it multiple times

palenshus (changes requested):
:clock1:

Tools/YamlCreate.ps1
OfficialEsco: We'll its not like its going to change any time soon

Tools/YamlCreate.ps1
OfficialEsco: Well it is a one time select under the `Read-WinGet-InstallerValues` Function, so there is no reason for it to be used anywhere else.
If the script gets a overhaul it will be changed.

palenshus: It could be used on line 181 too :) But that's fine, don't sweat it

palenshus (approved):

denelon (approved):

PR #17847, opened 2021-06-17
step1 removing the folder with improper casing

bearmannl:
After the current versions have been removed, I have the reworked folder names ready and waiting to merge right after it.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/17847)

PR #17863, opened 2021-06-17
OfficeDeploymentTool - version 16.0.12827.20268

bearmannl:
OfficeDeploymentTool - version 16.0.12827.20268

#17847

Adding version 6/6.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/17863)


bearmannl:
Aside from the identifier this is a copy of the version that was added before. No need to fix this since the newer versions are already added? Going to close if all the other versions merge without issue.

bearmannl:
Great, it was the old installer switches that weren't compliant. Fixed and now this closes the ODT package identifier renaming change!

ItzLevvie (approved):

PR #18211, opened 2021-06-20
Rename Powershell-Preview to PowerShell.Preview

bearmannl:
Rename Powershell-Preview to PowerShell.Preview 1 / 24.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/18211)


OfficialEsco:
Ok so it looks like you want to fix this kind of stuff, i would kinda prefer if you update them to a proper manifest once you're at it

bearmannl:
> Ok so it looks like you want to fix this kind of stuff, i would kinda prefer if you update them to a proper manifest once you're at it

Was actually looking to update the PowerShell Preview manifest to add the x86 installer, then thought, might as well do the version naming convention while I'm at it. Don't worry, not targeting any other packages after this today 😉

Will rework the manifests, so feel free to hold off on approving until it's all done.

OfficialEsco (approved):

PR #18505, opened 2021-06-22
add pgAdmin 5.4

dpprdan:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/18505)

### DIFF
--- a/manifests/p/PostgreSQL/pgAdmin/5.4/PostgreSQL.pgAdmin.installer.yaml
+++ b/manifests/p/PostgreSQL/pgAdmin/5.4/PostgreSQL.pgAdmin.installer.yaml
@@ -1,0 +1,24 @@
+# Created using YamlCreate.ps1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+
+PackageIdentifier: PostgreSQL.pgAdmin
+PackageVersion: 5.4
+MinimumOSVersion: 10.0.0.0
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+Installers:
+- InstallerLocale: en-US
+  Architecture: x64
+  InstallerType: exe
+  Scope: machine
+  InstallerUrl: https://ftp.postgresql.org/pub/pgadmin/pgadmin4/v5.4/windows/pgadmin4-5.4-x64.exe
+  InstallerSha256: 486CD09D9F8CE596A2B66AEF7A2DC032126551236A6471D720D681997E243D05
+  InstallerSwitches:
+    Custom: /NORESTART
+    Silent: /VERYSILENT
+    SilentWithProgress: /SILENT
+  UpgradeBehavior: install
+ManifestType: installer
+ManifestVersion: 1.0.0
--- a/manifests/p/PostgreSQL/pgAdmin/5.4/PostgreSQL.pgAdmin.installer.yaml
+++ b/manifests/p/PostgreSQL/pgAdmin/5.4/PostgreSQL.pgAdmin.installer.yaml
@@ -1,4 +1,4 @@
-# Created using YamlCreate.ps1
+# Created using wingetcreate 0.2.0.29
 # yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
 
 PackageIdentifier: PostgreSQL.pgAdmin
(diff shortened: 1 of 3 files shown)


quhxl:
With /SUPPRESSMSGBOXES you can prevent the prompt for the scope. This can then be set via /ALLUSERS or /CURRENTUSER.

dpprdan:
Apparently the pgAdmin installer (inno) supports both user machine scope

dpprdan:
Apparently the pgAdmin installer (inno) supports both user and machine scope as of version 5.4. IIUC from [this discussion](https://github.com/microsoft/winget-pkgs/discussions/16219#discussioncomment-829163), winget does not support both at the same time at the moment, i.e. in a way that one is the default which the user can override? In any case `Scope` can only be one of two according to the [schema](https://github.com/microsoft/winget-cli/blob/1cc58ce1996810b6f2c639b74451f476c7b58244/schemas/JSON/manifests/v1.0.0/manifest.installer.1.0.0.json#L65) and the [docs](https://github.com/microsoft/winget-pkgs/blob/9fdd6dc48694f9e5e24f908a747bfa67dfbca1f8/doc/manifest/schema/1.0.0/installer.md#L76) are not entirely clear on what to do if the installer supports both (i.e. "others support both").

How should I proceed, @msftrubengu @denelon? "Hard code" one scope as suggested by @quhxl and if yes, which one? Or is there a better way, i.e. one with a default and user choice?

BTW The message box to choose the scope only comes up if pdAdmin is not installed yet, which is why I missed it in my local install test (`winget install --manifest <path>`). I only saw it after I tried the sandbox test. Should the sandbox test be encouraged more as default, since it is closer to the tests run on GHA?

denelon:
@dpprdan it is possible to have two installer nodes for scope. One for "user" and the other for "machine". The user can specify the scope via command line arguments: `winget install <package> --scope <user|machine>`. The user can also have preferences or requirements in their settings.

The switches can be specified per installer to provide the two options for users.

The client will default to "user" scope if nothing is specified, and the manifest provides the option.

OfficialEsco:
This is what you should do https://github.com/microsoft/winget-pkgs/pull/19302/files, /CURRENTUSER and /ALLUSERS work for both Nullsoft and Inno Installers. AFAIK I'm the only one who adds this information and i only notice once i run the installer manually...

dpprdan:
Thanks @OfficialEsco, that seems to have fixed it!

manifests/p/PostgreSQL/pgAdmin/5.4/PostgreSQL.pgAdmin.installer.yaml
```diff
@@ -0,0 +1,35 @@
...
+- InstallerLocale: en-US
+  Architecture: x64
+  InstallerType: inno
+  Scope: user
+  InstallerUrl: https://ftp.postgresql.org/pub/pgadmin/pgadmin4/v5.4/windows/pgadmin4-5.4-x64.exe
+  InstallerSha256: 486CD09D9F8CE596A2B66AEF7A2DC032126551236A6471D720D681997E243D05
+  InstallerSwitches:
+    Silent: /VERYSILENT
+    SilentWithProgress: /SILENT
+    Custom: /SUPPRESSMSGBOXES /CURRENTUSER
```
OfficialEsco, line 32: When the InstallerType is Inno/Nullsoft etc you don't need to add Silent and SilentWithProgress switches btw

OfficialEsco (approved):

manifests/p/PostgreSQL/pgAdmin/5.4/PostgreSQL.pgAdmin.installer.yaml
dpprdan: Thanks, good to know. Those switches are already baked into winget, I assume?
I could clean up the existing manifests, but that's probably not necessary?

OfficialEsco: Yup atleast /VERYSILENT and /SILENT type of switches, imo it should also include /SUPPRESSMSGBOXES and /NORESTART (or detect it and prompt user at the end of installation/upgrade), but for now we just use those when needed (/NORESTART most of the time just because its nice and dosn't break anything)
Naaah just clean it up on the next PR :)

dpprdan: Agreed (esp. detect & prompt) and will do.

PR #18546, opened 2021-06-22
Microsoft.dotnetRuntime.3 version 3.1.9.29323

KaranKad:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/18546)

### DIFF
--- a/manifests/m/Microsoft/dotnetRuntime/3/3.1.9.29323/Microsoft.dotnetRuntime.3.yaml
+++ b/manifests/m/Microsoft/dotnetRuntime/3/3.1.9.29323/Microsoft.dotnetRuntime.3.yaml
@@ -1,0 +1,38 @@
+# Created using wingetcreate 0.2.0.29
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.singleton.1.0.0.schema.json
+
+PackageIdentifier: Microsoft.dotnetRuntime.3
+PackageVersion: 3.1.9.29323
+PackageLocale: en-US
+Publisher: Microsoft
+PackageName: .NET Core Runtime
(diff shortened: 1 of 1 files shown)


OfficialEsco:
Hmm, i think the PackageName might have to be `Microsoft Windows Desktop Runtime - 3`
image

KaranKad:
I'll create new PR's then

OfficialEsco:
Tbh it might be better if we get --scope and scope in dependency instead of making 3 different folders/manifest, so yeah never mind that for now

KaranKad:
Yeah I think `Microsoft.dotnetRuntime.3.x64` might be too much `Microsoft.dotnetRuntime.3-x64` might be better.

OfficialEsco:
I think we might have to consult with @denelon before you commit too much 😁
Because `Microsoft Windows Desktop Runtime - 3 (x64)` wont match ARP unless we type the full version, i think `winget upgrade` might just have to update both (which should work? if you run it twice)

KaranKad:
Because x64 and x86 versions appear different in control panel I think they need to have different manifest and I tried to kept `PackageName` closest to its ARP because version shouldn't be in `PackageName`. I've created an issue at [winget-cli/issues/982](https://github.com/microsoft/winget-cli/issues/982) to introduce `AppName` for per version of the application so winget can show you proper name using `PackageName` and can match the ARP using `AppName` field.

So in case of this application 
`PackageName` would be `Microsoft Windows Desktop Runtime - 3 (x64)` and 
`AppName` would be `Microsoft Windows Desktop Runtime - 3.1.16 (x86)` to better match ARP

PR #18598, opened 2021-06-23
Microsoft.Ironpython3 version 3.4.0.0001

oxygen-dioxide:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/18598)

### DIFF
--- a/manifests/m/Microsoft/Ironpython3/3.4.0.0001/Microsoft.Ironpython3.installer.yaml
+++ b/manifests/m/Microsoft/Ironpython3/3.4.0.0001/Microsoft.Ironpython3.installer.yaml
@@ -1,0 +1,15 @@
+# Created using wingetcreate 0.2.0.29
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+
+PackageIdentifier: Microsoft.Ironpython3
+PackageVersion: 3.4.0.0001
+Installers:
+- InstallerLocale: en-US
+  Architecture: x64
+  InstallerType: msi
+  InstallerUrl: https://github.com/IronLanguages/ironpython3/releases/download/v3.4.0-alpha1/IronPython-3.4.0a1.msi
+  InstallerSha256: 877995FE14C29F5D9CCD01BD3A65A3C78A2C369E98F860D1F3DC7152F9D552BE
+  ProductCode: '{51C8A31F-CB45-4454-B586-218AC7CD3DE4}'
+ManifestType: installer
+ManifestVersion: 1.0.0
+
--- a/manifests/m/Microsoft/Ironpython3/3.4.0.0001/Microsoft.Ironpython3.locale.en-US.yaml
+++ b/manifests/m/Microsoft/Ironpython3/3.4.0.0001/Microsoft.Ironpython3.locale.en-US.yaml
@@ -1,0 +1,13 @@
+# Created using wingetcreate 0.2.0.29
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.0.0.schema.json
+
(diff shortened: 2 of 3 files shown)


jedieaston (commented):
@OfficialEsco @ItzLevvie WinGet is picking this package up as `Microsoft.Ironpython2`, and I can't figure out why. Does anybody have a hypothesis? (I tried changing PackageName to what it appears as in the ARP table, that didn't help). I'm fine with merging this but upgrades will be broken for both IronPython2 and IronPython3 until this is resolved. 

(ps @denelon Are there any docs in microsoft/winget-cli that explain exactly how the ARP matching works? If not that might be something for the backlog, because it'd help a ton with diagnosing these types of issues.)

jedieaston (approved):

OfficialEsco:
> @OfficialEsco @ItzLevvie WinGet is picking this package up as `Microsoft.Ironpython2`, and I can't figure out why. Does anybody have a hypothesis? (I tried changing PackageName to what it appears as in the ARP table, that didn't help). I'm fine with merging this but upgrades will be broken for both IronPython2 and IronPython3 until this is resolved.
> 
> (ps @denelon Are there any docs in microsoft/winget-cli that explain exactly how the ARP matching works? If not that might be something for the backlog, because it'd help a ton with diagnosing these types of issues.)

Every Publisher and PackageName in that Package folder looks for a matching ARP, so [Microsoft/Ironpython2/2.7.11.1000/Microsoft.Ironpython2.yaml](https://github.com/microsoft/winget-pkgs/blob/master/manifests/m/Microsoft/Ironpython2/2.7.11.1000/Microsoft.Ironpython2.yaml) have `PackageName: IronPython` which is causing this match.

If `PackageName: IronPython 2` and `PackageName: IronPython 3` actually works we should use that (Some fails because of `.` and `-` in the package name)
This should also be moved to `Microsoft.IronPython.2` and `Microsoft.IronPython.3` where the manifests have a consistent `Publisher` and `PackageName`

jedieaston:
But why does WinGet show the name "IronPython" instead of "IronPython 3.4.0a1" as it appears in the ARP table? Where does it get that value from? (This happens with a lot of apps, not just this one, where the name WinGet comes up with != the name in the ARP table or the name in the package).

OfficialEsco:
It should show the `PackageName` from the latest Manifest (gimme a few examples), WinGet actually shows very little ARP Information without us providing it, if `Publisher` and `PackageName` matches it will show the `PackageIdentifier` which controls everything else. `Publisher` and `PackageName` by default looks for `PackageName/Publisher "wildcard"`, so we don't actually have to type the whole thing, but it helps.

jedieaston:
I didn't know it added a wildcard (thought that was just something that was proposed), so thanks, that makes a lot of sense. So `winget list` looks in the package source(s) for the "actual" name to display? 

(`Automattic.Wordpress` was the immediate example I thought of, but that does what you say, since `WordPress.com*` would match `WordPress.com 6.15.0`.)

OfficialEsco:
It's most likely a wildcard made for version numbers in the name because from experience if I don't remember wrong `WordPress*` would fail but `WordPress.com *` works (aka you need the space...)  

`WinGet list` displays the PackageName from the latest manifest probably because it's showing that one, but if you do `winget show "packageid" --version` it will show the PackageName from that specific manifest.

The suggestion is so we can use wildcards in the Manifest, like `Mozilla Firefox * (en-US)` or `IronPython 3*` in this case.

denelon:
We're getting ready to work on the https://github.com/microsoft/winget-cli/issues/1073 specification 
I'll add some notes to make sure we share the logic

PR #18940, opened 2021-06-26
Wolicheng.WoMicDriver version 1.6.0.0

AlexZorzi:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/18940)


quhxl:
I don't think there is a silent installer for this app/driver.

AlexZorzi:
@quhxl 
When I tried to install it gets succesfully installed but at the end of the setup it the program asks if the user wants to check for updates
And after that it quits with error 2

quhxl:
I think that's enough. It's not completely silent.

OfficialEsco:
image
This is actually why it fails, but this is a dependency that we can't install i believe, so this might have to go under not 100% unattended since as you said this also pops up
image

PR #19297, opened 2021-06-29
Cygwin.Cygwin version 3.2.0

onerandomusername:

-----


houstdav000:
This would be great to have, but I'm not sure there's a good way to do it without specifying a specific mirror site like with

```cmd
setup-x86_64.exe -q -W -O -s "http://www.gtlib.gatech.edu/pub/cygwin/"
```

If a script written to select a mirror from [https://cygwin.com/mirrors.lst](https://cygwin.com/mirrors.lst) was used, then you could pass that option to the installer, but I'm not sure there's a way to do it with manifests. Still pretty new to this though.

ItzLevvie (commented):
Stuck here, could you please add switches for silent install?
image

PR #19389, opened 2021-06-30
Visual Studio 2019 uses the actual version for the ARP table now instead of the build version

coehlrich:
Since the version number in the ARP table used the actual version instead of the build version winget might think that this is for an older version of Visual Studio 2019 since the build version for the previous release was 16.10.31410.357 and the actual version for the current version of Visual Studio 2019 is 16.10.3

Same thing also happened with build tools (And most likely other versions that the Visual Studio Installer has)

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/19389)

### DIFF
--- a/manifests/m/Microsoft/VisualStudio/2019/Community/16.10.3/Microsoft.VisualStudio.2019.Community.yaml
+++ b/manifests/m/Microsoft/VisualStudio/2019/Community/16.10.3/Microsoft.VisualStudio.2019.Community.yaml
@@ -1,6 +1,6 @@
 # yaml-language-server: $schema=https://aka.ms/winget-manifest.singleton.1.0.0.schema.json
 PackageIdentifier: Microsoft.VisualStudio.2019.Community
-PackageVersion: 16.10.31424.327
+PackageVersion: 16.10.3
 PackageName: Visual Studio Community 2019
 Author: Microsoft
 Publisher: Microsoft Corporation


jedieaston:
Did this change recently? I could've sworn VS used to used the build number in the ARP table.

coehlrich:
Yes it seems to have been changed in the last release and the version displayed is the actual version on both the computer where I updated and in the sandbox where I installed the newer version.

jedieaston:
If this only happens for the most recent version I wonder why. Seems like a regression. We'd have to move all of the other ones for it to work but then they won't match correctly.

I wonder if we should wait on microsoft/winget-cli#1073 to do this so we can put the shorter version number in the PackageVersion box and just set the version in the programs array to whatever it identifies itself as in the ARP table. 

Another moderator got an opinion? @OfficialEsco @ItzLevvie

coehlrich:
I did try to install an older version in a sandbox using winget after accidentally starting a install for the newest version but it installed the newest version and it seems like installations still on an older version still had the build version in the ARP table based on when I had kept Apps and Features open while upgrading my installation

jedieaston:
Yeah I see that on my other box that has the last build of VS Pro on it. This is odd.

OfficialEsco:
Think you need to get a answer from @ms-jb or @lszomoru
@denelon contacted them to fix the VS Code version number and since then they have been updating a few things

lszomoru:
I am managing the publishing pipeline for Visual Studio Code, and this seems to be a question related to Visual Studio.

jedieaston:
@denelon Can you put a Blocking-Issue or something on this so it doesn't get merged until we can figure this out? Thanks!

denelon:
I've added the "Blocking Issue" label. I've reached out to the Visual Studio team. They are holding for us to implement the package ownership feature to begin automated manifest publishing. We're expecting that feature in the next week or so to hit production. I'm not sure how soon they will be prepared to automate. I'll also ask about the version number changes in ARP.

ItzLevvie (changes requested):
https://github.com/microsoft/winget-pkgs/pull/19389#issuecomment-871504524

jedieaston:
Hey @msftbot, we're still waiting on stuff!

OfficialEsco:
bump

jedieaston:
@msftbot don't close this!

KevinLaMS:
@jedieaston  I have confirmed with Visual Studio that the versioning change was intentional.  The PM that drove the decision is out of the office, so I do not know the rationale. 
I think we should move forward with this being their pattern from here on out.

jedieaston:
@coehlrich I did all of them and they are merged, if you want to close this.

PR #19739, opened 2021-07-04
miHoYo.GenshinImpact version 1.6

NemmisPrime-07:


quhxl:
I don't think it can be installed silent.

PR #20696, opened 2021-07-12

vedantmgoyal9:
null


ItzLevvie:
https://github.com/microsoft/winget-create can find the application's InstallerType and the switches in the manifest depends on InstallerType as per https://github.com/microsoft/winget-cli/blob/cf6f8e58195bf033a1a5872f7b800866cd27a586/src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp#L348-L380.

vedantmgoyal2009:
But everyone (like me) don't use `wingetcreate`.

vedantmgoyal2009:
I manually author manifests or prefer `YAMLCreate.ps1`

ItzLevvie:
Most applications support `/?` to pass to the installer so you can find installer switches easily, and if it's not available, you can search for installer switches online, but not all installers will work with this program.

Companies create custom parameters to the installer to skip EULA which this program won't capture, you would have to check the documentation for that. For example,  VMWare Workstation is a MSI file inside an EXE file but it won't install unattended because there's an interactive EULA blocking the install, so you would've to do `ACCEPT_EULA=1` or something similar.

vedantmgoyal2009:
On doing `/?`, it launches GUI: 
image

vedantmgoyal2009:
@denelon ??

ItzLevvie:
>On doing /?, it launches GUI

Not all programs support this, some or most do. Git is one example of this. If it doesn't work, try using `wingetcreate` or otherwise search on the internet for it.

>But everyone (like me) don't use `wingetcreate`.

Microsoft has no control over third-party applications and `wingetcreate` is a supported way of determining what the installer type for an application is. AppGet follows the same (with no third-party tools needed), see https://github.com/appget/appget/tree/37e71b7f8405f5514f2b41563e38008de97f1162/src/AppGet/Installers/InstallerWhisperer.

>I manually author manifests or prefer YAMLCreate.ps1

Before `wingetcreate` existed, everyone created the manifest using a tool or manually, but you still have to test the application in in a virtual machine to see if it can install unattended; there's a reason why there's checkboxes in PR.

This PR isn't any different or better than what we've today, installer switches are automatically added by winget-cli depending on `InstallerType` (as per https://github.com/microsoft/winget-cli/blob/cf6f8e58195bf033a1a5872f7b800866cd27a586/src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp#L348-L380) and that makes Universal Silent Switch Finder redundant and a bit complicated.

denelon:
@vedantmgoyal2009 I think @ItzLevvie probably gave you the best advice. Some programs just don't support silent install. We would love to get to a point where supported tooling does as much as possible to generate a high quality manifest. The YAML Create was a community contribution to help with the process. We would love to get to the point where supported tooling exceeds those capabilities, but we're still a bit away from that. We aren't trying to tell anyone how to create manifests, we're just trying to make it easier.

The day when @OfficialEsco thinks the tooling is as good as or better than what the current process is, I'll celebrate. The requirements for building and releasing software by Microsoft are fairly complex. Globalization, Localization, Internationalization, Security, and Privacy are just a subset of the concerns. Due to those requirements we often aren't able to build and release code as quickly as we would like. 

It will take the entire community to support and promote the Windows Package Manager and our shared desired experiences to ISVs and publishers. We will continue to build the best product and the best user experience we can based on the current state of the installers we have. When customers show that they want the products, and they want them to work well in the Windows Package Manager then the producers may be motivated to make the changes necessary. In some cases they may not, it's their choice. We've already seen a few exceptions where publishers have intentionally requested not to be included in the catalog. Some of them generate the revenue to cover some of their costs by having users go to their websites and view their sponsors advertisements. We respect their choices.

OfficialEsco (approved):

ghost (approved):

PR #20907, opened 2021-07-14

vedantmgoyal9:
null


ItzLevvie (changes requested):
If the manifest has the correct version, it should be preserved for application downgrading purposes.

Older Visual Studio Code Insider manifests were removed because it had an incorrect version compared to what's shown in Control Panel.

ItzLevvie (approved):

PR #21028, opened 2021-07-15
Added Amazon.AWSCLI version 2.2.19.0

ImJoakim:
image

```
PS C:\Users\WDAGUtilityAccount\Desktop\manifests> winget install -m "C:\Users\WDAGUtilityAccount\Desktop\manifests\a\Amazon\AWSCLI\2.2.19.0"
Found AWS Command Line Interface v2 [Amazon.AWSCLI]
Downloading https://awscli.amazonaws.com/AWSCLIV2-2.2.19.msi
  ██████████████████████████████  26.6 MB / 26.6 MB
Starting package install...
Successfully installed
```

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/21028)

### DIFF
--- a/manifests/a/Amazon/AWSCLI/2.2.19.0/Amazon.AWSCLI.installer.yaml
+++ b/manifests/a/Amazon/AWSCLI/2.2.19.0/Amazon.AWSCLI.installer.yaml
@@ -1,0 +1,26 @@
+# Created using wingetcreate 0.2.0.29
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+
+PackageIdentifier: Amazon.AWSCLI
+PackageVersion: 2.2.19.0
+MinimumOSVersion: 10.0.0.0
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+Commands:
+- aws
+Installers:
+- Platform:
+  - Windows.Desktop
+  Architecture: x64
+  InstallerType: msi
+  Scope: machine
+  InstallerUrl: https://awscli.amazonaws.com/AWSCLIV2-2.2.19.msi
+  InstallerSha256: 5FF4ECA8B105FBBB8CC69478961476734EE216BA4FF3B8CA42C05AFF06062C6B
+  ProductCode: '{CBFC876C-E75D-4206-B907-8EB4EDFABA28}'
+  InstallerLocale: en-US
+  UpgradeBehavior: install
+ManifestType: installer
+ManifestVersion: 1.0.0
+
--- a/manifests/a/Amazon/AWSCLI/2.2.19.0/Amazon.AWSCLI.locale.en-US.yaml
+++ b/manifests/a/Amazon/AWSCLI/2.2.19.0/Amazon.AWSCLI.locale.en-US.yaml
@@ -8,7 +8,7 @@ Publisher: Amazon Web Services
 PublisherUrl: https://github.com/aws/aws-cli
 PublisherSupportUrl: https://github.com/aws/aws-cli/issues
 Author: Amazon Web Services
-PackageName: AWS Command Line Interface v2
+PackageName: AWS Command Line Interface
 PackageUrl: https://aws.amazon.com/cli/
 License: Apache 2.0 license
 LicenseUrl: https://github.com/aws/aws-cli/blob/develop/LICENSE.txt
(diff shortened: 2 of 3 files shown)


ItzLevvie (changes requested):
The reason why `v2` was removed was because it messes up `winget list` for those that have `v1` installed on their machine.

OfficialEsco:
> The reason why `v2` was removed was because it messes up `winget list` for those that have `v1` installed on their machine.

Isn't v1 and v2 2 different softwares? or is v2 just a newer software which you want the v1 users to update to?

jedieaston (approved):

jedieaston:
Yes, let me fix that...

ItzLevvie:
> Isn't v1 and v2 2 different softwares? or is v2 just a newer software which you want the v1 users to update to?

They are both different versions but the issue is that having the version displayed will also show the `v2` name for those who've the `v1` version of AWS CLI installed as this is what winget uses to display the name to everyone.

If you named this package as `AWS Command Line Interface 2.2.19.0` those that have v1 installed and did `winget list`, it would show up as 
```
AWS Command Line Interface 2.2.19.0                         1.0.0.0 (example)                        2.2.19.0 (latest version)
```

This would confuse people into thinking that they've `2.2.19.0` but instead they have `1.0.0.0`.

OfficialEsco:
That is not the intended behaviour, winget-cli uses all manifest in the Software folder for metadata matching, so if one locale contains `PackageName: AWS Command Line Interface` it will match against any `AWS Command Line Interface *wildcard*` until its renamed to `AWS Command Line Interface v2`

As far as i can see
<------- This dummy
is guilty in version 2.2.2.0

But yeah we should split them into v1 and v2 eitherway

ItzLevvie:
🤷. It might be on my end but it seems to be happening to K-Lite Codec Pack, it picks up the name from the latest manifest (from https://github.com/microsoft/winget-pkgs/blob/master/manifests/c/CodecGuide/K-LiteCodecPackBasic/16.3.0/CodecGuide.K-LiteCodecPackBasic.locale.en-US.yaml#L8) rather than the one installed on the machine (from https://github.com/microsoft/winget-pkgs/blob/master/manifests/c/CodecGuide/K-LiteCodecPackBasic/16.2.5/CodecGuide.K-LiteCodecPackBasic.locale.en-US.yaml#L7).

Install an old version:
image

`winget list` shows:
image

OfficialEsco:
> It might be on my end but it seems to be happening to K-Lite Codec Pack which has the PackageVersion in the PackageName.

The PackageName is intended because K-Lite have tons of package versions with a bad naming schema, however yes you are correct that winget-cli shows the Metadata from the latest Manifest, this is probably done for a few reasons, however K-Lite `Codec Pack 16.2.5 Basic` will not be detected in `winget list` unless the `PackageName` in that version of the Manifest is `K-Lite Codec Pack 16.2.5 Basic`

This is something that the winget-cli team have to solve, we're just the workaround people 🤷‍♂️
cc @denelon feature idea
Maybe we should have a Standard Manifest Metadata yaml file and version specific Metadata?
```
- Publisher
-- Software
---  Version 1
----    version.yaml
----    locale.yaml
----    installer.yaml
---  Version 2
----    version.yaml
----    locale.yaml
----    installer.yaml
---  locale.yaml
```

ItzLevvie:
>however K-Lite Codec Pack 16.2.5 Basic will not be detected in winget list unless the PackageName in that version of the Manifest is K-Lite Codec Pack 16.2.5 Basic

I'll do a PR test in a couple of days to see if that's the case, if it breaks, then we'll have to revert the PR and wait for the WinGet team to implement better ARP matching in manifests because the way it's currently set up would confuse people into thinking that they've the latest version installed. but in reality, they've the old version of it installed when doing `winget list` and it also shows up in third-party GUIs (like https://winstall.app) a bit weirdly.

If I recall correctly, there was an application that had `blah <version> blah` similar to the K-Lite Codec Pack manifests and when the `<version>` part was removed, it still detected it in `winget list` and I believe there was no `ProductCode` in that manifest back when I made a PR for it.

denelon:
I'm hoping when we get the "AppsAndFeatures" schema bit implemented, this will solve most of these problems. There will likely be a ton of new PRs to change PublisherName and PackageName back to the "Marketing" values. The tougher one is figuring out how to handle the "Marketing" versions. I think JetBrains uses the Year in the marketing version, but the actual version is like 211.y.z or something like that.

denelon:
>This is something that the winget-cli team have to solve, we're just the workaround people 🤷‍♂️
>cc @denelon Demitrius Nelon FTE feature idea
>Maybe we should have a Standard Manifest Metadata yaml file and version specific Metadata?

You are the rock stars who are making the best experiences possible!

We considered the idea of a standard file, and version specific files, but the complexity was pretty messy when considering things like major new features in newer versions, and deprecated or obsolete features in older versions. We also found examples of license changes and publisher name changes. It seemed better to keep the metadata for a version all in one location. It was a decision where there didn't seem to be "one best option", so we went with the one that appeared to be more correct at the expense of more files, and offered a deterministic way to figure out which file(s) needed to change or be copied.

PR #21996, opened 2021-07-21

vedantmgoyal9:
null


KaranKad:
Because the exe version in control panel shows version as `21.02 alpha` and msi version shows as `21.02.00.0`, to winget they seem completely different versions and thus cause problem using winget upgrade.

quhxl:
Please add the ProductCodes...

vedantmgoyal2009:
Ok, will add tomorrow

ItzLevvie (approved):

PR #22738, opened 2021-07-29
​

ItzLevvie:
​


OfficialEsco:
Wait what? I swear i installed this last night

ItzLevvie:
There might be a issue here, I can't figure out the difference between `android-studio-ide` and `android-studio` both mean even though both versions are 2020.3.

On the Android Studio website, you get a download link for 2020.3, https://developer.android.com/studio#downloads but what about 4.1 (which is also 2020.3) and 4.2? cc @tiziodcaio

PR #25010, opened 2021-08-14
Update: Joplin.Joplin version 2.3.3

vedantmgoyal9:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/25010)

### DIFF
--- a/manifests/j/Joplin/Joplin/2.3.3/Joplin.Joplin.installer.yaml
+++ b/manifests/j/Joplin/Joplin/2.3.3/Joplin.Joplin.installer.yaml
@@ -25,6 +25,8 @@ Installers:
   Scope: machine
   InstallerUrl: https://github.com/laurent22/joplin/releases/download/v2.3.3/Joplin-Setup-2.3.3.exe
   InstallerSha256: 7071A013D803D8AD68706E161B191480CF52F8CDC35C3DE72F4A2E581A74B4E5
+  InstallerSwitches:
+    Custom: /ALLUSERS /RESTARTAPPLICATIONS
   UpgradeBehavior: install
 ManifestType: installer
 ManifestVersion: 1.0.0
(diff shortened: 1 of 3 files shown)


OfficialEsco (approved):

laurent22:
@vedantmgoyal2009, it came to my attention that the winget Joplin package also deploys pre-releases. In general that should be avoided because these pre-releases are mainly for beta testing, and can contain bugs or incompatibilities with other Joplin versions.

For example the latest desktop pre-release 2.3 cannot sync with the current Joplin iOS or CLI applications. This will be solved once the official release is out, but for now they are incompatible. So would it be possible to update your process so that pre-releases are not deployed by winget?

OfficialEsco:
This is the reason we have added the Pre Releases https://github.com/microsoft/winget-pkgs/pull/13668#issuecomment-846431219

laurent22:
In general it's relatively safe to install pre-releases, but there are cases where it has to be a user decision. For example in the case I mentioned above if you use the iOS version, you shouldn't install the desktop pre-release because you will no longer be able to sync with your phone until the new iOS app is released.

This situation is rare (it happened twice so far) but can be troublesome enough that it'd be best to keep pre-releases out of the main channel. I hope that makes sense if you need further clarifications let me know.

OfficialEsco:
Hmm okay, would you prefer having the Pre Releases in another "package" or that they use the Pre Release toggle to get them?

laurent22:
@OfficialEsco, I've answered in the pull request thread. I think other packages, such as Chocolatey, simply ignore pre-releases and only deploy actual releases.

OfficialEsco:
Chocolatey does have a Pre-Release branch, however i'm not sure how it works tho, the Package Maintainer probably chooses if they want to use Pre-Release or not.

PR #14722, opened 2021-05-30
New version: uvncbvba.UltraVnc version 1.3.2

OfficialEsco:
closes #14699

- [X] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)?
- [X] Have you checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/change?
- [X] Have you validated your manifest locally with `winget validate --manifest <path>`? 
- [X] Have you tested your manifest locally with `winget install --manifest <path>`?
- [X] Does your manifest conform to the [1.0 schema](https://github.com/microsoft/winget-cli/blob/master/doc/ManifestSpecv1.0.md)?

https://github.com/microsoft/winget-cli/issues/163

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/14722)

PR #25392, opened 2021-08-18

vedantmgoyal9:
null


ItzLevvie (changes requested):
Please test manifests in VM.

EDIT: Last time I checked, the two commented out installers failed validation due to driver installation prompt.

ItzLevvie (changes requested):
https://github.com/microsoft/winget-pkgs/pull/14722#issuecomment-852330774 - driver dependency issue.

Someone needs to double-check these EXE installers by installing it in a VM. I think Windows Sandbox should be enough because it caught the NextDNS driver issue which was not reproducible in VirtualBox or VMware.

ItzLevvie (changes requested):
image

ItzLevvie (approved):

PR #25553, opened 2021-08-20
New Package: JetBrains.DataSpell version 213.1558.17

vedantmgoyal9:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/25553)

### DIFF
--- a/manifests/j/JetBrains/DataSpell/213.1558.17/JetBrains.DataSpell.installer.yaml
+++ b/manifests/j/JetBrains/DataSpell/213.1558.17/JetBrains.DataSpell.installer.yaml
@@ -1,0 +1,19 @@
+# Created with YamlCreate.ps1 v2.0.0 using YAML parsing
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+
+PackageIdentifier: JetBrains.DataSpell
+PackageVersion: 213.1558.17
+MinimumOSVersion: 10.0.0.0
+InstallModes:
+- interactive
+- silent
+Installers:
+- InstallerLocale: en-US
+  Architecture: x64
+  InstallerType: nullsoft
+  Scope: machine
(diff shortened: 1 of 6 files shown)


jedieaston (changes requested):
The JetBrains site seems pretty clear that this is a private installer intended only for people in the private beta. Are we good to put this in the repo? (I currently do not have an invite so I can't confirm if there is an NDA or not).

denelon:
I'd agree with @jedieaston on this one. Let's hold off at least until it's a public beta.

vedantmgoyal2009:
I am posting this comment to remove `Needs: author feedback` and `Needs-Author-Feedback` labels. Also, I have a question: Is there any reason that there are two labels for the same purpose.

denelon:
@vedantmgoyal2009 I prefer labels without spaces as it makes it easier to "search" issues. We had code to modify and it was better to support both until it's implemented everywhere. We should be removing the label without spaces fairly soon [tm].

vedantmgoyal2009:
> I'd agree with @jedieaston on this one. Let's hold off at least until it's a public beta.

It is public beta now. Check https://www.jetbrains.com/dataspell

jedieaston:
That's probably fine then, although you'll have to update this PR to the new build.

ItzLevvie (approved):

vedantmgoyal2009:
> @vedantmgoyal2009 I prefer labels without spaces as it makes it easier to "search" issues. We had code to modify and it was better to support both until it's implemented everywhere. We should be removing the label without spaces fairly soon [tm].

@denelon So, you have implemented this?

denelon:
@vedantmgoyal2009 yes, I was just combing through the bot used to add most of the labels, and I've updated everything I've seen so far.

PR #25847, opened 2021-08-25
New version: Microsoft.PowerShell.Preview version 7.2.0.9

vedantmgoyal9:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/25847)

### DIFF
--- a/manifests/m/Microsoft/PowerShell/Preview/7.2.0.9/Microsoft.PowerShell.Preview.installer.yaml
+++ b/manifests/m/Microsoft/PowerShell/Preview/7.2.0.9/Microsoft.PowerShell.Preview.installer.yaml
@@ -1,0 +1,27 @@
+# Created with YamlCreate.ps1 v2.0.0 using YAML parsing
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+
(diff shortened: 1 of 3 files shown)


jedieaston (changes requested):
winget is having issues matching it now that they've changed the name in the ARP table, does the wildcard it uses not work when there are dashes in the name? If so, I'll merge this, since it's out of our hands, (unless someone feels like opening a bug in PowerShell/PowerShell asking why the name was changed?)


image

OfficialEsco:
Most of our Pre Release/Beta/Dev/Alpha Packages don't actually match ARP.

Yes the Wildcard does not work when there are dashes in the name, which is why we add ProductCode for some exe's.

What was the previous PackageName in ARP for PS7 Preview? I think it might have always been pure ProductCode matching instead of PackageName/Publisher matching

jedieaston:
It was the same thing, you're right it's ProductCode only. I wonder why winget can't match on the ProductCode when it's using a local manifest.

:shipit:

jedieaston (approved):

jedieaston (approved):

PR #26083, opened 2021-08-27
New version: Microsoft.vc++2015-2019.x64 14.29.30133.0

lychichem:

-----

I really what to use https://aka.ms/vs/16/release/vc_redist.x64.exe as download link to let wingetbot to automatically update this package, as I hate manually check the version. But it is not allowed by the principle now. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/26083)

### DIFF
--- a/manifests/m/Microsoft/VC++2015-2019Redist-x64/14.29.30040.0/Microsoft.VC++2015-2019Redist-x64.installer.yaml
+++ b/manifests/m/Microsoft/VC++2015-2019Redist-x64/14.29.30040.0/Microsoft.VC++2015-2019Redist-x64.installer.yaml
@@ -1,0 +1,25 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+PackageIdentifier: Microsoft.VC++2015-2019Redist-x64
+PackageVersion: 14.29.30040.0
+MinimumOSVersion: 10.0.0.0
+InstallerSuccessCodes:
+  - 3010
+InstallModes:
+  - interactive
+  - silent
+  - silentWithProgress
+Installers:
+  - Architecture: x64
+    InstallerType: exe
+    InstallerUrl: https://download.visualstudio.microsoft.com/download/pr/36e45907-8554-4390-ba70-9f6306924167/97CC5066EB3C7246CF89B735AE0F5A5304A7EE33DC087D65D9DFF3A1A73FE803/VC_redist.x64.exe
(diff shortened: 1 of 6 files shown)


OfficialEsco (commented):
Why are you deleting the old version if the old version is still up? This is why dynamic urls are not allowed

lychichem:
> Why are you deleting the old version if the old version is still up? This is why dynamic urls are not allowed

The vc++ redist package will keep only the latest version for a main version number when manually updates. So I always change the folder with greatest version number to the latest version number. If it really affects the work of automatic update, I will keep old versions next time. Sorry for that.

OfficialEsco:
> 
> 
> > Why are you deleting the old version if the old version is still up? This is why dynamic urls are not allowed
> 
> The vc++ redist package will keep only the latest version for a main version number when manually updates. So I always change the folder with greatest version number to the latest version number. If it really affects the work of automatic update, I will keep old versions next time. Sorry for that.

What do you mean? When installing VC++ it will uninstall the previous version? This is what most programs do.

There is multiple reasons why we want older version in the repo

The most important one is that VC++ is "broken", it NEEDS the ProductCode for matching, so if you keep removing the older version they will not get this update with `winget upgrade`.

Imo if you can find older versions and add them with a ProductCode that would make A LOT of people update their VC++

lychichem:
> > 
> > 
> > > Why are you deleting the old version if the old version is still up? This is why dynamic urls are not allowed
> > 
> > The vc++ redist package will keep only the latest version for a main version number when manually updates. So I always change the folder with greatest version number to the latest version number. If it really affects the work of automatic update, I will keep old versions next time. Sorry for that.
> 
> What do you mean? When installing VC++ it will uninstall the previous version? This is what most programs do.
> 
> There is multiple reasons why we want older version in the repo
> 
> The most important one is that VC++ is "broken", it NEEDS the ProductCode for matching, so if you keep removing the older version they will not get this update with `winget upgrade`.
> 
> Imo if you can find older versions and add them with a ProductCode that would make A LOT of people update their VC++

ok，i will try to keep all versions from now, as I need to first Uninstall the latest version and install from the first version to find product code, which is impossible for me because some programs I installed require vc++ libs.

OfficialEsco:
Use Sandbox or a VM

OfficialEsco:
`The pull request contains more than one manifest. Either singleton or multi file manifest under one version directory.`

PR #26224, opened 2021-08-29
​

ItzLevvie:
​


ItzLevvie (commented):
Installer URL points to the latest version, so the hash will always change. This PR switches the URL to point to a specific version to match `PackageVersion`.

OfficialEsco (approved):

PR #27205, opened 2021-09-08

vedantmgoyal9:
null


ItzLevvie (approved):

ItzLevvie (changes requested):
Oops, pressed the wrong button.

This PR doesn't address the reboot issue.

vedantmgoyal2009:
> This PR doesn't address the reboot issue.

It worked on my machine.

ItzLevvie (changes requested):
It wants you to provide installer switches to prevent rebooting which you can get from https://docs.microsoft.com/en-us/windows/win32/msi/standard-installer-command-line-options.

You can reproduce this issue by running the installer with the `/passive` switch without using Windows Package Manager.

vedantmgoyal2009:
Should the package be removed?

arjun-msft:
We are seeing error 3010 in the log files, which means the installation needs a reboot.  We are tracking this via [this issue](https://github.com/microsoft/winget-cli/issues/994).

ItzLevvie (approved):

PR #27374, opened 2021-09-10
​

ItzLevvie:
​


KaranKad (dismissed):

ChuckFerring:
It looks like this may be a 64bit application but the installer is specified as x86. Please verify that the architecture is correct.

Masamune3210:
WinFsp installs at least on my box as x86

palenshus:
Looks like this installer contains exes for both x86 and x64. Validation failed because the command-line tools return an error code when run with no arguments. I think in this case, the architecture being set to x86 would be better to allow more flexibility

manifests/w/WinFsp/WinFsp/1.9.21096/WinFsp.WinFsp.yaml
```diff
@@ -10,10 +10,11 @@ Tags:
 ShortDescription: Windows File System Proxy - FUSE for Windows
 PackageUrl: https://github.com/billziss-gh/winfsp
 Installers:
-- Architecture: x86
+- Architecture: x64
```
palenshus, line 13: Sorry, looks like this installer supports both x86 and x64, so we should specify x86 for flexibility

```suggestion
- Architecture: x86
```

palenshus:
Added waiver

/azp run

ImJoakim (approved):

PR #28871, opened 2021-09-28
Metadata: NSIS.NSIS 3.08

Trenly:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/28871)

### DIFF
--- a/manifests/n/NSIS/NSIS/3.08/NSIS.NSIS.installer.yaml
+++ b/manifests/n/NSIS/NSIS/3.08/NSIS.NSIS.installer.yaml
@@ -1,0 +1,20 @@
+# Created with YamlCreate.ps1 v2.0.0 $debug=AUSU.5-1-19041-1151
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+
+PackageIdentifier: NSIS.NSIS
+PackageVersion: "3.08"
+MinimumOSVersion: 4.0.1381.0
+InstallerType: nullsoft
+Scope: machine
+InstallModes:
+- interactive
+- silent
+FileExtensions:
+- nsi
+- nsh
+Installers:
+- Architecture: x86
+  InstallerUrl: https://sourceforge.net/projects/nsis/files/NSIS%203/3.08/nsis-3.08-setup.exe/download
+  InstallerSha256: BBC76BE36ECB2FC00D493C91BEFDAF71654226AD8A4FC4DC338458916BF224D0
+ManifestType: installer
+ManifestVersion: 1.0.0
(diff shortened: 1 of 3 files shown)


jedieaston (approved):

ItzLevvie:
>This is rather surprising and disappointing to me.

Singletons are how packages used to work before March 2021 so it was always limited to one installer.

Back then I think we had `Microsoft.Edge.x64` `Microsoft.Edge.arm64` and `Microsoft.Edge.x86` and now it's all in one package rather than separate packages thanks to multi-file manifests.

>Switches are allowed at the root of the manifest IIRC (assuming they are the same for all installers.

Yes, but when it comes to `/ALLUSERS` and `/CURRENTUSER`, it won't be able to work that way, only multi-file manifests will be able to achieve that.

> who is responsible for updating the other locales?

If it's contibuted by a contributor, then that person will likely be responsible for it because they've a better understanding of their own language compared to most of us.

Locales are also inherited from the whatever's written in the default locale (en-US), so only the fields that need to be localized i.e. tags and description are needed.

Usually having more locales brings a much better user experience because they can have tags and descriptions in their own language rather than having to understand English.

Additionally, but it's something that I haven't tried, locales could probably be used to better map the application to `winget list` (perhaps useful for EXEs) where installing the package gives you a different `PackageName` on a localized system compared to a English (US) system.

sredna:
> Yes, but when it comes to `/ALLUSERS` and `/CURRENTUSER`, it won't be able to work that way, only multi-file manifests will be able to achieve that.

I don't see why that would be the case. 

```yaml
InstallerType: nullsoft
InstallerSwitches:
   Silent: /S
Installers: 
 - Architecture: x64
   InstallerUrl: ...
   Scope: machine
 - Architecture: x64
   InstallerUrl: ...
   Scope: user
   InstallerSwitches:
      Custom: /CURRENTUSER
```

I should probably open a separate issue for this.

ItzLevvie:
> I don't see why that would be the case.

Are you sure? What happens when you run `winget validate` on that?

This is exactly what we're doing for multi-file manifests, but the sample you wrote seems to resemble a multi-file manifest, so I don't think that works here.

For singletons, I'm more thinking along the lines of:

```yaml
InstallerType: nullsoft
Installers: 
- Architecture: x64
  InstallerUrl: [...]
  Scope: machine
  Custom: /ALLUSERS
```

> I should probably open a separate issue for this.

This looks more like a feature request rather than an issue or a bug, and it's a change that should be done in the schema, but even if you wanted to open up an issue, I know singletons are just one file but the whole point of it is so only include **one** installer, it most likely won't be picked up on (as in there won't be any changes made to support multiple installers)

For singletons, you're not allowed to have multiple architectures: that means you're not allowed to have 2 x64 installers listed, or both an x64 and and x86 installers listed.

PR #29360, opened 2021-10-04
Windows Terminal Preview 1.11.2731.0

zadjii-msft:
I copy pasted the last one so this should be the same.

@miniksa fyi
-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/29360)


zadjii-msft:
I messed something up, Michael is fixing it.

PR #29363, opened 2021-10-04
Windows Terminal 1.10.2714.0

zadjii-msft:
Same deal as #29360, copy-pasted with a string replace, fyi @miniksa

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/29363)


zadjii-msft:
I messed something up, Michael is fixing it.

PR #29454, opened 2021-10-05
Fix: YamlCreate using invalid filename for downloading

SpecterShell:

-----
The script obtains the file name by slicing InstallerUrl variable and appends it to the destinaiton path, but sometimes the url doesn't contain a valid filename, which may cause errors.
For example, if InstallerUrl is set to "https://pinyin.thunisoft.com/webapi/v1/setup/downloadSetup?setupid=21f0e88683894ebcb440759c5aa25c47&filename=HuayuPY-V7.2.0.213.exe", the script will use "downloadSetup?setupid=21f0e88683894ebcb440759c5aa25c47&filename=HuayuPY-V7.2.0.213.exe" as file name, which contains an invalid file name character "?".
Many servers tell the client the real file name by adding a "ContentDisposition" field to the response header. If existed, the modified code will use it instead.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/29454)

### DIFF
--- a/Tools/YamlCreate.ps1
+++ b/Tools/YamlCreate.ps1
@@ -257,7 +257,7 @@ Function Read-WinGet-InstallerValues {
 
         try {
             $ContentDisposition = [System.Net.HttpWebRequest]::Create($InstallerUrl).GetResponse().Headers['Content-Disposition']
-            if ($null -ne $ContentDisposition -and $ContentDisposition -match 'filename="?(.*)"?;?') {
+            if ($null -ne $ContentDisposition -and $ContentDisposition -match 'filename="?([^";]+)') {
                 $FileName = $Matches[1]
                 $dest = "$env:TEMP\$FileName"
             }
--- a/Tools/YamlCreate.ps1
+++ b/Tools/YamlCreate.ps1
@@ -256,6 +256,11 @@ Function Read-WinGet-InstallerValues {
         $dest = "$env:TEMP\$FileName"
 
         try {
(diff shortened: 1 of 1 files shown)


Trenly:
I'm not familiar with what this code is actually doing? Would you be able to help me understand how it works?

SpecterShell:
> I'm not familiar with what this code is actually doing? Would you be able to help me understand how it works?

I updated the code so that it will not invoke web requests twice.
The script obtains the file name by slicing InstallerUrl variable and appends it to the destinaiton path, but sometimes the url doesn't contain a valid filename, which may cause errors.
For example, if InstallerUrl is set to "https://pinyin.thunisoft.com/webapi/v1/setup/downloadSetup?setupid=21f0e88683894ebcb440759c5aa25c47&filename=HuayuPY-V7.2.0.213.exe", the script will slice it into "downloadSetup?setupid=21f0e88683894ebcb440759c5aa25c47&filename=HuayuPY-V7.2.0.213.exe", which contains an invalid file name character "?".
Many servers tell the client the real file name by adding a "ContentDisposition" field to the response header. If existed, the modified code will use it instead.

jedieaston:
Why not just name the file `{PackageIdentifier}.{PackageVersion}.{extension}` like winget does? You'll have to write in what the extension is for each installer type, but that way you don't have to rely on a header.  (Or just name it "temp" or something, the extension doesn't really matter for this case I guess). 

I'm fine with this, but I'm just afraid if someone didn't configure their web server correctly you could still get a yucky name.

Trenly:
For the purposes of the script, do we even need the file extension other than for determining the installer type?

PR #29790, opened 2021-10-08
New version: TranslucentTB.TranslucentTB version 2021.5.0.0

vedantmgoyal9:
Auto-updated by [vedantmgoyal2009/winget-pkgs-automation](https://github.com/vedantmgoyal2009/winget-pkgs-automation)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/29790)

### DIFF
--- a/manifests/t/TranslucentTB/TranslucentTB/2021.5.0.0/TranslucentTB.TranslucentTB.installer.yaml
+++ b/manifests/t/TranslucentTB/TranslucentTB/2021.5.0.0/TranslucentTB.TranslucentTB.installer.yaml
@@ -13,7 +13,6 @@ Installers:
 - Architecture: x64
   InstallerType: msix
   InstallerUrl: https://github.com/TranslucentTB/TranslucentTB/releases/download/2021.5/bundle.msixbundle
-  Scope: machine
   InstallerSha256: DA0EF31E5D9E91218E54F93A068FBDD42E3868D1E331C3B32986422E105B8530
   SignatureSha256: 8ff8df14abca5154d9ef69dba98ca33caabdce97ccad4fa57f642072b8b56111
   PackageFamilyName: 28017CharlesMilette.TranslucentTB_v826wp6bftszj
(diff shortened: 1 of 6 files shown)


vedantmgoyal2009:
@KevinLaMS

ItzLevvie (changes requested):
This issue is caused by a dependency not available in the validation pipelines.

from the latest version of WinGet:
image

from an older version of WinGet:
image

ItzLevvie (commented):
Considering that TranslucentTB is available in the Microsoft Store and the `msstore` source may likely be the default source. I don't know if this is needed.

winget install 9PF4KZ2VN4W9 should install `TranslucentTB` from the Microsoft Store.

sylveon:
But `winget install TranslucentTB` should *not* install an old version. The version WinGet installs is old, broken on Windows 11, and not supported anymore.

Windows Terminal also depends on WinUI 2.6, and is updated regularly on WinGet, so what's up with that?

ItzLevvie:
> Worse even, while it does find a duplicate when the msstore source is enabled, it does not say which one is newer:

You should be able to find the version of the package by searching for it via `winget search` or by being specific with `winget show`:
image

Currently, the Microsoft Store API that WinGet depends on doesn't expose a version; so "Unknown" in this case is the latest version that you published to the Microsoft Store (Retail).

> Windows Terminal also depends on WinUI 2.6+, and is updated regularly on WinGet, so what's up with that?

Windows Terminal ships with a pre-built version of WinUI 2.7 which is outside of the installation process.

I should have been a bit more clearer but this only affects applications that have `Microsoft.UI.Xaml` listed as a dependency (`<Dependency></Dependency>`) in their `AppxManifest.xml` file.

> If the version WinGet distributes cannot be updated to something newer, then TranslucentTB should be removed entirely from WinGet.

PR was created for it: #44603

PR #30250, opened 2021-10-13

vedantmgoyal9:
null


jedieaston (dismissed):

kdpuvvadi:
Rufus is a portable application, it's not an installer. I had weird conversation with the author and he had different idea of how it should be done. i don't think it works without user interaction.

PR #30346, opened 2021-10-14
Update validation-pipeline.yaml

vedantmgoyal9:
https://github.com/microsoft/winget-pkgs/issues/30321#issuecomment-943521645

Resolves #30321 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/30346)

### DIFF
--- a/DevOpsPipelineDefinitions/validation-pipeline.yaml
+++ b/DevOpsPipelineDefinitions/validation-pipeline.yaml
@@ -4,6 +4,7 @@
 name: '$(Build.DefinitionName)-$(Build.DefinitionVersion)-$(System.PullRequest.PullRequestNumber)-$(Date:yyyyMMdd)-$(Rev:r)'
 
 trigger: none
+pr:
   branches:
    include:
      - master
--- a/DevOpsPipelineDefinitions/validation-pipeline.yaml
+++ b/DevOpsPipelineDefinitions/validation-pipeline.yaml
@@ -4,7 +4,11 @@
 name: '$(Build.DefinitionName)-$(Build.DefinitionVersion)-$(System.PullRequest.PullRequestNumber)-$(Date:yyyyMMdd)-$(Rev:r)'
 
 trigger: none
-pr: none
+pr:
+  branches:
+    include: master
+  paths:
+    include: manifests/
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


ghost (dismissed):

Trenly:
> Yes, I was thinking if it could at least **list** the pipelines.

https://dev.azure.com/ms/winget-pkgs/_build

Trenly:
> It can’t, since you aren’t a maintainer. ADO projects store pipelines in many repositories, being able to enumerate them even if you don’t have access to said repositories would be a security uh-oh. For this project, the only relevant ones are in this repo (and public in ADO), but that’s why you can’t use commands.
> 
> The reason I didn’t draft a PR was because I’m only 70% sure that that’s the problem (without being able to check the settings on the pipeline in ADO, I cannot be certain that’s what’s up). I was hoping the comment would get a maintainer to say “@jedieaston you dummy it was something completely different, here’s an explanation:”, or if it was the right answer, for them to do it since they’ll also have to go check the settings in ADO to make sure that they’re correct as well.



> It can’t, since you aren’t a maintainer. ADO projects store pipelines in many repositories, being able to enumerate them even if you don’t have access to said repositories would be a security uh-oh. For this project, the only relevant ones are in this repo (and public in ADO), but that’s why you can’t use commands.
> 
> The reason I didn’t draft a PR was because I’m only 70% sure that that’s the problem (without being able to check the settings on the pipeline in ADO, I cannot be certain that’s what’s up). I was hoping the comment would get a maintainer to say “@jedieaston you dummy it was something completely different, here’s an explanation:”, or if it was the right answer, for them to do it since they’ll also have to go check the settings in ADO to make sure that they’re correct as well.

I would also add that it appears that there is some form of automation which also updates the pipelines, so a Maintainer should also check to make sure that this wouldn't break that automation

ItzLevvie (commented):
cc @msftrubengu

msftrubengu (changes requested):

msftrubengu (approved):

msftrubengu:
The behavior is the same. Internally in the DevOps configuration we had an override for the pr triggers with the same triggers that you added. I don't know what happened, but this is broken. I'll follow up with the DevOps team internally to find out what went wrong.

Untitled

PR #30387, opened 2021-10-15
Create rerun_pipelines_mods.yml

### LINKED ISSUE #30364: [New Feature]: Moderators should be able to request another run of the tests.
### Description of the new feature/enhancement

I'd like to (and I'm sure the other mods would too) be able to request the pipelines to re-run if they fail, as sometimes PRs get stuck in an unhealthy state and we have to close and reopen the PR (or tell the contributor to do it), which clutters the history up as GitHub doesn't really expect for PRs to open and close very often. It's confusing when trying to figure out the current state of the PR to scroll through and see "@jedieaston closed this pull request. @jedieaston opened this pull request" eight times, with notifications being sent everywhere at the same time. And sometimes labels just disappear during the close/re-open even if they'd be useful to maintainers for debugging.

Sometimes it makes sense for a maintainer to take a look, but often some backend service timed out (*cough* defender *cough*) and if we run it again everything is fine. But @msftbot is extremely okay with just assigning it and leaving Chrome or Edge or other safe apps in limbo.

### Proposed technical implementation details (optional)

Two ways I can think of to implement this:

- the best for UX would probably be to have @wingetbot start listening for keywords/commands in our comments. That way we don't have to be delegated permissions in ADO to talk to the Azure DevOps app (which I understand you can't do), but since @wingetbot has the permissions, we could say "@wingetbot rerun" which would trigger it to do "/azp rerun". If @wingetbot can't listen for commands right now, that might be a problem.
- If that will be too hard to implement, we could just have a GitHub Action that looks for comments and triggers the pipelines. That might be easier to implement depending on maintainer preferences. You could just have a action that looks for [issue_comments](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#issue_comment) that are in pull requests, if it finds one that says "rerun" written by someone on the mods list, then it triggers the pipeline to run again (either via `/azp run` or [the GitHub Actions -> pipelines integration](https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/github-actions?view=azure-devops)).

vedantmgoyal9:
* Resolves #30364 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/30387)

### DIFF
--- a/.github/workflows/rerun_pipelines_mods.yml
+++ b/.github/workflows/rerun_pipelines_mods.yml
@@ -14,7 +14,7 @@ jobs:
         shell: pwsh
         run: |
           $moderators = @('jedieaston','ItzLevvie','OfficialEsco','quhxl','KaranKad','ImJoakim')
-          if ("${{ github.event.comment.user.login }}" -in @($moderators,'${{ github.event.issue.user.login }}'))
+          if ("${{ github.event.comment.user.login }}" -in $moderators)
           {
             Write-Output "IsRequestValid=yes" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
           }
@@ -42,5 +42,5 @@ jobs:
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              body: '<samp>This feature can only be used by PR authors and Moderators.</samp>'
+              body: '<samp>This feature can only be used by Moderators.</samp>'
             })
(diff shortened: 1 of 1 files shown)


jedieaston (changes requested):
I kind of wanted some discussion before we hopped into a solution here. I'm sure they're working on stuff internally with wingetbot and the pipelines and I don't want this to get in the way (maybe it can't, but I'd like a maintainer to sign off first). 

I had the Actions implementation in mind as a stopgap if command support for wingetbot required too many architectural changes to be worthwhile for this feature. Given that we don't know this, we should hold off on this for now IMO. 

(It looks like it works though, which is cool. Thanks for doing this!)

PR #31185, opened 2021-10-24
Update RetroArch to version 1.9.11

nopeinomicon:

-----

Update RetroArch to version 1.9.11 and start using multi-manifest structure to enable 32-bit installer

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/31185)


ItzLevvie (changes requested):
:(

image

ItzLevvie (changes requested):
> This also happens with the current RetroArch version in the repository, just as an extra note. This wasn't a recent breakage.

Thanks. A PR was submitted to remove this package from the repository: https://github.com/microsoft/winget-pkgs/pull/31658

Since all packages are validated against VMs in Azure Pipelines, it doesn't have any GPU attached and has no dependencies installed, and therefore fails.

This package existed months ago because that was when automatic validation in pipelines didn't exist and was validated on a actual hardware with possibly consumer versions of Intel or AMD CPUs with NVIDIA or AMD GPUs.

ItzLevvie (approved):
>Ah fair enough. I'm not sure of the mechanism but maybe this should be brought to the attention of upstream so they can fix that (considering I'm not entirely sure that blocking dialog has much use) or if it's possible they do provide a UWP version (although it requires a separate developer signature so I'm not sure if that would work here)

If it's open-source, you could probably check their nullsoft source and see if there's any switches to bypass it. Otherwise, you would likely need to contact them to see if it's something that could be added.

PR #31667, opened 2021-10-25
​

ItzLevvie:
​


jedieaston (approved):
This should probably be a step in the pipelines (or during Installation Validation, at least). It would be simple to check the `PackageFamilyName`, and then fail the pipeline if it is wrong (which it almost never should be).

ghost (approved):

PR #32219, opened 2021-10-26
Update Microsoft Git metadata

ldennington:
The Git Fundamentals team is in the process of switching to wingetcreate to
publish its tools to microsoft/winget-pkgs. During this effort, we noticed
fields in the latest manifest that need to be modified both for correctness
and compatibility with wingetcreate.

- [X] Does your manifest conform to the [1.0 schema](https://github.com/microsoft/winget-cli/blob/master/doc/ManifestSpecv1.0.md)? Note: This will become a multi-file manifest in the next release using wingetcreate.

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/32219)

### DIFF
--- a/manifests/m/Microsoft/Git/2.33.0.0.3/Microsoft.Git.yaml
+++ b/manifests/m/Microsoft/Git/2.33.0.0.3/Microsoft.Git.yaml
@@ -1,16 +1,13 @@
 PackageIdentifier: Microsoft.Git
 PackageVersion: 2.33.0.0.3
 PackageName: Microsoft Git
-Publisher: The Git Client Team at GitHub
+Publisher: The Git Fundamentals Team at GitHub
 Moniker: microsoft-git
 PackageUrl: https://aka.ms/ms-git
 Tags:
 - microsoft-git
 License: GPLv2
-ShortDescription: |
-  Git distribution to support monorepo scenarios.
-  Note: This is not Git for Windows. Unless you are working in a monorepo and require
-  specific Git modifications, please run `winget install git` to start using Git for Windows.
+ShortDescription: Git distribution to support monorepo scenarios.
 Installers:
 - Architecture: x64
   InstallerUrl: https://github.com/microsoft/git/releases/download/v2.33.0.vfs.0.3/Git-2.33.0.vfs.0.3-64-bit.exe


ItzLevvie (dismissed):

manifests/m/Microsoft/Git/2.33.0.0.3/Microsoft.Git.yaml
```diff
@@ -1,18 +1,15 @@
...
 - microsoft-git
 License: GPLv2
-ShortDescription: |
-  Git distribution to support monorepo scenarios.
-  Note: This is not Git for Windows. Unless you are working in a monorepo and require
-  specific Git modifications, please run `winget install git` to start using Git for Windows.
+ShortDescription: Git distribution to support monorepo scenarios.
 Installers:
-- Architecture: x64
+- Architecture: x86
```
ItzLevvie, line 12:
```suggestion
- Architecture: x64
```
Is this change accidental?

Setting `Architecture:` boots up an x86 VM to install a 64-bit app which doesn't work.

ItzLevvie (changes requested):

manifests/m/Microsoft/Git/2.33.0.0.3/Microsoft.Git.yaml
ldennington: This change was very much done on purpose. Although the application is `x64` the installer is `x86`. We are in the process of switching over to `wingetcreate`, which breaks when `x64` is specified here because it detects that the installer it downloads is `x86` but the manifest specifies `x64`. This feels a bit like a catch-22 - I will discuss with my team how to fix.

ldennington: Basically we have an `x86` installer with a 64-bit payload. We don't have a way to create an `x64` installer. Would there be any way to test this on an `x64` VM?

ItzLevvie: > Would there be any way to test this on an `x64` VM?

I don't think so as it's hardcoded in the Azure Pipelines, so when the `Architecture` is set to `x86`, it boots up an x86 VM, and if the `Architecture` is set to `x64`, it boots up an x64 VM.

I haven't used wingetcreate in a while, so I would personally file an issue in the issue tracker and see if there are ways to bypass the message.

ItzLevvie (changes requested):
https://github.com/microsoft/winget-pkgs/pull/32219#discussion_r741549734

manifests/m/Microsoft/Git/2.33.0.0.3/Microsoft.Git.yaml
ldennington: The `wingetcreate` team has made the necessary updates on their end to fix this, so I've changed the architecture back to `x64`. Thanks for your help @ItzLevvie!

ItzLevvie (approved):

PR #34087, opened 2021-11-05
Mozilla.Firefox-msix version 94.0.0.0

oxygen-dioxide:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/34087)

### DIFF
--- a/manifests/m/Mozilla/Firefox-msix/94.0.0.0/Mozilla.Firefox-msix.installer.yaml
+++ b/manifests/m/Mozilla/Firefox-msix/94.0.0.0/Mozilla.Firefox-msix.installer.yaml
@@ -1,0 +1,18 @@
+# Created using wingetcreate 0.3.0.3
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+
+PackageIdentifier: Mozilla.Firefox-msix
+PackageVersion: 94.0.0.0
+Installers:
+- Platform:
+  - Windows.Desktop
+  MinimumOSVersion: 10.0.17763.0
+  Architecture: x64
+  InstallerType: msix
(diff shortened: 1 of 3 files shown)


jedieaston (changes requested):
Why can't this just be part of the `Mozilla.Firefox` manifest?

oxygen-dioxide:
> Why can't this just be part of the `Mozilla.Firefox` manifest?

The version number format of this msix package is different from the `Mozilla.Firefox` manifest. For example, Mozilla.Firefox 94.0 is Mozilla.Firefox-msix 94.0.0.0

jedieaston:
That's true, but due to Firefox's versioning scheme it shouldn't matter in practice (they only use to the second decimal point for any release), it shouldn't matter (94.0.0.0 would upgrade to 94.0.1 in winget's logic).

OfficialEsco:
> 
> 
> That's true, but due to Firefox's versioning scheme it shouldn't matter in practice (they only use to the second decimal point for any release), it shouldn't matter (94.0.0.0 would upgrade to 94.0.1 in winget's logic).

WinGet also "ignore"/remove zero's, so if we flipped the numbers 94.0.0.0 would not show up as a upgrade for version 94.0 (Its just good practice to keep the 0's if the versions are the same)
This can also be corrected in Manifest Version 1.1 with ease since the msix dosn't break winget

oxygen-dioxide:
#34147

oxygen-dioxide:
> Why can't this just be part of the `Mozilla.Firefox` manifest?

Then how can the user choose which package to install between the exe package and the msix package?

ItzLevvie (changes requested):
Added in https://github.com/microsoft/winget-pkgs/pull/34147?

PR #34333, opened 2021-11-08
Open ARP Entries after Install

Trenly:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/34333)

### DIFF
--- a/Tools/SandboxTest.ps1
+++ b/Tools/SandboxTest.ps1
@@ -212,13 +212,12 @@ Write-Host @'
 --> Refreshing environment variables
 '@
 Update-EnvironmentVariables
-`$currentARP = Get-ARPTable
 
 Write-Host @'
 
 --> Comparing ARP Entries
 '@
-(Compare-Object `$currentARP `$originalARP -IncludeEqual -ExcludeDifferent).InputObject 
+(Compare-Object (Get-ARPTable) `$originalARP -Property DisplayName,DisplayVersion,Publisher,ProductCode)| Select-Object -Property * -ExcludeProperty SideIndicator | Format-Table
 
 "@
 }
--- a/Tools/SandboxTest.ps1
+++ b/Tools/SandboxTest.ps1
@@ -165,7 +165,13 @@ function Update-EnvironmentVariables {
   }
 }
 
-
+function Get-ARPTable {
+  $registry_paths = @('HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\*','HKLM:\Software\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\*', 'HKCU:\Software\Microsoft\Windows\CurrentVersion\Uninstall\*', 'HKCU:\Software\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\*')
+  return Get-ItemProperty $registry_paths -ErrorAction SilentlyContinue | 
+       Select-Object DisplayName, DisplayVersion, Publisher, @{N='ProductCode'; E={$_.PSChildName}} |
+       Where-Object {$null -ne $_.DisplayName } |
+       Where-Object {$null -ne $_.Publisher }
+}
 '@
 
 $bootstrapPs1Content += @"
@@ -192,7 +198,7 @@ Write-Host @'
 --> Configuring Winget
 '@
 winget settings --Enable LocalManifestFiles
-
+`$originalARP = Get-ARPTable
 Write-Host @'
 
 
@@ -206,12 +212,13 @@ Write-Host @'
 --> Refreshing environment variables
 '@
 Update-EnvironmentVariables
+`$currentARP = Get-ARPTable
 
 Write-Host @'
 
---> Opening ARP Entries
+--> Comparing ARP Entries
 '@
-appwiz.cpl
+(Compare-Object `$currentARP `$originalARP -IncludeEqual -ExcludeDifferent).InputObject 
 
 "@
 }
@@ -280,7 +287,7 @@ if (-Not [String]::IsNullOrWhiteSpace($Manifest)) {
   Write-Host @"
     - Installing the Manifest $manifestFileName
     - Refreshing environment variables
-    - Opening ARP Entries
+    - Comparing ARP Entries
 "@
 }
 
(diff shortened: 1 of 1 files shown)


Tools/SandboxTest.ps1
```diff
@@ -207,6 +207,11 @@ Write-Host @'
 '@
 Update-EnvironmentVariables
 
+Write-Host @'
+
+--> Opening ARP Entries
+'@
+appwiz.cpl
```
jedieaston, line 214:
```suggestion
winget list --accept-source-agreements
```
It exposes most of the same info at the command line, and if the ID doesn't appear as the user thinks it should they can investigate? I'm just not a big fan of opening windows I can't pipe to a file :) 

The other way to do this is via PowerShell so that you can JSONify and analyze the info outside the sandbox for _fully automated luxury ARP verification_ but it's not a one liner.

Trenly: Hmm. I guess that would work too. I think that having the full ARP verification would be helpful, I'll take a look at that

jedieaston: My idea for doing that was to get the new ARP entries, JSONify them, dump them to a text file which is accessible outside the sandbox, and then on the outside, `ConvertFrom-JSON` and see if you can match them to the `PackageName`, `Publisher`, and `PackageVersion` in the manifest (excluding MSIX, which you'd have to do via `Get-AppxPackage` 🤢).

This is way more complicated than your one liner, so I understand if you don't want to do that with this (i.e. I'll approve this and take a look at doing that sometime myself).

Trenly: My thought with this was more that it sucks to have to open the ARP table every time I use the sandbox. I think full validation would be great, but we may also want to be careful about adding too much automation. Part of the beauty of the sandbox script is that it allows for manual validation of the installation and the ARP entries, which is a key part of moderation IMO

jedieaston: Yeah, in my patched version of the script `winget list` was the first thing I added (never got around to upstreaming the change), totally understandable.

I think we need to write some automated client-side (i.e. not in the validation pipeline) ARP validation purely for peace-of-mind (sometimes you're going fast and think you remember how the version numbers for that package looked, which isn't a good habit but it happens with the momentum of the repo). I'll approve this then.

jedieaston (dismissed):

Tools/SandboxTest.ps1
Trenly: I think I got it showing the new entries now

Trenly: Ugh. That only works if its the only package that gets installed. Time to do more digging

Tools/SandboxTest.ps1
```diff
@@ -206,7 +212,13 @@ Write-Host @'
 --> Refreshing environment variables
 '@
 Update-EnvironmentVariables
+`$currentARP = Get-ARPTable
 
+Write-Host @'
+
+--> Comparing ARP Entries
+'@
+(Compare-Object `$currentARP `$originalARP -IncludeEqual -ExcludeDifferent).InputObject
```
jedieaston, line 221:
```suggestion
(Compare-Object `$currentARP `$originalARP).InputObject 
```

This way it only shows the new ones, the way it was in your last commit it would only show the ARP entries that were on the machine before you ran `winget install`.

Example:
```
PS C:\Users\WDAGUtilityAccount\Desktop\winget-pkgs> (Compare-Object $currentARP $originalARP -IncludeEqual -ExcludeDifferent).InputObject

DisplayName    DisplayVersion Publisher             ProductCode
-----------    -------------- ---------             -----------
Microsoft Edge 90.0.818.66    Microsoft Corporation Microsoft Edge


PS C:\Users\WDAGUtilityAccount\Desktop\winget-pkgs> (Compare-Object $currentARP $originalARP).InputObject

DisplayName                  DisplayVersion Publisher        ProductCode
-----------                  -------------- ---------        -----------
JetBrains ETW Service        212.16.26.0    JetBrains s.r.o  {DEA96022-4B1D-4D5F-B590-4625EAFFF810}
JetBrains dotTrace 2021.2.2  2021.2.2       JetBrains s.r.o. {2d5368a1-3297-5a09-87f5-79a5b4864a3c}
JetBrains dotMemory 2021.2.2 2021.2.2       JetBrains s.r.o. {41109cb0-ef67-5b7b-96d3-23797d822c32}
JetBrains dotPeek 2021.2.2   2021.2.2       JetBrains s.r.o. {546640b4-6e04-545d-94f9-f7ce6cff2993}
JetBrains dotCover 2021.2.2  2021.2.2       JetBrains s.r.o. {e0b1c9b0-9540-5e20-8a2f-3524e9a06db0}
JetBrains Rider 2021.2.2     2021.2.2       JetBrains s.r.o. {e2d2761c-f4ec-5bb3-aad5-521b644d1fdc}


PS C:\Users\WDAGUtilityAccount\Desktop\winget-pkgs>
```

jedieaston (changes requested):

Tools/SandboxTest.ps1
Trenly: I tried that, but it doesn't work for me on PS5. It only fetches one object

jedieaston: I'm... on PowerShell 5 too? Hmm. What package are you testing with?

Trenly: > I'm... on PowerShell 5 too? Hmm. What package are you testing with?

I tested with GOG.Galaxy, Google.Chrome, and Git.Git

Trenly: I think I got it this time

Trenly: image

jedieaston:
I was literally about to comment that solution.

Works well for my tests.

jedieaston (approved):

Trenly:
> I was literally about to comment that solution.
> 
> Works well for my tests.
> 
> Edit: Citation for time travelers: https://stackoverflow.com/questions/18259861/compare-object-not-working-if-i-dont-list-the-properties/18346380

Amazing how we both found the same stack overflow article 😄

jedieaston:
@denelon this is ready for merge if you're cool with it.

denelon (approved):

PR #34451, opened 2021-11-09

vedantmgoyal9:
null


jedieaston (changes requested):
image

It's trying to install drivers, like the other versions of Cura.

vedantmgoyal2009:
I'll add this package to the skip list.

ItzLevvie (approved):

PR #34869, opened 2021-11-14
The Silver Searcher release 2.2.5

JFLarvoire:

NOTE: This so-called manifest spec page is obviously obsolete. The examples do not even pass the validation step.

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/34869)


JFLarvoire:
While at it, I have a few questions:
- Why does my ag_setup.exe program trigger Defender warnings when run by winget, but not when run directly?
- This setup actually contains both the 32 and 64-bits versions of the ag.exe program. I tried declaring both the x86 and x64 installers in the yaml manifest, exactly as shown in the last example in the [YAML specification](https://github.com/microsoft/winget-cli/blob/master/doc/ManifestSpecv1.0.md), but this does not pass the winget validation. On the other hand, several YAML items in wide use are not defined in that specification. Is there another specification somewhere that is more up-to-date?

ItzLevvie:
> NOTE: This so-called manifest spec page is obviously obsolete. The examples do not even pass the validation step.

How did you create this manifest? Have you taken a look at wingetcreate or YamlCreate.ps1

> I wish I could install it with the command winget install ag instead of the longer winget install "The Silver Searcher". But the ag name matches another package with a name ending with ag. Yet the ag suffix in that name is an abbreviation for the German word for company, like the Co suffix commonly used in the USA, or Ltd in the UK. Is there a way to make sure that "ag" alone unambiguously selects only my package?

`Moniker` should do that for you, but at the moment, it seems like it's a bit messy due to the Microsoft Store source.

> Why does my ag_setup.exe program trigger Defender warnings when run by winget, but not when run directly? What should I do to avoid getting these warnings?

`winget install --manifest` is likely using Zone Level 3, so if your users want to install it, it shouldn't be an issue.

JFLarvoire:
Thanks for the explanations

> How did you create this manifest? Have you taken a look at wingetcreate or YamlCreate.ps1 which would avoid this issue?
I started from another manifest I randomly selected in the repository, then adapted it :-)
I'll try the other tools you mention.

> For singletons, I believe you can use neutral to define x86 and x64, or just add x86 and x64 in multi-file manifest.
I'll try that too. But, again, the YAML manifest 1.0 schema reference page that we're supposed to conform to definitely does not list this option. It really needs to be updated to list all options, and remove the examples that are now invalid.

ItzLevvie (approved):

PR #35764, opened 2021-11-22
Move .validation for Teams Exploration to the new path

vedantmgoyal9:
`Microsoft.TeamsExploration` has been moved to `Microsoft.Teams.Exploration`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/35764)


vedantmgoyal2009:
@denelon I would like to have a GitHub team that we can mention directly instead of mentioning specific engineers. I think you can make @microsoft/windows-package-manager-admins public or create a new team.

denelon:
@vedantmgoyal2009 It's in the works 😁. We have an internal "on-call" system that allows us to have a rotation. That will allow issues to get automatically assigned to the right engineer, and we can use that to make it easier to @mention the on call team.

palenshus (approved):

PR #36002, opened 2021-11-24
Adobe.AdobeDigitalEditions version 4.5.11

NathanBnm:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/36002)

### DIFF
--- a/manifests/a/Adobe/AdobeDigitalEditions/4.5.11/Adobe.AdobeDigitalEditions.installer.yaml
+++ b/manifests/a/Adobe/AdobeDigitalEditions/4.5.11/Adobe.AdobeDigitalEditions.installer.yaml
@@ -6,6 +6,8 @@ PackageVersion: 4.5.11
 InstallerSwitches:
   Silent: '/S'
   SilentWithProgress: '/S'
+InstallerSuccessCodes:
+  - 122
 Installers:
 - Architecture: x64
   InstallerType: nullsoft
--- a/manifests/a/Adobe/AdobeDigitalEditions/4.5.11/Adobe.AdobeDigitalEditions.installer.yaml
+++ b/manifests/a/Adobe/AdobeDigitalEditions/4.5.11/Adobe.AdobeDigitalEditions.installer.yaml
@@ -3,6 +3,9 @@
 
 PackageIdentifier: Adobe.AdobeDigitalEditions
 PackageVersion: 4.5.11
+InstallerSwitches:
+  Silent: '/S'
+  SilentWithProgress: '/S'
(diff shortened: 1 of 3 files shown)


ItzLevvie (changes requested):
I'm stuck here:
image

Are there any switches to suppress this prompt so it can install fully unattended without requiring any user interaction?

When I click the **No, Thank You** button, I get 1223 as exit code, which seems like a success code, so you can add that to `InstallerSuccessCodes`.
image

NathanBnm:
Curiously I don't have such a popup. I'll check the installer options but I will not be able to test the effect.

Regarding the exit code apparently it's a know issue (See https://community.adobe.com/t5/digital-editions-discussions/adobe-digital-editions-exit-code-1223-upon-successful-install/td-p/9560528) but the installations proceeds without any issues so I think it can be added to succes codes as a workaround (anyway Adobe will probably not update this software before a while.)

NathanBnm:
Just made the changes. Let me know if install is still not silent.

ItzLevvie (changes requested):
I'm still seeing the same issue:
image

You could try and play around with the switches, or contact Adobe to see if there's any way to suppress this prompt.

I'm using Windows Sandbox to test the PR.

NathanBnm:
@ItzLevvie after looking for a solution I found [this](https://community.adobe.com/t5/digital-editions-discussions/adobe-digital-editions-4-5-silent-install/m-p/7616109). Apparently this pop up is appearing if the installer detects that there is no installed antivirus software by checking a registry value.

The only way to remove the pop up is to fake the installer by adding a registry key before the installation but this could maybe have unwanted side effects.

Any idea?

ItzLevvie (changes requested):
> The only way to remove the pop up is to fake the installer by adding a registry key before the installation but this could maybe have unwanted side effects.

WinGet doesn't allow scripts to be used during install and there's no support for it in WinGet.

There used to be a Adobe Digital Editions package in this repository before, but I ran into issues with installing it, more specifically the NSS=0 switch didn't work on Windows Sandbox.

It was in the repository before because old manifests were tested on bare-metal hardware before automation came into effect, so now it uses Virtual Machines which has built-in Microsoft Defender and a few other AV software (which is possibly executed through an API - not VirusTotal though).

I removed it because you'll run into prompt issues on a clean install of Windows, so if you're wanting to install 30+ software from WinGet, the prompt blocks the entire install until user interaction is needed.

PR #36210, opened 2021-11-26

vedantmgoyal9:
null


ItzLevvie (changes requested):
This is CD build (50-99). 328**80**

Exploration usually has 00-49 at the end of the of the build number.

vedantmgoyal2009:
> This is CD build (50-99). 328**80**
> 
> Exploration usually has 00-49 at the end of the of the build number.

It seems that the issue is because I am using the same variables in both scripts.

ItzLevvie (changes requested):
https://github.com/microsoft/winget-pkgs/pull/36210#issuecomment-980215636

ItzLevvie (approved):

PR #36973, opened 2021-12-05
update rproject.r's url

eitsupi:
Replace the currently used URLs with URLs that can be used permanently since they are only available in the latest version (e.g. #35010).

And, sets up the URL that are automatically forwarded to mirror sites around the world, rather than Microsoft's CRAN mirror in U.S.A. c.f. https://cran.r-project.org/mirrors.html

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/36973)

### DIFF
--- a/manifests/r/RProject/R/4.1.2/RProject.R.installer.yaml
+++ b/manifests/r/RProject/R/4.1.2/RProject.R.installer.yaml
@@ -15,7 +15,7 @@ FileExtensions:
 - r
 Installers:
 - Architecture: x64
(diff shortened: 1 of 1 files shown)


ItzLevvie (approved):

ItzLevvie (changes requested):
Looks like the pipelines were canceled for this PR.

Please close and re-open the PR to re-run the pipelines.

PR #37076, opened 2021-12-07
Update Microsoft.OpenJDK.11

VAllens:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/37076)

### DIFF
--- a/manifests/m/Microsoft/OpenJDK/11/11.0.13.8/Microsoft.OpenJDK.11.installer.yaml
+++ b/manifests/m/Microsoft/OpenJDK/11/11.0.13.8/Microsoft.OpenJDK.11.installer.yaml
@@ -13,17 +13,237 @@ Installers:
   InstallerUrl: https://aka.ms/download-jdk/microsoft-jdk-11.0.13.8.1-windows-x64.msi
   InstallerSha256: A6651E68C945D880D753C08515BE123DA4BBFEC8FE9A684463D2C336268C7FEE
   Scope: machine
+  InstallerLocale: de-DE
+  ProductCode: '{3CCA84F4-964E-44E4-B8C1-6AF65AAC80CA}'
+  UpgradeBehavior: uninstallPrevious
+  InstallerType: msi
+  InstallerSwitches:
+    Custom: ADDLOCAL=FeatureMain,FeatureEnvironment,FeatureJarFileRunWith,FeatureJavaHome
+- Architecture: x64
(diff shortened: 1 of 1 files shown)


jedieaston (changes requested):
This isn't necessary. If your machine's locale isn't in the Installers array, winget will use the default locale instead. Since the JDK doesn't have different installers for different locales, this behavior is correct.

VAllens:
> This isn't necessary. If your machine's locale isn't in the Installers array, winget will use the default locale instead. Since the JDK doesn't have different installers for different locales, this behavior is correct.

Hi, @jedieaston

My computer locale is `zh-CN`, JDK only `en-US`, so winget update failed.
The `winget update` command does not support the `--locale` parameter option.

My winget settings:
```json
{
    "$schema": "https://aka.ms/winget-settings.schema.json",

    // For documentation on these settings, see: https://aka.ms/winget-settings
    // "source": {
    //    "autoUpdateIntervalInMinutes": 5
    // },

    "installBehavior": {
        "preferences": {
            "scope": "machine",
            "locale": [ "zh-CN", "en-US" ]
        }
    },

    "telemetry": {
        "disable": false
    }
}
```

VAllens:
> This isn't necessary. If your machine's locale isn't in the Installers array, winget will use the default locale instead. Since the JDK doesn't have different installers for different locales, this behavior is correct.

@jedieaston

When I run the `winget upgrade Microsoft.OpenJDK.11` command to update, 
winget prompts that there are no updates available.

But in fact, the `winget list` command tells me that there is an update available for `Microsoft.OpenJDK.11`.

After I analyzed the cause of the problem and added the `zh-CN` locale to the `Microsoft.OpenJDK.11.installer.yaml` file,
and executed the `winget upgrade Microsoft.OpenJDK.11 -m "D:\repo\github\winget-pkgs\manifests\m\Microsoft\OpenJDK\11\11.0.13.8"` command.
it updated successfully !!!

jedieaston:
I figured it out, turns out the manifest for Microsoft.OpenJDK.11 was explicitly saying the installer only applied to en-US. 

I've resolved it in #37163.

ItzLevvie (changes requested):
Should be solved by https://github.com/microsoft/winget-pkgs/pull/37076#issuecomment-988437795.

VAllens:
Okay, Let's close it.
Follow up to this PR: https://github.com/microsoft/winget-pkgs/pull/37163

PR #37208, opened 2021-12-08
New version: Microsoft.OneDrive.Insiders version 21.245.1128.0001

vedantmgoyal9:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/37208)


ItzLevvie (approved):

OfficialEsco:
@ryfu-msft would be fantastic if you could manually validate this one aaaand explain the reason why it fails

ryfu-msft:
> @ryfu-msft would be fantastic if you could manually validate this one aaaand explain the reason why it fails

Hey @OfficialEsco,

The reason why OneDrive always fails is because automatic updates from OneDrive
can actually interfere with the automation that takes place when performing a validation. OneDrive is just one of the few packages that we block from running to ensure that the automation flow proceeds without interruption. We are looking into ways to resolve this, but for now we will have to continue performing a manual validation for this package.

OfficialEsco:
Ooooo well that makes sense then, thanks, should we ping someone when there is a new PR for OneDrive?

ryfu-msft:
> Ooooo well that makes sense then, thanks, should we ping someone when there is a new PR for OneDrive?

Yes, you can ping me, as I often overlook these PRs :)

PR #38018, opened 2021-12-17
New version: Erlang.ErlangOTP version 24.2

OfficialEsco:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/38018)


OfficialEsco:
@Trenly this package is missing the Publisher registry field if you want to go digging :)

Trenly:
> @Trenly this package is missing the Publisher registry field if you want to go digging :)

It looks like there's more going on than just the publisher and version. It looks like the previous version entries aren't removed even after update because they use different registry entries for each version

Trenly:
I think I was able to fix both issues at the same time; If the PR I submitted gets rejected, I'll set it to just add the publisher and version and call it good

ItzLevvie (approved):

PR #38084, opened 2021-12-17
Fix: Allow packages with 'Backup' in the name

Trenly:
@denelon  This line in the gitignore file is causing issues when trying to update packages that have 'Backup' as part of their package id such as `AOMEI.Backupper` or `Google.BackupAndSync`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/38084)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -236,7 +236,7 @@ Generated_Code/
 # to a newer Visual Studio version. Backup files are not needed,
 # because we have git ;-)
 _UpgradeReport_Files/
-Backup*/
+# Backup*/
 UpgradeLog*.XML
 UpgradeLog*.htm
 ServiceFabricBackup/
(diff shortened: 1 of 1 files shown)


ItzLevvie (approved):

Trenly:
@denelon

denelon:
Do we have any concerns with other unintended consequences?

Trenly:
> Do we have any concerns with other unintended consequences?

I tried to think of some and couldn't. In theory, any commits made into the manifests folder should be either manifests, or .validation files and should be tracked by git regardless of what the name of the package is. This change wouldn't effect anything outside of the manifests folder, so even if there were an edge case of someone having a Backup folder outside of /manifests, that would still be excluded

denelon (approved):

PR #38325, opened 2021-12-20
New Application Tiger VNC

alexwilcox9:

-----

Hi there,

First time contributing here so thank you in advance for you patience! 
I've had a go at adding a new manifest using the `wingetcreate` tool and then modified the produced yaml file to add another architecture and some silent install switches

Looking forward to your feedback,
Alex

### DIFF
--- a/manifests/t/TigerVNCproject/TigerVNC/1.12.0/TigerVNCproject.TigerVNC.installer.yaml
+++ b/manifests/t/TigerVNCproject/TigerVNC/1.12.0/TigerVNCproject.TigerVNC.installer.yaml
@@ -9,13 +9,13 @@ Installers:
   InstallerUrl: https://sourceforge.net/projects/tigervnc/files/stable/1.12.0/tigervnc-1.12.0.exe/download
   InstallerSha256: 68555159BC7F2D10F5D24112F43D6533B8F86B0F687183C1B180AA5C5096D6C6
   InstallerSwitches:
-    Custom: /VERYSILENT /NORESTART
+    Custom: /NORESTART
 - Architecture: x64
   InstallerType: inno
   InstallerUrl: https://sourceforge.net/projects/tigervnc/files/stable/1.12.0/tigervnc64-1.12.0.exe/download
   InstallerSha256: 027B45B2179F16BF95B9B125FC6F3F6B4070D402788117EC7D3CE70AE495C0F8
   InstallerSwitches:
-    Custom: /VERYSILENT /NORESTART
+    Custom: /NORESTART
 ManifestType: installer
 ManifestVersion: 1.0.0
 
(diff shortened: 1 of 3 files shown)


manifests/t/TigerVNCproject/TigerVNC/1.12.0/TigerVNCproject.TigerVNC.installer.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+PackageIdentifier: TigerVNCproject.TigerVNC
+PackageVersion: 1.12.0
+Installers:
+- Architecture: x86
+  InstallerType: inno
+  InstallerUrl: https://sourceforge.net/projects/tigervnc/files/stable/1.12.0/tigervnc-1.12.0.exe/download
+  InstallerSha256: 68555159BC7F2D10F5D24112F43D6533B8F86B0F687183C1B180AA5C5096D6C6
+  InstallerSwitches:
+    Custom: /VERYSILENT /NORESTART
```
OfficialEsco, line 12:
```suggestion
    Custom: /NORESTART
```

manifests/t/TigerVNCproject/TigerVNC/1.12.0/TigerVNCproject.TigerVNC.installer.yaml
```diff
@@ -0,0 +1,21 @@
...
+- Architecture: x64
+  InstallerType: inno
+  InstallerUrl: https://sourceforge.net/projects/tigervnc/files/stable/1.12.0/tigervnc64-1.12.0.exe/download
+  InstallerSha256: 027B45B2179F16BF95B9B125FC6F3F6B4070D402788117EC7D3CE70AE495C0F8
+  InstallerSwitches:
+    Custom: /VERYSILENT /NORESTART
```
OfficialEsco, line 18:
```suggestion
    Custom: /NORESTART
```

OfficialEsco (changes requested):
I swear i've seen this package before, but i guess not 🤷‍♂️

So first off all, the `Custom` field is not supposed to have any silent switches, there are `Silent` and `SilentWithProgress` for that. However when the `InstallerType` is defined as anything else than `exe` the `Silent`, `SilentWithProgress`, `Log` and `InstallLocation` switch is already defined on a global standard and should only be overwritten if the package don't like to install as `SilentWithInstall`.

If you want the Description to look better you can do this:
```
Description: |
  Line 1
  Line 2
  Etc
```

And for the `Validation-Unapproved-URL` label we have to wait for a MS Employee to fix tomorrow

alexwilcox9:
Hi @OfficialEsco thanks for the comments

I've made the changes suggested and everything still seems to install fine but we do get the installer UI pop up with a progress bar which I assume is acceptable.

Regarding the installer type being set to `inno`, should it be set to `exe`? That part was automatically created by `wingetcreate` so I have naively assumed it knows best.

OfficialEsco:
> I've made the changes suggested and everything still seems to install fine but we do get the installer UI pop up with a progress bar which I assume is acceptable.

This is because it uses the /SILENT switch, if you don't want the progress bar you have to use `winget install <PackageName> -h` or `--silent` (on your side)

> Regarding the installer type being set to inno, should it be set to exe? That part was automatically created by wingetcreate so I have naively assumed it knows best.

Its always best if the InstallerType is set to the correct type, wingetcreate detects the correct InstallerType 99.9% of the time.

OfficialEsco (approved):

PR #38499, opened 2021-12-21
WatchGuardTechnologies,Inc.WatchGuardSystemManager12.7.2 version 12.7.2

Crazy-Clown:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/38499)

### DIFF
--- a/manifests/w/WatchGuardTechnologies,Inc/WatchGuardSystemManager12/7/2/12.7.2/WatchGuardTechnologies,Inc.WatchGuardSystemManager12.7.2.installer.yaml
+++ b/manifests/w/WatchGuardTechnologies,Inc/WatchGuardSystemManager12/7/2/12.7.2/WatchGuardTechnologies,Inc.WatchGuardSystemManager12.7.2.installer.yaml
@@ -1,7 +1,7 @@
 # Created using wingetcreate 0.4.4.1
 # yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
 
-PackageIdentifier: WatchGuardTechnologies,Inc.WatchGuardSystemManager12.7.2
+PackageIdentifier: WatchGuardTechnologies.WatchGuardSystemManager
 PackageVersion: 12.7.2
 Installers:
 - Architecture: x86
--- a/manifests/w/WatchGuardTechnologies,Inc/WatchGuardSystemManager12/7/2/12.7.2/WatchGuardTechnologies,Inc.WatchGuardSystemManager12.7.2.installer.yaml
+++ b/manifests/w/WatchGuardTechnologies,Inc/WatchGuardSystemManager12/7/2/12.7.2/WatchGuardTechnologies,Inc.WatchGuardSystemManager12.7.2.installer.yaml
@@ -1,0 +1,13 @@
+# Created using wingetcreate 0.4.4.1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+
+PackageIdentifier: WatchGuardTechnologies,Inc.WatchGuardSystemManager12.7.2
+PackageVersion: 12.7.2
+Installers:
+- Architecture: x86
+  InstallerType: inno
+  InstallerUrl: https://cdn.watchguard.com/SoftwareCenter/Files/WSM/12_7_2/WSM12_7_2.exe
+  InstallerSha256: 31C7B219EB157FA102D985E0E4CAF3EC0604AF2E0652F324D8006F2860E0133F
+ManifestType: installer
+ManifestVersion: 1.0.0
+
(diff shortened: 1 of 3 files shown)


Trenly (changes requested):
Looks like this should pass validation, great job!

However, I'm marking as Needs-Author-Feedback just so the mods see this.
Both of us submitted PR's at the same time, and it looks like #38500 has a bit more metadata added to the locale file.

Edit: Whoops, found something that should be updated.

manifests/w/WatchGuardTechnologies,Inc/WatchGuardSystemManager12/7/2/12.7.2/WatchGuardTechnologies,Inc.WatchGuardSystemManager12.7.2.installer.yaml
```diff
@@ -0,0 +1,13 @@
+# Created using wingetcreate 0.4.4.1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+
+PackageIdentifier: WatchGuardTechnologies,Inc.WatchGuardSystemManager12.7.2
```
Trenly, line 4:
```suggestion
PackageIdentifier: WatchGuardTechnologies.WatchGuardSystemManager
```

manifests/w/WatchGuardTechnologies,Inc/WatchGuardSystemManager12/7/2/12.7.2/WatchGuardTechnologies,Inc.WatchGuardSystemManager12.7.2.locale.en-US.yaml
```diff
@@ -0,0 +1,13 @@
+# Created using wingetcreate 0.4.4.1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.0.0.schema.json
+
+PackageIdentifier: WatchGuardTechnologies,Inc.WatchGuardSystemManager12.7.2
```
Trenly, line 4:
```suggestion
PackageIdentifier: WatchGuardTechnologies.WatchGuardSystemManager
```

manifests/w/WatchGuardTechnologies,Inc/WatchGuardSystemManager12/7/2/12.7.2/WatchGuardTechnologies,Inc.WatchGuardSystemManager12.7.2.yaml
```diff
@@ -0,0 +1,9 @@
+# Created using wingetcreate 0.4.4.1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.version.1.0.0.schema.json
+
+PackageIdentifier: WatchGuardTechnologies,Inc.WatchGuardSystemManager12.7.2
```
Trenly, line 4:
```suggestion
PackageIdentifier: WatchGuardTechnologies.WatchGuardSystemManager
```

Trenly (changes requested):
Since there is already an older version of this package, it should use the same PackageIdentifier as the previous version
image

When changing the PackageIdentifier, you'll also have to change the name of the folder to match

Crazy-Clown:
> Since there is already an older version of this package, it should use the same PackageIdentifier as the previous version image
> 
> When changing the PackageIdentifier, you'll also have to change the name of the folder to match

Yea, for some reason within the existing package, the version field is blank for some reason.
unknown

Crazy-Clown:
@Trenly Are you able to assist with the latest comment from @msftbot ? I don't understand what metadata was not compliant.

Crazy-Clown:
> It is always blank, that is the fault of the app creator. They are not setting the version in Control Panel, and so winget can't determine what version the app is because the publisher does not set the DisplayVersion registry key. The only way to really fix this is to contact the publisher and request that they add the version property in their build scripts

Oh so this value has to be implemented?
version

Trenly:
> Oh so this value has to be implemented? version
> 
> I will approach WatchGuard and ask them to implement this. Thanks for the info on that.

Yes, that is exactly the value that would need to be added. I can provide a command that should fix it on your computer as a workaround, but the best way to get it resolved is for publishers to properly set their app information

Crazy-Clown:
> but the best way to get it resolved is for publishers to properly set their app information

I've put in a request for enhancement with the vendor, will hopefully see a change in the future. Thanks.

Trenly:
> @Trenly I've put in a request for enhancement with the vendor, will hopefully see a change in the future. Thanks. Does this info mean the app will not be published unless it's resolved?

No, it is published now; All it means is that you'll still see the issue where you're stuck in the upgrade loop of unknown version

PR #40072, opened 2022-01-06
Remove Old: JanDeDobbeleer.OhMyPosh version 4.0.0

vedantmgoyal9:
null


ItzLevvie (changes requested):
Looks like PowerToys moved to a new installer (so switches are going to be different):
image

Since the UI is similar to VC++, it's probably Burn or Wix.

ItzLevvie (approved):

PR #40115, opened 2022-01-07
Chore: Update Moniker for better search experience

Trenly:
* Allows app to be found using winget search eadesktop
* Allows app to be found using winget search eaapp

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/40115)


Trenly:
I have no idea how this passed validation in https://github.com/microsoft/winget-pkgs/pull/37326
It doesn't appear to have changed at all. If this can't pass validation on a simple metadata change, It should probably be removed?

OfficialEsco:
EA Desktop/App is funny, it SOMETIMES prompts for EULA, which i accepted with ACCEPT=YES until it didn't work anymore, we then switched to the Burn InstallerType which solved it for a while, however now we're back to this
WindowsSandboxClient_mrN93emI40

Trenly:
> EA Desktop/App is funny, it SOMETIMES prompts for EULA, which i accepted with ACCEPT=YES until it didn't work anymore, we then switched to the Burn InstallerType which solved it for a while, however now we're back to this !

Looks like they just released a new version too, since the hash changed. I updated it to be the new version, and your switch seemed to work. I threw it in the custom switches so that it shouldn't effect the burn silent behavior in case it somehow switches back

OfficialEsco:
Whut, i tested my switch when typing that command and it did nothing again.. as i said, weird package...
I also didn't get the new version tho, seems to have released in the last 5 hours, could you also change the URL to https://origin-a.akamaihd.net/EA-Desktop-Client-Download/installer-releases/EAappInstaller.exe? (If that also have the new version)

ItzLevvie (approved):
image

PR #40265, opened 2022-01-08
New version: Microsoft.WindowsWDK version 10.1.22000.1

OfficialEsco:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/40265)

### DIFF
--- a/manifests/m/Microsoft/WindowsWDK/10.1.22000.1/Microsoft.WindowsWDK.installer.yaml
+++ b/manifests/m/Microsoft/WindowsWDK/10.1.22000.1/Microsoft.WindowsWDK.installer.yaml
@@ -9,7 +9,7 @@ InstallerSwitches:
   Silent: /q
   SilentWithProgress: /q
 Installers:
-- Architecture: x86
+- Architecture: x64
   InstallerUrl: https://download.microsoft.com/download/7/d/6/7d602355-8ae9-414c-ae36-109ece2aade6/wdk/wdksetup.exe
   InstallerSha256: 6538402F1D71836D5B9C2B9A705A2599F2FC2D03A9616B33BCA3D9D9768CA374
 ManifestType: installer
--- a/manifests/m/Microsoft/WindowsWDK/10.1.22000.1/Microsoft.WindowsWDK.installer.yaml
+++ b/manifests/m/Microsoft/WindowsWDK/10.1.22000.1/Microsoft.WindowsWDK.installer.yaml
@@ -1,0 +1,16 @@
+# Created with YamlCreate.ps1 v2.0.4 $debug=AUSU.7-2-1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.1.0.schema.json
+
+PackageIdentifier: Microsoft.WindowsWDK
+PackageVersion: 10.1.22000.1
+MinimumOSVersion: 10.0.0.0
+InstallerType: burn
+InstallerSwitches:
+  Silent: /q
+  SilentWithProgress: /q
+Installers:
+- Architecture: x86
+  InstallerUrl: https://download.microsoft.com/download/7/d/6/7d602355-8ae9-414c-ae36-109ece2aade6/wdk/wdksetup.exe
+  InstallerSha256: 6538402F1D71836D5B9C2B9A705A2599F2FC2D03A9616B33BCA3D9D9768CA374
+ManifestType: installer
+ManifestVersion: 1.1.0
--- a/manifests/m/Microsoft/WindowsWDK/10.1.22000.1/Microsoft.WindowsWDK.installer.yaml
+++ b/manifests/m/Microsoft/WindowsWDK/10.1.22000.1/Microsoft.WindowsWDK.installer.yaml
@@ -9,7 +9,7 @@ InstallerSwitches:
   Silent: /q
   SilentWithProgress: /q
 Installers:
-- Architecture: x64
+- Architecture: x86
   InstallerUrl: https://download.microsoft.com/download/7/d/6/7d602355-8ae9-414c-ae36-109ece2aade6/wdk/wdksetup.exe
   InstallerSha256: 6538402F1D71836D5B9C2B9A705A2599F2FC2D03A9616B33BCA3D9D9768CA374
 ManifestType: installer
--- a/manifests/m/Microsoft/WindowsWDK/10.1.22000.1/Microsoft.WindowsWDK.installer.yaml
+++ b/manifests/m/Microsoft/WindowsWDK/10.1.22000.1/Microsoft.WindowsWDK.installer.yaml
@@ -8,6 +8,7 @@ InstallerType: burn
 InstallerSwitches:
   Silent: /q
   SilentWithProgress: /q
+  Log: ' ' # Do not use, path is too long.
 Installers:
 - Architecture: x86
   InstallerUrl: https://download.microsoft.com/download/7/d/6/7d602355-8ae9-414c-ae36-109ece2aade6/wdk/wdksetup.exe
--- a/manifests/m/Microsoft/WindowsWDK/10.1.22000.1/Microsoft.WindowsWDK.locale.en-US.yaml
+++ b/manifests/m/Microsoft/WindowsWDK/10.1.22000.1/Microsoft.WindowsWDK.locale.en-US.yaml
@@ -1,0 +1,26 @@
+# Created with YamlCreate.ps1 v2.0.4 $debug=AUSU.7-2-1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.1.0.schema.json
+
+PackageIdentifier: Microsoft.WindowsWDK
+PackageVersion: 10.1.22000.1
+PackageLocale: en-US
+Publisher: Microsoft Corporation
+PublisherUrl: https://www.microsoft.com/en-us/
+PublisherSupportUrl: https://support.microsoft.com/en-us
+PrivacyUrl: https://privacy.microsoft.com/en-us/privacystatement
+Author: Microsoft Corporation
+PackageName: Windows Driver Kit
+PackageUrl: https://docs.microsoft.com/en-us/windows-hardware/drivers/download-the-wdk
+License: Proprietary
+# LicenseUrl: 
+Copyright: Copyright (c) Microsoft Corporation
+# CopyrightUrl: 
+ShortDescription: The WDK is used to develop, test, and deploy Windows drivers.
+# Description: 
+# Moniker: 
+# Tags: 
+# Agreements: 
+# ReleaseNotes: 
+# ReleaseNotesUrl: 
+ManifestType: defaultLocale
+ManifestVersion: 1.1.0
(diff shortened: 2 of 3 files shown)


OfficialEsco:
```
Enter a pull request number: #40265

Found Windows Driver Kit [Microsoft.WindowsWDK] Version 10.1.22000.1
Downloading https://download.microsoft.com/download/7/d/6/7d602355-8ae9-414c-ae36-109ece2aade6/wdk/wdksetup.exe
  ██████████████████████████████  1.29 MB / 1.29 MB
Starting package install...
Successfully installed

Name        : Windows Driver Kit - Windows 10.0.22000.1
Publisher   : Microsoft Corporation
Version     : 10.1.22000.1
ProductCode : {bf6c899f-b2a8-46b2-83dd-b58b48ecc2ba}
Uninstall   : winget uninstall "{bf6c899f-b2a8-46b2-83dd-b58b48ecc2ba}"
```

OfficialEsco:
@ItzLevvie any idea?

msftrubengu:
I looked at the logs. x64 is failing with 0x8A150006

OfficialEsco:
This also seems to be a client side issue https://github.com/microsoft/winget-cli/issues/500

ItzLevvie (dismissed):
image

This is from a Microsoft Azure VM which is a clean installation of Windows 10 based on version 21H2.

denelon:
I'm going to trigger the build again so we can take a look at the output artifacts to see if we can determine what is happening.

/azp run

denelon:
@OfficialEsco I just met with the WindowsWDK team. It appears we have an issue when passing the silent install to the package. There are some custom actions looking for some of the "run time" dependencies. In order for the Windows WDK to function properly, the Windows SDK and a supported Version of Visual Studio also need to be installed. As they are not "install time" dependencies, I think we're going to be able to allow these packages to continue to be included in the Community Repository. As we work towards dependency support, we will want to enrich the manifest to include those dependencies.

kjlk2j3sd:
Hi.  This is great.  A few issues.

1. The package is named 10.1.22000.1.  This seems to be a bug.  It should be named 10.0.22000.1 .   And it should definitely be 100% aligned with the SDK.  See for example here: https://github.com/microsoft/winget-pkgs/commit/6cddea4ea14849ca13b434ac9ed8c302a297a912 .

Note there is a similar bug for https://github.com/microsoft/winget-pkgs/blob/5ce9493bcca5653de3a954c638bc70dcdc2448c9/manifests/m/Microsoft/WindowsWDK/10.1.19041.685/Microsoft.WindowsWDK.yaml .  It should have been 10.0.19051.685 .

Why three yaml files rather than one?  The WDK is not localized.   I may be missing pieces of how Winget packages are to work.

The wdk.vsix does not get installed.  That is perhaps the right solution for now.  But before you can use an installed WDK you will definitely have to run the VSIX installer on "C:\Program Files (x86)\Windows Kits\10\Vsix\VS2019\WDK.vsix"

The WDK has dependencies on matching SDK as well as an appropriate installation of VS.  I guess the user would have to manually discover and provide those dependencies.

Thank you.

Jakob

kjlk2j3sd:
That's not really a reason to have the files have misleading names.  See Kevin's PR for the SDK that I linked above.  I still believe the right file name and version number would be 10.0.*.  What am I missing?

ItzLevvie:
> That's not really a reason to have the files have misleading names. See Kevin's PR for the SDK that I linked above. I still believe the right file name and version number would be 10.0.*. What am I missing?

The pull request uses `10.1` here because WinGet heavily relies on the information provided from the Control Panel's Add or Remove Programs. The `10.1` version is a bug in the SDK itself

If the version in the package does not match the version shown in Control Panel, it creates upgrade loops where it'll download 1 or 2 GB an unlimited amount of times.

For this package, the pull request author can choose to use either `10.1` or `10.0` for `PackageVersion` as the upgrade loops only kick in if the Control Panel version is less than this package's version.

Some of the development kit packages got reverted back to `10.0` because the team behind that reached out to the WinGet team - https://github.com/microsoft/winget-pkgs/pull/29912#issuecomment-941155303:
image

Once WinGet 1.3 is out of preview, there can be fixes implemented in the package to display it as `10.0.22000.1` to the user or the PR author can change the package version to `10.0.22000.1`.

> Why three yaml files rather than one? The WDK is not localized. I may be missing pieces of how Winget packages are to work.

For the WDK, localization files can be used for descriptions and tags.

i.e. A non-native English person may not understand what Microsoft.WindowsSDK is if they looked at the description and using Google Translate or Bing Translate for translation may not provide accurate results.

> The WDK has dependencies on matching SDK as well as an appropriate installation of VS. I guess the user would have to manually discover and provide those dependencies.

Once dependency support is out of experimental stages, it can be included in the package.

ItzLevvie:
> The ADK, SDK and WDK are not localized.

ADK, SDK & WDK aren't localized but there are **localized documentations** of them which are useful to be used in WinGet ~~which is what I'm talking about~~ ;)

For example - here's a Chinese page for Windows SDK: https://developer.microsoft.com/zh-CN/windows/downloads/windows-sdk/. It's useful to implement localized strings for descriptions, tags, the package's URL, etc.

> I do not claim to understand the finer versioning or file naming details.

It's documented over at https://github.com/microsoft/winget-pkgs/blob/master/doc/manifest/schema/1.0.0/installer.md#packageversion.

> I am simply saying SDK, ADK and WDK no matter what should be symmetrically named and versioned.

I do agree with consistency for Microsoft's development kits but we are following the documentation listed at https://github.com/microsoft/winget-pkgs/blob/master/doc/manifest/schema/1.0.0.

If we don't then many people will complain about `winget upgrade --all` or `winget upgrade XXXXX.XXXXX` installing the same application more than 50 times.

For other packages outside of Microsoft's development kits, this issue will be resolved in the future WinGet versions.
For this package, the pull request author can choose to use `10.0.XXXXX.X` instead.

> there is no reason for SDK to use one yaml style and WDK to use another.

It's a leftover from the schema transition to schema 1.0.0 from schema 0.1.0 and shouldn't cause any issues.

It can be easily converted to multi-file manifest format at any time by anyone via the `wingetcreate` tool or the `YamlCreate.ps1` script.

kjlk2j3sd:
Good nedws.  I was able to validate the WDK package.  Here you go:

winget install Microsoft.VisualStudio.2019.Community
REM Now add a ton of extra VS components…
winget install Microsoft.WindowsSDK -v 10.0.22000.194
winget install -m C:\packages\WindowsWDK\10.1.22000.1
"C:\Program Files (x86)\Windows Kits\10\Vsix\VS2019\WDK.vsix"

Now you can instantiate a driver template and build with VS 😊

lechacon:
@OfficialEsco Out of curiosity, how do you know that the installer has type Burn and not just Exe?

kjlk2j3sd:
Howdy. Jakob here.  Let me ask that question differently. Why the change since prior WDK package. And also why change of architecture.  See screenshot.

[10b55e55-b347-482a-8f5b-be89f7f740f4]

On May 5, 2022, at 17:30, Chacón ***@***.***> wrote:

﻿

@OfficialEsco<https://nam12.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FOfficialEsco&data=05%7C01%7C%7C70dcd233a7b34bfb839508da2ef78ef6%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637873938031746796%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=Bv6WpKKbFvAhs6j09GVcrCIy0%2F5sW8AAQ7qO%2FFpflI4%3D&reserved=0> Out of curiosity, how do you know that the installer has type Burn and not just Exe?

lechacon:
And, while we are talking about the installer type...

I believe it being listed as Burn instead of Exe is the cause for the installation errors. Since it is Burn, we know what the parameter for log location is and set it automatically during installation. This defaults to the winget logs folder plus a pretty long file name, and then the WDK installer itself appends something pretty long to that, which ends up exceeding the max path length.

So @OfficialEsco would you mind changing the installer type to Exe to see if that passes validation? We may also want to open an issue in winget-cli about long log paths.

Also, @kjlk2j3sd says that this version of WDK should be compatible with x86, not only x64.

lechacon:
Ah, there's already an issue about the log path length: microsoft/winget-cli#2036
It's the same thing as here: The installer adding something to the log path we provide

manifests/m/Microsoft/WindowsWDK/10.1.22000.1/Microsoft.WindowsWDK.installer.yaml
```diff
@@ -0,0 +1,16 @@
+# Created with YamlCreate.ps1 v2.0.4 $debug=AUSU.7-2-1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.1.0.schema.json
+
+PackageIdentifier: Microsoft.WindowsWDK
+PackageVersion: 10.1.22000.1
+MinimumOSVersion: 10.0.0.0
+InstallerType: burn
```
lechacon, line 7:
```suggestion
InstallerType: exe
```

manifests/m/Microsoft/WindowsWDK/10.1.22000.1/Microsoft.WindowsWDK.installer.yaml
```diff
@@ -0,0 +1,16 @@
...
+InstallerSwitches:
+  Silent: /q
+  SilentWithProgress: /q
+Installers:
+- Architecture: x64
```
lechacon, line 12:
```suggestion
- Architecture: x86
```

OfficialEsco:
So.. you want me to change the InstallerType to exe to avoid using the log switch? 🤔 i can add the log switch back and break it, currently the only thing the Burn InstallerType is doing is adding the log path since the Silent Switches are defined.

Changing the InstallerType to exe will also break WinGet Upgrade because it will detect it as a different installation type.

The Manifest works locally, which means this is a pipeline issue because the logs are named `InstallationVerification-x64-Scope_Undefined-Locale_Undefined-WinGet-Microsoft.WindowsWDK.10.1.22000.1-2022-05-04-08-51-13.394.log`

kjlk2j3sd:
Hi.

I don’t think “winget upgrade” makes a lot of sense for Windows Kits.  It is normal to install a bunch of those kits “side by side”.  For example a user installing 22000 would often not want to uninstall prior installed 20348.

Yes I was also able to use the manifest manually.

I am quite new to winget package management, but my bigger point is that:

  *   The various Windows kits, say: ADK, SDK, WDK
  *   The various builds of said kits: 17763, 18362, 19041, 20348, 22000, 22xxx (coming real soon 😊).
All as a starting point should have identical packages except of course different name and different location.  But… As identical as possible.  Perhaps even autogenerated from a template.

Again thanks for putting these packages together.

Best.

Jakob

OfficialEsco:
My reply was to @lechacon
If this is a side-by-side installation there is bigger winget upgrade issues at hand and we need to split them into different folders.

lechacon:
> So.. you want me to change the InstallerType to exe to avoid using the log switch? 🤔 i can add the log switch back and break it, currently the only thing the Burn InstallerType is doing is adding the log path since the Silent Switches are defined.

Yes, unless you have a better idea. I'm suggesting changing to exe and not adding a log switch. I tested it locally and it didn't work for me, so maybe it depends on the username length, which would cause random errors. IIRC, the path was 262 characters long, so just a couple of chars of difference in the username would . If it is exe and there is no log switch, then it would work by default, and if somebody adds it as

ItzLevvie (approved):

ghost (approved):

PR #40444, opened 2022-01-11
Microsoft.WindowsSubsystemForLinuxKernel version 5.10.16

felipecrs:

This greatly simplifies the [manual installation/configuration of WSL2](https://docs.microsoft.com/en-us/windows/wsl/install-manual) in Windows 10 versions older than 2004, which is the case for the company that I work as we use the Enterprise edition which is still supported. Also, it helps in the upgrade in case the company has disabled the Windows Updates for Microsoft features. Otherwise, users would have to check for updates and download the installer by themselves.

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/40444)


felipecrs:
This package requires some very specific requirements. For example, if it's installed in a too new system, it will exit by saying that a newer version is already installed. If the environment does not have the required Windows features enabled such as WSL and Virtual Machine Platform. Not to mention it requires nested virtualization capabilities in the (virtual) machine.

@denelon any tip on the way forward?

OfficialEsco:
Can you convert this into draft since its unfinished and most likely unsupported?
I'm pretty sure you have to add this
```
Dependencies:
  WindowsFeatures:
  - 'Virtual Machine Platform'
  - 'Windows Subsystem for Linux'
```

However Dependencies is not supported yet, and definitively not WindowsFeatures, however the [Schema](https://raw.githubusercontent.com/microsoft/winget-cli/master/schemas/JSON/manifests/v1.1.0/manifest.installer.1.1.0.json) accepts it so it will pass validation

ItzLevvie (changes requested):
​

felipecrs:
I'm booting up a virtual machine here for testing this.

felipecrs:
@OfficialEsco I converted to draft as you asked. I also tested this more thoroughly and the actual requirement for the installation is the "Windows Subsystem for Linux" feature only (tested in a Virtual Machine).

@ItzLevvie thanks a lot for the feedback.

felipecrs:
Ok, the same issue again. I just have one concern, should the WindowsFeature be named like `Windows Subsystem for Linux`:

image

Or `Microsoft-Windows-Subsystem-Linux`?

```console
dism.exe /online /enable-feature /featurename:Microsoft-Windows-Subsystem-Linux /all /norestart
```

I'm asking this because it seems that this is the first manifest using such key.

Anyway, would it be possible for the test to automatically detect when a Windows Feature is required and enable it? It's not needed to enable _Virtual Machine Platform_ or to have nested virtualization capabilities according to my tests to make the installation succeed.

OfficialEsco:
Just gonna reply so you don't tinker around endlessly 

> However Dependencies is not supported yet, and definitively not WindowsFeatures, however the Schema accepts it so it will pass validation

By this i meant that it will pass `winget validate`/step 1 of wingetbot's validation, but it will not pass step 2 (installing) because winget-cli does currently not do anything with the Dependencies metadata.

So until @denelon replies and technically until WindowsFeatures is supported sit back and relax =)

felipecrs:
Ok, got it. Thank you.

denelon:
I'm still working on the specification for dependencies. It's a monster! I'm having to involve multiple teams to get all the experts on the same page with how we should handle all these scenarios. The initial entries in the schema for dependencies were very rough assumptions. We are looking at what APIs we can use to enable Windows Features, and we're looking at all the options related to rebooting and resuming.

My gut is that we would use the /featurname enumeration. Initially we would just inform the user of the need to enable the feature, and eventually we would get to turning it on for users (based on settings / Group Policy / etc...).

denelon:
@felipecrs sorry for the late reply. I've been out sick all week. I'm just now starting to feel a bit more like a human and catching up on things. I will be referencing this in the specification to help drive clarity and make sure we get the right feature built here.

felipecrs:
@denelon, no worries at all! I hope all is well for you now.

While the support for _Dependencies_ and specially _Windows Features_ is amazing and will hopefully be very helpful feature of WinGet in future, I just wonder if there isn't another way to proceed without it for now. Because, currently, this is what happens:

If users do not have WSL enabled, `winget install Microsoft.WindowsSubsystemForLinuxKernel` fails. Which is not perfect, but is reasonable, as the official documentation makes this requirement clear.
  - But we could temporarily also add this note to the Manifest description, until support for Windows Features isn't provided by WinGet.

If the feature is enabled, it works flawlessly, which means that `winget upgrad`ing would also work.

If, however, there isn't a way to set a package as exempted on the automated installation tests, one approach would be to have the WSL feature pre-enabled in the Windows image used in Azure Pipelines (but I don't know if a customized Windows image for winget-pkgs already exist, or it just uses a prebuilt default image for Azure -- if it's the latter case I would agree that it may not worth the effort).

---

Off-topic, but for reference, I discovered that there is a PowerShell scriptlet in Windows for enabling optional features too:

```powershell
Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Windows-Subsystem-Linux -NoRestart
```

As an alternative to:

```cmd
dism.exe /online /enable-feature /featurename:Microsoft-Windows-Subsystem-Linux /all /norestart
```

denelon:
Do we know exactly what error is returned on failure? Maybe we could use "ExpectedReturnCodes".

felipecrs:
> Do we know exactly what error is returned on failure? Maybe we could use "ExpectedReturnCodes".

Yes! That's a great idea. I made the changes. Let's see... (IDK if the tests already takes manifest 1.1.0 into account)

felipecrs:
[As pointed by](https://github.com/microsoft/winget-pkgs/pull/40444#pullrequestreview-849053746) @ItzLevvie and on my own tests, the return code seems to be 1603.

I don't know if I did something wrong in the manifest, but the installer validation seems to be still failing. Is there any way we can go deeper in the logs?

denelon:
I've asked an engineer to take a look at this one.

OfficialEsco:
> Hello @felipecrs, The package manager bot determined there was an issue with installing the application correctly. Please check the application installs correctly. Once repaired, please push an update to your pull request.

```
Enter a pull request number: #40444

Found Windows Subsystem for Linux Kernel [Microsoft.WindowsSubsystemForLinuxKernel] Version 5.10.16
Downloading https://wslstorestorage.blob.core.windows.net/wslblob/wsl_update_x64.msi
  ██████████████████████████████  16.3 MB / 16.3 MB
Starting package install...
Installer failed with exit code: 1603
Installer log is available at: C:\Users\User\AppData\Local\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir\WinGet-Microsoft.WindowsSubsystemForLinuxKernel.5.10.16-2022-01-14-23-16-31.431.log
This package has a dependency missing from your system.
```

I don't believe the validation will pass just because you've hit a `ExpectedReturnCode`?
AFAIK all that is used for is to describe the issue.

Eitherway you need to pass wingetbot validation step 3 and 4 which is `Detect the installation` and `Execute the Software`

So in theory if they enable WSL for the Pipeline and rerun the PR it will pass the validation, but if a client is missing the dependency it will say so

denelon:
I'm going to mark this as blocked by dependency support for now. We don't really have a good way to validate a user's system should be able to run the package. We were getting exit code -1978334972 in one of the environments. It was failing due to the WSL dependency.

felipecrs:
@denelon alright. Thank you.

PR #40573, opened 2022-01-12
Add WSLtty from https://github.com/mintty/wsltty

me-and:

-----

[WSLtty](https://github.com/mintty/wsltty) is an open source terminal for connecting to Windows Subsystem for Linux, based on Mintty (the terminal used for Cygwin, and with a few modifications, for Git for Windows), which is in turn based on PuTTY.

(For the sake of explicitness: I am a Microsoft employee, but this project is well outside the scope of my normal work. I don't have any more context/experience/knowledge about winget than anyone else contributing to this project in their own time.)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/40573)


me-and:
I can't see what's causing the validation checks to fail. I do note that the warning about a lack of quiet flags seems spurious, as the installer.yaml file absolutely includes `InstallerSwitches: Silent: /Q`; my best guess is that that's warning about that flag not being present in a different YAML file and the warning's a hangover from when there was a single YAML file rather than a set.

OfficialEsco:
VirtualBoxVM_296uYQ109R

ItzLevvie (changes requested):
I'm seeing this error:
image

When I run the installer manually:
image

Perhaps this might be something to report to the developers of wsltty? cc @mintty

manifests/m/Mintty/WSLtty/3.5.1/Mintty.WSLtty.installer.yaml
```diff
@@ -0,0 +1,19 @@
+# Created using wingetcreate 0.4.4.1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+
+PackageIdentifier: Mintty.WSLtty
+PackageVersion: 3.5.1
+InstallerSwitches:
+  Silent: /Q
```
OfficialEsco, line 7: > I can't see what's causing the validation checks to fail.

```suggestion
InstallerSwitches:
  Silent: /Q
  SilentWithProgress: /Q
```

mintty:
> Perhaps this might be something to report to the developers of wsltty? cc @mintty

Please note that there is also a repository https://github.com/mintty/wsltty.appx, which would initially support manual installation of an appx file. But it never worked out to become installable via the usual means and the project is now very outdated.
I'd appreciate to get wsltty available as an installable app but I have no idea what the current problems are.

me-and:
@mintty I think the issue here is with the `/Q` switch.

If you run `wsltty-3.5.1-x86_64-install.exe /?`, you get a help dialog that suggests command line options including "/Q -- Quiet modes for the package," so I added `/Q` as the way for winget to run the installer quietly. By default, winget runs all installers in quiet mode, so that switch gets specified. But when I run the installer with that switch, I see the same error @OfficialEsco and @ItzLevvie saw: "Error creating process \<\<None\>\>. Reason: The system cannot find the file specified."

(The reason everything worked fine for me when I was testing, I think, is that I tested – as I nomally do – with `winget install --interactive ...`.)

So for the installer to be available as part of winget, there needs to be a quiet/silent/non-interactive mode for the WSLtty installer that actually works. I've not looked at how the WSLtty installer is generated, so I have no idea what work is necessary for that. Let me know if it'd be useful to raise an issue over at mintty/wsltty, though!

mintty:
I used the most basic and easily usable installer I could find a while ago, the quite antique iexpress which seems to be something intended to preconfigure internet explorer somehow. I would have liked to rather use a modern installer generator but was appalled by their complexity and the need to spend significant research into them before getting a result, so that's the story; iexpress itself has a /Q option but it doesn't make the resulting installer q-able.
Can you use a .bat file as an installer? A simple wrapper that accepts and ignores /Q?

mintty:
There is an initial "Install wsltty?" confirmation dialog, then a command window which logs all the extraction and setup commands, and finally a confirmation popup "Installed". The middle part, the command line log, can actually be suppressed. Would that help?

mintty:
Actually it is also possible to skip the messages. There will be an instant flicker of an initial window, for 0.1 s.
I can provide such an installer for testing if you like.

mintty:
> There will be an instant flicker of an initial window, for 0.1 s.

Managed to skip that too, so...

OfficialEsco:
Any installer submitted to WinGet need to be from the Official publisher, so if you want wsltty on WinGet please do =)
Will the new installer require no switches?

mintty:
Yes, no switches:
[wsltty-3.5.1-quiet-installer.zip](https://github.com/microsoft/winget-pkgs/files/7857608/wsltty-3.5.1-quiet-installer.zip)

OfficialEsco:
> Yes, no switches:
> [wsltty-3.5.1-quiet-installer.zip](https://github.com/microsoft/winget-pkgs/files/7857608/wsltty-3.5.1-quiet-installer.zip)

heh
VirtualBoxVM_XPFnl6uDaj
VirtualBoxVM_i61pWciy6C

mintty:
There have been repeated reports of virus suspicions over time in the wsltty issue tracker, all different.
It's likely because Windows does not trust its own legacy installer, or the old installer does not provide any security information, or whatever. I do not know how to sort that out. In case you have doubts, you are welcome to follow the installer creation procedure yourself from the repository, based on a minimal cygwin installation.

OfficialEsco:
https://github.com/mintty/wsltty/releases/download/3.5.1/wsltty-3.5.1-x86_64-install.exe seems to not be flagged tho, however if the new method is flagged for any of the AV's it will not pass one of the validation.

mintty:
I had also stripped some stale settings on that occasion which I've now reverted, so may this is a less different version:
[wsltty-3.5.1-quiet-installer2.zip](https://github.com/microsoft/winget-pkgs/files/7858006/wsltty-3.5.1-quiet-installer2.zip)

me-and:
IIRC winget doesn't currently support extracting files from zip archives,
so to test this we'd need to have it uploaded somewhere as a bare .exe
installer.

I'm guessing the restriction on only using official installers might mean
it also needs to be clearly attached to an official source -- so probably
on the wsltty repo, or somewhere on https://mintty.github.io/ -- but I
assume that we could use an .exe attached to this ticket to at least prove
winget is capable of downloading and installing the package.

mintty:
> prove winget is capable of ...

That's exactly the purpose of this upload of course. Just unpack the zip for a test case; github does not accept exe uploads.

OfficialEsco:
>  I've specifying Silent: null (which my reading of https://aka.ms/winget-manifest.installer.1.0.0.schema.json implies is correct), Silent: /Q and Silent:, all of which produced the same behaviour.

So this is why people use `Silent: null` or `Silent: false`/`true`?
the InstallerSwitches field is not required, so it can be removed, however it will throw a validation Warning
a Workaround to the warning is
```
InstallerSwitches
  Silent: ' '
  SilentWithProgress: ' '
```
I believe null refers to NullOrWhiteSpace

I executed the second installer yesterday, and it didn't get blocked, VirusTotal showed 4 Flags, one from Microsoft marked as Trojan which is weird since it didn't get blocked
A shortcut appeared on the Desktop but no `A`dd or `R`emove `P`rograms entries which might be intended behavior

mintty:
> A shortcut appeared on the Desktop but no Add or Remove Programs entries which might be intended behavior

A successful installation should give you one desktop shortcut, and N+1 Start Menu shorcuts (one for each WSL distribution available at the time of installation, plus one for the default WSL distribution), plus a WSLtty folder in the Start Menu with additional shorcuts and configuration scripts.

me-and:
So it sounds like that has confirmed the new silent installer works as
expected! @mintty are you able to upload that (presumably in both 32- and
64-bit flavours) to the main WSLtty repository? Once the .exe installer is
available there, it should be possible to update this PR with the new
links, and make WSLtty available through winget.

mintty:
I've skipped mintty 3.5.2 for wsltty, and I'm about to release mintty 3.5.3, possibly also with an update of the wslbridge2 gateway. Should I wait or would you still like an upload of the 3.5.1 release?

me-and:
It sounds like waiting makes sense if it's going to happen imminently :)

ItzLevvie (dismissed):

me-and:
I've just converted this PR to a draft: notwithstanding it having moderator approval already, I don't think it's ready to merge until the new wsltty release, with the silent installer, is available to download from the wsltty GitHub repo (or some other official source, but definitely not the spare system I happen to be able to serve files from!)

mintty:
Released wsltty 3.5.3: https://github.com/mintty/wsltty/releases/tag/3.5.3

OfficialEsco (approved):

mintty:
Thank you, I appreciate this very much.
How should future releases be handled? Should I notify them here?

OfficialEsco:
* If you want to maintain it your self you can make some automation or use wingetcreate.
* You can request the package to be added to https://github.com/vedantmgoyal2009/winget-pkgs-automation (Will most likely be added some time eitherway)
* Or since you're a good boy that release to GitHub, we with RSS feeds will eventually catch the update until MS adapts the automation

mintty:
Released wsltty 3.6.0.

OfficialEsco:
> Released wsltty 3.6.0.

Microsoft Defender dosn't really like you tho
```
PS M:\> winget install -m m\Mintty\WSLtty\3.6.0\
Found WSLtty [Mintty.WSLtty] Version 3.6.0
Downloading https://github.com/mintty/wsltty/releases/download/3.6.0/wsltty-3.6.0-x86_64-install-quiet.exe
  ██████████████████████████████  2.40 MB / 2.40 MB
An anti-virus product reports an infection in the installer
```
VirtualBoxVM_oQmfDuIogV

PR #40984, opened 2022-01-15
enabled use of InstallLocation for Oracle.JDK.17

Cube707:

-----

made the `-l,--location` option functional for this package

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/40984)

### DIFF
--- a/manifests/o/Oracle/JDK/17/17.0.1.0/Oracle.JDK.17.installer.yaml
+++ b/manifests/o/Oracle/JDK/17/17.0.1.0/Oracle.JDK.17.installer.yaml
@@ -3,10 +3,17 @@
 
 PackageIdentifier: Oracle.JDK.17
 PackageVersion: 17.0.1.0
+InstallerType: msi
+InstallModes:
+- interactive
+- silent
+InstallerSwitches:
+  Silent: /qn
+  SilentWithProgress: /passive
+  InstallLocation: INSTALLDIR=<INSTALLPATH>
 Installers:
-- InstallerLocale: en-US
-  Architecture: x64
-  InstallerType: msi
+- Architecture: x64
+  InstallerLocale: en-US
   InstallerUrl: https://download.oracle.com/java/17/latest/jdk-17_windows-x64_bin.msi
   InstallerSha256: 61FE8EA15D2DAEC053593414D6BE0C930CAC9905F2A895FB9F88A0EFBCAEE2B6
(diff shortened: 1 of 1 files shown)


ItzLevvie (changes requested):
I saw https://stackoverflow.com/a/8841965; which is enabled by default in https://github.com/microsoft/winget-cli/blob/b79e4087821c993f83f7c4731bbde7aa25612c73/src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp#L461 for all packages using MSI

Cube707:
hm, ok than I must have bodged up somewhere else, because the `-l` option wasn't working for me. I will test again....

Cube707:
@lutzroeder  did some more Testing and even on a fresh Win10 VM it isn't working as expected:

All of the following result in a installation to the default directory, *NOT* to the given directory:

`winget install Oracle.JDK.17 -l C:\ProgStuff\jdk`
`winget install Oracle.JDK.17 -l "C:\ProgStuff\jdk"`
`winget install Oracle.JDK.17 --override "/passive TARGETDIR=C:\ProgStuff\jdk"`

The only thing that work and results in a different installation directory is:

`winget install Oracle.JDK.17 --override "/passive INSTALLDIR=C:\ProgStuff\jdk"`

ItzLevvie:
I double-checked, looks like this installer supports `INSTALLDIR` instead of `TARGETDIR`.

ItzLevvie (approved):

Cube707:
reading [the link you provided](https://stackoverflow.com/a/8841965), as I understand it, this is also the correct behavior:

> `TARGETDIR` represents [..] when the user runs Setup.exe or MsiExec.exe *with the /a command-line switch* 

the targetdir option may not be the correct option here? The `/a` switch for the msi process is not used or is it?

Cube707:
@ItzLevvie I tested a few more packages and there seems to be some more inconsistency.

`ChristianHohnstadt.xca` and `StrawberryPerl.StrawberryPerl` seem to suffer from the same problem while `Borvid.HttpMasterExpress` works absoloutely fine as is and doesn't work with the INSTALLDIR option.

just using both options at the same time seems to work fine though. At least in my quick tests.

should I open a issue about this?

OfficialEsco:
> should I open a issue about this?

Please do, this will break waaaaaaaaaaaaay too many Packages

PR #43095, opened 2022-01-26
Safing.Portmaster version 0.7.14

derkrasseleo:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/43095)

### DIFF
--- a/manifests/s/Safing/Portmaster/0.7.14/Safing.Portmaster.installer.yaml
+++ b/manifests/s/Safing/Portmaster/0.7.14/Safing.Portmaster.installer.yaml
@@ -1,0 +1,21 @@
+# Created using wingetcreate 0.4.4.1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.1.0.schema.json
+
+PackageIdentifier: Safing.Portmaster
+PackageVersion: 0.7.14
+Platform:
+- Windows.Desktop
+MinimumOSVersion: 10.0.0.0
+Scope: machine
+InstallModes:
+- interactive
+- silent
+UpgradeBehavior: install
+Installers:
+- Architecture: x86
+  InstallerType: nullsoft
+  InstallerUrl: https://updates.safing.io/latest/windows_amd64/packages/portmaster-installer.exe
+  InstallerSha256: 2A7EA1568F83F5579C9921975A05AAF7E853BC29E5F5009D86EE8C1C27C5D336
+ManifestType: installer
+ManifestVersion: 1.1.0
+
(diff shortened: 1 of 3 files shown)


ItzLevvie (approved):

leochras:
The hash is the same?!

leochras:
It's an update, so the hash should be the same right?

ItzLevvie:
> It's an update, so the hash should be the same right?

Looks like `0.7.0` installs `0.7.14` so you should probably fix the version in the `0.7.0` package.

https://github.com/safing/portmaster/releases/tag/v0.7.14

leochras:
> > It's an update, so the hash should be the same right?
> 
> Looks like `0.7.0` installs `0.7.14` so you should probably fix the version in the `0.7.0` package.
> 
> https://github.com/safing/portmaster/releases/tag/v0.7.14

The 0.7.0 manifest's download URL is `https://updates.safing.io/windows_amd64/packages/portmaster-installer_v0-7-0.exe`
So it probably installs 0.7.0. But I downloaded both versions, and they seem to have the exact same size and hash, so maybe they're the same after all.
So should I update the version AND the installer URL of the 0.7.0 manifest?

OfficialEsco:
> Because the Portmaster auto-updates, the installers we distribute are only minimal and usually only describe themselves, similar to all these "light or web installers" that only download the real deal when they are run.
> 
> Based on that, I have the following feedback:
> 
>   * Depending on the usage and behavior of winget, possibly use an identifier specific to the installer: `Safing.Portmaster.Installer`
>     
>     * While this would be more technically correct, just using `Safing.Portmaster` probably makes more sense for all practical use cases.
> 
>   * Use the version of the installer: v0.5.2
> 
>   * Use the direct link to the installer, not the "latest" redirect: https://updates.safing.io/windows_amd64/packages/portmaster-installer_v0-5-2.exe
>       
>     * Especially as you are providing a hash for it.
> 
>   * Mention to the user that the Portmaster auto-updates when doing a silent install, if possible.
> 
> 
> Also, we will be issuing an update to the installer within the next month that will change the installation experience: The Portmaster service will not be started automatically after the install finished, but the user will be prompted to restart the device to complete installation. It would be interesting how this could be communicated in the install process here. If the installer is run interactively, then this should not be a problem, but I wonder how the silent installs would then work.
> 
> I'm not all too familiar with winget, so take my advice with a pinch of salt.
>
> _Originally posted by dhaavi in https://github.com/microsoft/winget-pkgs/issues/11651#issuecomment-827931214_

leochras:
Thank you for clarifying! Didn't know the devs were already on it..

PR #45416, opened 2022-02-08
New version: Kitware.CMake version 3.23.0-rc1

vedantmgoyal9:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/45416)

### DIFF
--- a/manifests/k/Kitware/CMake/3.23.0-rc1/Kitware.CMake.installer.yaml
+++ b/manifests/k/Kitware/CMake/3.23.0-rc1/Kitware.CMake.installer.yaml
@@ -1,0 +1,29 @@
+# Created with YamlCreate.ps1 v2.0.6 using InputObject 🤖 $debug=QUSU-7-2-1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.1.0.schema.json
+
+PackageIdentifier: Kitware.CMake
+PackageVersion: 3.23.0-rc1
+Platform:
+- Windows.Desktop
+MinimumOSVersion: 10.0.0.0
+InstallerType: wix
+Scope: machine
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+InstallerSwitches:
+  Custom: ADD_CMAKE_TO_PATH=System
+UpgradeBehavior: install
+ReleaseDate: 2022-02-08
+Installers:
+- Architecture: x86
+  InstallerUrl: https://github.com/Kitware/CMake/releases/download/v3.23.0-rc1/cmake-3.23.0-rc1-windows-i386.msi
+  InstallerSha256: 4F210A0595C5DBEE09C8F341F88BBCFD9FB7AE7689ACE3312818E230C2D54372
+  ProductCode: '{978FD729-1476-455B-9975-CE5963F88FDD}'
+- Architecture: x64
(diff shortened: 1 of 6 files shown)


KaranKad (changes requested):
This is a release client version but with no way do differentiate it from stable release, so it might be better to not add it.

vedantmgoyal2009:
I don't know much about release management, but according to this, I think we can add it because "this release goes into production". 
image

manifests/k/Kitware/CMake/3.23.0-rc1/Kitware.CMake.yaml
```diff
@@ -0,0 +1,8 @@
+# Created with YamlCreate.ps1 v2.0.6 using InputObject 🤖 $debug=QUSU-7-2-1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.version.1.1.0.schema.json
+
+PackageIdentifier: Kitware.CMake
+PackageVersion: 3.23.0-rc1
```
KaranKad, line 5:
```suggestion
PackageVersion: 3.23.0
```

manifests/k/Kitware/CMake/3.23.0-rc1/Kitware.CMake.locale.en-US.yaml
```diff
@@ -0,0 +1,29 @@
+# Created with YamlCreate.ps1 v2.0.6 using InputObject 🤖 $debug=QUSU-7-2-1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.1.0.schema.json
+
+PackageIdentifier: Kitware.CMake
+PackageVersion: 3.23.0-rc1
```
KaranKad, line 5:
```suggestion
PackageVersion: 3.23.0
```

manifests/k/Kitware/CMake/3.23.0-rc1/Kitware.CMake.installer.yaml
```diff
@@ -0,0 +1,29 @@
+# Created with YamlCreate.ps1 v2.0.6 using InputObject 🤖 $debug=QUSU-7-2-1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.1.0.schema.json
+
+PackageIdentifier: Kitware.CMake
+PackageVersion: 3.23.0-rc1
```
KaranKad, line 5:
```suggestion
PackageVersion: 3.23.0
```

KaranKad (changes requested):

manifests/k/Kitware/CMake/3.23.0-rc1/Kitware.CMake.yaml
KaranKad: image

ItzLevvie (approved):

Carbenium:
@vedantmgoyal2009 @KaranKad @ItzLevvie Please consider not adding RCs (Release Candidates) to winget without an explicit opt-in. The default should be stable releases only.
The citation given by @vedantmgoyal2009 (from https://en.wikipedia.org/wiki/Software_release_life_cycle) is not really applicable. Yes, it might be true that the last RC before stable is promoted to the stable version, but that does not mean that RC1 (the first RC) is actually going stable (as seen with 3.23.0-rc2).

PR #45636, opened 2022-02-09
DRAFT rcmaehl.MSEdgeRedirect version 0.6.2.0

Macleykun:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/45636)

### DIFF
--- a/manifests/r/rcmaehl/MSEdgeRedirect/0.6.2.0/rcmaehl.MSEdgeRedirect.installer.yaml
+++ b/manifests/r/rcmaehl/MSEdgeRedirect/0.6.2.0/rcmaehl.MSEdgeRedirect.installer.yaml
@@ -1,0 +1,13 @@
+# Created using wingetcreate 0.4.4.1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+
+PackageIdentifier: rcmaehl.MSEdgeRedirect
+PackageVersion: 0.6.2.0
+Installers:
+- Architecture: x64
+  InstallerType: exe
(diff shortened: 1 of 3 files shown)


Trenly (changes requested):
This might be blocked as interactive-only due to this message -
image

Macleykun:
> This might be blocked as interactive-only due to this message - image

I'm deeply greatful for the incredible fast response and feedback on this pull!
Regarding that message, while i'm not the main developer of said program. I do agree that this will be an interactive only.

For any futher questions regarding improving the installation experience of set program, i would like to ask @rcmaehl for his insight.

Trenly:
> I do agree that this will be an interactive only.

Unfortunately, at this point, all installers must support silent install before they can be added into the repository. Assuming that the pipelines will catch this, you can set this PR to draft mode and then we can label this to revisit it once interactive installers are supportive

Macleykun:
> > I do agree that this will be an interactive only.
> 
> Unfortunately, at this point, all installers must support silent install before they can be added into the repository. Assuming that the pipelines will catch this, you can set this PR to draft mode and then we can label this to revisit it once interactive installers are supportive

Sure thing! Thank you agani for your feedback! I'll talk to the dev/check the documentation if we have a quiet mode!

Trenly:
@denelon - Can you tag this please?
Blocking-Issue, Interactive-Only-Installer

rcmaehl:
`/silentinstall` will work but it'll want a setup.ini file. I'll add the option to skip the file requirement with some default options.

PR #45676, opened 2022-02-10
remove Money.Net version 1.1.0.87

clovett:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/45676)


OfficialEsco:
Why do you remove old versions?... And why not just overwrite the current manifest?

clovett:
Because the version number is in the path: manifests/l/LovettSoftware/MyMoney/Net/`1.1.0.85`/LovettSoftware.MyMoney.Net.installer.yaml

I suppose I could remove the version number from the path, but all the manifests seem to be authored this way... for example: ~\m\Microsoft>dir SQLServer\2019\Developer\`15.0.2000.5`\Microsoft.SQLServer.2019.Developer.yaml

OfficialEsco (approved):
Sorry my comment was unclear, you're allowed to add and remove one version in a PR (under the same software folder), GitHub should threat it as Move and Update and would show as moving the manifest and updating the PackageVersion, InstallerUrl and Hashes

I don't see the reason for removing old versions tho, the newest version would always be installed when you do winget install

clovett:
I see, thanks.  I was told before I could not delete and old version and add a new version in one PR, but perhaps what you are saying is I can do a "git mv".  Unfortunately, my CI pipeline is not setup to do "git mv" but I will consider changing that.  Doing one PR per release would be nice :-)

OfficialEsco:
It worked like that before, but we contributors made them allow delete + add since even GitHub Desktop sees it as Move and Modify, + IMO the history looks cleaner

> So for why delete old version at all, I thought I was just being good citizen by not letting the winget-pkgs repo grow unbounded with junk...

Its a Package Manager, we're supposed to allow installation of older versions instead of forcing everyone to download the latest version (Its only clutter if it Auto Updates to the latest version)

But you're the Publisher so you obv do what you want, hopefully you adapt one of these methods in the future =)

Also at the moment your PR Name is confusing as you say you remove the same version as you added, but you actually removed `1.1.0.85` instead of `1.1.0.87`

clovett:
ok, thanks for the bug report on the PR title, I'll be sure to fix that next time.

PR #46123, opened 2022-02-14
Swift.Toolchain 5.5.3

svanimpe:

-----

I've generated these using `YamlCreate` and the manifest version is 1.1.0 instead of 1.0. I assume that's OK?

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/46123)


manifests/s/Swift/Toolchain/5.5.3/Swift.Toolchain.installer.yaml
```diff
@@ -0,0 +1,21 @@
+# Created with YamlCreate.ps1 v2.0.7 $debug=NVS1.7-2-1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.1.0.schema.json
+
+PackageIdentifier: Swift.Toolchain
+PackageVersion: 5.5.3
```
OfficialEsco, line 5:
```suggestion
PackageVersion: 0.0.0.0
```

manifests/s/Swift/Toolchain/5.5.3/Swift.Toolchain.locale.en-US.yaml
```diff
@@ -0,0 +1,27 @@
+# Created with YamlCreate.ps1 v2.0.7 $debug=NVS1.7-2-1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.1.0.schema.json
+
+PackageIdentifier: Swift.Toolchain
+PackageVersion: 5.5.3
```
OfficialEsco, line 5:
```suggestion
PackageVersion: 0.0.0.0
```

manifests/s/Swift/Toolchain/5.5.3/Swift.Toolchain.yaml
```diff
@@ -0,0 +1,8 @@
+# Created with YamlCreate.ps1 v2.0.7 $debug=NVS1.7-2-1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.version.1.1.0.schema.json
+
+PackageIdentifier: Swift.Toolchain
+PackageVersion: 5.5.3
```
OfficialEsco, line 5:
```suggestion
PackageVersion: 0.0.0.0
```

OfficialEsco (changes requested):
Ohhnooo this installer writes the wrong version to the registry 🤦‍♂️

```
Name        : Swift
Publisher   : dt.compnerd.org
Version     : 0.0.0.0
ProductCode : {51a6ef4e-285a-44fe-8f21-cdf5a6ee7245}
Uninstall   : winget uninstall --id "{51a6ef4e-285a-44fe-8f21-cdf5a6ee7245}"
```

Hmm, this is a Burn installer, would this be something one of you guys fix? @jedieaston @Trenly

jedieaston:
> Ohhnooo this installer writes the wrong version to the registry 🤦‍♂️
> 
> ```
> Name        : Swift
> Publisher   : dt.compnerd.org
> Version     : 0.0.0.0
> ProductCode : {51a6ef4e-285a-44fe-8f21-cdf5a6ee7245}
> Uninstall   : winget uninstall --id "{51a6ef4e-285a-44fe-8f21-cdf5a6ee7245}"
> ```
> 
> Hmm, this is a Burn installer, would this be something one of you guys fix? @jedieaston @Trenly

I looked into it last week. I spent like an hour trying to figure out how the heck the build scripts work but didn't get anywhere (it's like three repos, one of which is private so I can't see if they expose the version number to msbuild anywhere). If someone could drop something in the mailing list I bet they'd fix it.

manifests/s/Swift/Toolchain/5.5.3/Swift.Toolchain.yaml
svanimpe: May I ask what exactly this does? I assume it's a workaround, and not the value we actually want here?

OfficialEsco:
> If I may also ask for some advice: Am I following the correct procedure here for patch version upgrades? That is: Should I create them as new versions with new manifests? Or should I edit the existing 5.5 manifest to point to 5.5.3? I'm not sure how the updater works, so I figured it may be best to create a new manifest for each version.

You should create a new Manifest for each version yes, if it Auto updates you CAN delete and create a new manifest for the new version in the same PR (will be counted as Move and Modify as long as its in the same Software folder)

But we only do that for urls that don't contain versions like [Spotify](https://github.com/microsoft/winget-pkgs/pull/44727/files)
Since we can't remember every software that auto updates its best to make a brand new manifest for urls with versions

denelon:
It looks like we have 5.5 as the latest version in the repository currently. Unfortunately, without the package properly reporting the version to Windows Apps & Features, the Windows Package Manager would not be able to recognize this as an upgrade. 

@OfficialEsco what are your thoughts about PRs like this where the package doesn't report the correct version, or doesn't report one at all? Should we block on the related issue below? It's the feature where the client begins to understand the "AppsAndFeatures" entry in the manifest.

What additional label would make sense in that case? "Version" or "Apps-and-Features"? or something like that?

Related to:
* microsoft/winget-cli/issues/1952

OfficialEsco:
Yeah that makes sense, `DisplayVersion-Isssue` might be clearer than `Apps-and-Features` as Apps-and-Features issues can be solved by ProductCodes

denelon:
Sorry for the noise to come. I'm testing some additional automation.

ItzLevvie (changes requested):
Is this PR still needed?

Developers of the Swift Toolchain have fixed their MSI version in latest releases: https://github.com/microsoft/winget-pkgs/pull/57288

svanimpe:
I was unaware this had been fixed. So yes, we can close this!

PR #47068, opened 2022-02-19
Prevent auto restart on LibreOffice install

Kissaki:
Earlier LibreOffice manifests have these flags which prevent automatic system restarts on installation.

See also previous fixups 92ca7106023d0e2b1fa536f6502a0502ab3f06f9 and 1a5c1039ffb73edad007fd9ecaff8b08e335f1b3
and their related PRs and issue tickets.

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/47068)

### DIFF
--- a/manifests/t/TheDocumentFoundation/LibreOffice/7.2.5.1/TheDocumentFoundation.LibreOffice.installer.yaml
+++ b/manifests/t/TheDocumentFoundation/LibreOffice/7.2.5.1/TheDocumentFoundation.LibreOffice.installer.yaml
@@ -7,6 +7,9 @@ Platform:
 - Windows.Desktop
 MinimumOSVersion: 10.0.0.0
 InstallerType: msi
+InstallerSwitches:
+  Silent: /qn /norestart
+  SilentWithProgress: /qn /norestart
 Scope: machine
 UpgradeBehavior: install
 FileExtensions:
--- a/manifests/t/TheDocumentFoundation/LibreOffice/7.2.5.2/TheDocumentFoundation.LibreOffice.installer.yaml
+++ b/manifests/t/TheDocumentFoundation/LibreOffice/7.2.5.2/TheDocumentFoundation.LibreOffice.installer.yaml
@@ -7,6 +7,9 @@ Platform:
 - Windows.Desktop
 MinimumOSVersion: 10.0.0.0
 InstallerType: msi
+InstallerSwitches:
+  Silent: /qn /norestart
+  SilentWithProgress: /qn /norestart
 Scope: machine
 UpgradeBehavior: install
 FileExtensions:
--- a/manifests/t/TheDocumentFoundation/LibreOffice/7.3.0.3/TheDocumentFoundation.LibreOffice.installer.yaml
+++ b/manifests/t/TheDocumentFoundation/LibreOffice/7.3.0.3/TheDocumentFoundation.LibreOffice.installer.yaml
@@ -6,6 +6,9 @@ PackageVersion: 7.3.0.3
 Platform:
 - Windows.Desktop
 MinimumOSVersion: 10.0.0.0
+InstallerSwitches:
+  Silent: /qn /norestart
+  SilentWithProgress: /qn /norestart
 Scope: machine
 UpgradeBehavior: install
 FileExtensions:


manifests/t/TheDocumentFoundation/LibreOffice/7.2.5.1/TheDocumentFoundation.LibreOffice.installer.yaml
```diff
@@ -7,6 +7,9 @@ Platform:
 - Windows.Desktop
 MinimumOSVersion: 10.0.0.0
 InstallerType: msi
+InstallerSwitches:
+  Silent: /qn /norestart
+  SilentWithProgress: /qn /norestart
```
OfficialEsco, line 12: Also this can probably quiet with basic ui for progressbar?
```suggestion
  SilentWithProgress: /qb /norestart
```

OfficialEsco (changes requested):

OfficialEsco:
ops forgot to post comment
You can only change 1 version per PR

manifests/t/TheDocumentFoundation/LibreOffice/7.2.5.1/TheDocumentFoundation.LibreOffice.installer.yaml
Kissaki: For reference as I didn’t find any docs; from `msiexec /?`:

```
	/q[n|b|r|f]
		Sets user interface level
		n - No UI
		b - Basic UI
		r - Reduced UI
		f - Full UI (default)
```

I got the /qn from the other manifests, but yeah, /qb seems more appropriate. (Shouldn’t this be a default for all msi installers called by winget? Same for /norestart.)

OfficialEsco: by default it passes /quiet for silent and /passive for silent with progress
/qn is the same as /quiet
/qb shows only the progress bar
i've not tried /qr but its probably simple UI
/qf is probably normal ui but can be forced if the default is /qb or /qn

OfficialEsco:
> Is the multiple manifests why the bot claims the path is not correct?

yes it is

manifests/t/TheDocumentFoundation/LibreOffice/7.2.5.1/TheDocumentFoundation.LibreOffice.installer.yaml
Kissaki: Wouldn’t that mean /qn and /qb are not necessary? Or does specifying the InstallerSwitches replace the winget defaults?

OfficialEsco: if the InstallerType is defined anything else than exe you do not need to define the InstallerSwitches.

HOWEVER in this case where you NEED the /norestart switch and we don't have a NoRestart field you will need to define the install and silent switch + the norestart switch.

We can however try a experiment and put /norestart in the `Custom:` and `Upgrade:` field since `Custom:` is applied on Install and `Upgrade:` is applied on Upgrade
https://github.com/microsoft/winget-cli/blob/master/schemas/JSON/manifests/v1.1.0/manifest.installer.1.1.0.json#L88-L131

Kissaki: It happened to me on upgrade. I assume because the shell extensions were in use because file explorer/shell has been in use? So I presume it would only happen on upgrades.

Kissaki:
Superseded by #47070, #47071, #47072.

manifests/t/TheDocumentFoundation/LibreOffice/7.2.5.1/TheDocumentFoundation.LibreOffice.installer.yaml
Kissaki: > Upgrade
> Upgrade is the value that should be passed to the installer when user chooses an upgrade

implies /upgrade (or similar) would have to be passed too here. So we do not save in not specifying default flags.

Given that restarts are only expected on upgrades, that’d be a reason to use that instead. If upgrade does not touch the default quiet or not quiet flags we would not have to differentiate between those. But there’s no damage in always passing /norestart anyway - arguably simplifies reasoning.

> Custom
> Custom switches will be passed directly to the installer by winget

It’s not explicit about it, but if this is always in addition to other flags, then this seems like it would be the simplest solution. Setting Custom to /norestart.

OfficialEsco: Here is the default flags for MSI/Burn/Wix
https://github.com/microsoft/winget-cli/blob/b79e4087821c993f83f7c4731bbde7aa25612c73/src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp#L451-L461

Upgrade refers to `winget upgrade <ID>`

I've been told Custom only works with `winget install` (we've used /norestart in that field before and it didn't work with upgrade)

Until winget-cli gets support for NoReboot/NoRestart it would probably be simplest to keep using the Silent and SilentWithProgress switches yeah since there is no harm like you said

PR #50144, opened 2022-03-04
MSI.Afterburner version 4.6.3

BareGamer:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/50144)


OfficialEsco (changes requested):
We need the URL to be from either Guru3d or MSI, which uses Zip which is not supported yet.

BareGamer:
> We need the URL to be from either Guru3d or MSI, which uses Zip which is not supported yet.

Another problem with that is the fact that they both have a query download link instead of a direct one. Wingetcreate just doesn't work with this kind of links.

ItzLevvie (changes requested):
> Maybe it would be possible to somehow unzip the downloaded files before wingetcreate tries to parse them?

`.zip` support will be coming to WinGet in a future release.

PR #55052, opened 2022-03-20
Modify: iFlytek.iFlyIME version 3.0.1727

SpecterShell:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/55052)

### DIFF
--- a/manifests/i/iFlytek/iFlyIME/3.0.1727/iFlytek.iFlyIME.installer.yaml
+++ b/manifests/i/iFlytek/iFlyIME/3.0.1727/iFlytek.iFlyIME.installer.yaml
@@ -1,22 +1,24 @@
-# Created using YamlCreate.ps1 v1.1.6
-# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+# Created with YamlCreate.ps1 v2.1.1 $debug=QUSU.7-2-2
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.1.0.schema.json
 
 PackageIdentifier: iFlytek.iFlyIME
 PackageVersion: 3.0.1727
 MinimumOSVersion: 10.0.0.0
-FileExtensions:
-- it
+InstallerType: nullsoft
+Scope: machine
 InstallerSuccessCodes:
 - 2
+UpgradeBehavior: install
+FileExtensions:
+- it
 Installers:
-- InstallerLocale: zh-CN
-  Architecture: neutral
-  InstallerType: nullsoft
-  Scope: machine
+- Architecture: x64
   InstallerUrl: https://download.voicecloud.cn/200ime/iFlyIME_Setup3.0.1727.exe
   InstallerSha256: 4DA3940CA59A84AA13DF219AF9B1A66C5AA42FF5EF28E6EA6F9607BBA1FDD36D
-  ProductCode: 'iFlyIME'
-  UpgradeBehavior: install
+  ProductCode: iFlyIME
+- Architecture: x86
+  InstallerUrl: https://download.voicecloud.cn/200ime/iFlyIME_Setup3.0.1727.exe
+  InstallerSha256: 4DA3940CA59A84AA13DF219AF9B1A66C5AA42FF5EF28E6EA6F9607BBA1FDD36D
+  ProductCode: iFlyIME
 ManifestType: installer
-ManifestVersion: 1.0.0
-
+ManifestVersion: 1.1.0
--- a/manifests/i/iFlytek/iFlyIME/3.0.1727/iFlytek.iFlyIME.locale.en-US.yaml
+++ b/manifests/i/iFlytek/iFlyIME/3.0.1727/iFlytek.iFlyIME.locale.en-US.yaml
@@ -1,26 +1,33 @@
-# Created using YamlCreate.ps1 v1.1.6
-# yaml-language-server: $schema=https://aka.ms/winget-manifest.locale.1.0.0.schema.json
+# Created with YamlCreate.ps1 v2.1.1 $debug=QUSU.7-2-2
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.1.0.schema.json
 
 PackageIdentifier: iFlytek.iFlyIME
 PackageVersion: 3.0.1727
 PackageLocale: en-US
-Publisher: Iflytek Co.,Ltd.
-PublisherUrl: https://www.xunfei.cn
-PublisherSupportUrl: http://bbs.xunfei.cn
+Publisher: 科大讯飞股份有限公司
+PublisherUrl: https://www.xunfei.cn/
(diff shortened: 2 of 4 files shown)


ItzLevvie (approved):

manifests/i/iFlytek/iFlyIME/3.0.1727/iFlytek.iFlyIME.installer.yaml
```diff
@@ -1,22 +1,24 @@
...
-  Scope: machine
+- Architecture: x64
   InstallerUrl: https://download.voicecloud.cn/200ime/iFlyIME_Setup3.0.1727.exe
   InstallerSha256: 4DA3940CA59A84AA13DF219AF9B1A66C5AA42FF5EF28E6EA6F9607BBA1FDD36D
-  ProductCode: 'iFlyIME'
-  UpgradeBehavior: install
+  ProductCode: iFlyIME
+- Architecture: x86
+  InstallerUrl: https://download.voicecloud.cn/200ime/iFlyIME_Setup3.0.1727.exe
+  InstallerSha256: 4DA3940CA59A84AA13DF219AF9B1A66C5AA42FF5EF28E6EA6F9607BBA1FDD36D
```
JohnMcPMS, line 21: This one exe can't be both x86 and x64.

JohnMcPMS (changes requested):
Fix the installer duplicate.

manifests/i/iFlytek/iFlyIME/3.0.1727/iFlytek.iFlyIME.installer.yaml
OfficialEsco: Actually it can (which is why there is neutral), however we're moving away from neutral to splitting them into x64, x86, arm and arm64 (decided by the wingetcreate crew) so it gives us better visual of when they remove the x86 support.

However this installer seems to be x86 only

SpecterShell: While the IME config tool is x86, the IME core has both x64 and x86 architectures and no arm(64) ones.

JohnMcPMS: If the installer is x86 exe, but determines which architecture to pull down resources for, it should still be marked as x86.  This manifest won't stop the install attempt on arm64, as x86 (and x64 in later versions) are supported on arm64 Windows Desktop.  But if x86 isn't supported on the particular image of x64 Windows (it can be the case), then the "x64" installer would still be used and fail to run.

What we need is the ability to state specific machine architecture support, or to state that the installer will not work on arm64 (or whatever architecture), or the ability to indicate that the architecture must match the system exactly (due to drivers for example).  Lying about the architecture of the installer isn't the way to go.

Or to split out the concerns into `InstallerArchitecture`, `InstalledBinaryArchitectures`, `SupportedSystemArchitecture`, and `UnsupportedSystemArchitectures` so that the complex cases can be fully realized.

I will take my fight up with "the wingetcreate crew" for now though, this change doesn't need to be blocked for that.

JohnMcPMS (approved):

manifests/i/iFlytek/iFlyIME/3.0.1727/iFlytek.iFlyIME.installer.yaml
OfficialEsco: AHA i understand the issue here, the systems are definitively not ready for this, currently the Architecture is being tested on THAT specific Architecture, neutral is ONLY being tested on x64 when it should definitively be tested on x64, x86, arm and arm64 to check if it actually is neutral.

With your logic we need to start using `UnsupportedOSArchitectures` but wingetbot still needs to test all the Architectures.

wingetcreate does pick up the INSTALLER Architecture and have selected x86 for many packages which does not support being installed on a x86 system.

We the winget-pkgs contributors are mostly selecting the Architecture based on the Software Process Architecture, A LOT of the Packages are currently x64 because we don't have a x86 VM to check if the installer can install the x64 and x86 Software.
Therefore because we can finally see the Artifact logs we are splitting the Neutral architecture into x64 and x86.

The Architecture description also says "The installer target architecture" which IMO does not specify that it should be the Installers Architecture.

Trenly: I think the main point of contention here is what `Architecture` was intended for versus what it has been used for, and how it has been implemented in the client. It seems that from the contributor standpoint, `Architecture` refers to the supported OS architectures for the application. I would argue that this is the same logic that the validation pipelines and the cli have been built with - where the `Architecture` determines the supported OS architecture(s). The client prefers an architecture of x64 on an x64  machine - If the `Architecture` were to be the installer architecture instead of the application architecture, then there would be cases where there may be multiple entries listed as x86 when only one of them works for x64 - specifically when publishers create different binaries for installing on x64 and x86 platforms but package both of the installer files as x86. As Esco mentioned, there are many cases where an installer that has an x86 architecture will only work on an x64 OS. The interpretation of `Architecture` as the installer architecture could also lead to weird metadata where `Architecture` is listed as x86 but `UnsupportedOSArchitectures` is also listed as x86. Given that the validation pipelines here at pkgs use the `Architecture` to determine which VM to boot up, then it would create an x86 VM for an installer that doesn't work on x86?

Even if the original intent for `Architecture` was to be the architecture of the installer, the way that the community has treated it and the way it has been implemented seems to have been more closely aligned to the concept of `SupportedOSArchitecture`. 

-----
Overall, I agree with and see the need for determining complex architecture scenarios, and would fully support having `InstallerArchitecture`, `SupportedOSArchitectures`,`UnsupportedOSArchitectures` and other fields as necessary. However, I believe they should be part of a 2.0 manifest schema where `Architecture` is either removed or renamed to be more descriptive of its current use as `SupportedOSArchitecture`

Note - here is the issue Esco mentioned which asks for Architecture to be an Array (although that ask is also part of an ask to remove `neutral`) https://github.com/microsoft/winget-cli/issues/1981

JohnMcPMS: I agree 100% that the best interpretation of a single `Architecture` value is the one that is being used commonly; the architecture of the binaries delivered to the system by the installer.  It is a failing on my part to not have foreseen the complexity here, although I can imagine that I would have had pushback to needing 4 different fields in the manifest 😄 

I think we could still support both without a 2.0, it is not an uncommon scenario for the installer and binaries to be the same architecture, and to have no reasons to exclude any specific architectures.  Given that is the only thing we have it must work most of the time.  But to then allow for more complex situations like this by adding additional details would reduce the chances of requiring a download + install attempt to find out it doesn't work.

PR #55107, opened 2022-03-20
Slonopotamus.Stevedore version 0.3.2

slonopotamus:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/55107)

### DIFF
--- a/manifests/s/Slonopotamus/Stevedore/0.3.2/Slonopotamus.Stevedore.installer.yaml
+++ b/manifests/s/Slonopotamus/Stevedore/0.3.2/Slonopotamus.Stevedore.installer.yaml
@@ -1,0 +1,20 @@
+# Created using wingetcreate 0.5.0.1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.1.0.schema.json
+
+PackageIdentifier: Slonopotamus.Stevedore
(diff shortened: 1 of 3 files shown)


slonopotamus:
I'm not sure what's winget policy on accepting _older_ releases of software, but thought it could be convenient to users if they could install them. If this is OK, I'll follow-up with PRs for other versions too. If not, just say so and close this PR.

Trenly:
> I'm not sure what's winget policy on accepting _older_ releases of software, but thought it could be convenient to users if they could install them. If this is OK, I'll follow-up with PRs for other versions too. If not, just say so and close this PR.

As long as the older versions pass validation, there usually isn't a problem with it. Occasionally when there are many versions of a package it is best to only have the latest patch of each minor revision of the format x.x.* (Example, keep 2.1.4 but not 2.1.3 or 2.1.2, keep 2.0.9, but not 2.0.8 etc.). If there aren't a lot of versions, having all of them available isn't an issue

slonopotamus:
Thanks for explanation. After some more thinking, currently published Stwvedore release is a very good one, and there should not be any reasons to use older versions. I think I'll leave everything as it is currently.

PR #55398, opened 2022-03-23
MyFaronics.MyAnti-Virus version 4.31.3102.572

goreswapnil:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/55398)


ItzLevvie (changes requested):
It's failing because the installer reboots the VM with 3010.

I tried `/norestart` but it doesn't work.

JohnMcPMS:
I don't really know how to read the MSI logs, but late in them, even with `/norestart` manually added, I see:
```
Property(S): REBOOT = Force
```
So I think that the MSI will always reboot.  Given that I have created https://github.com/microsoft/winget-cli/issues/2041 and this PR should probably be blocked on that feature.  I don't really know if we can support automated validation with reboots though, but we should probably not treat it as an internal error if we can help it.

PR #55498, opened 2022-03-23
Remove 21.05 because it appears to be gone from the origin server

JohnMcPMS:
21.05 is no longer listed at:
https://sourceforge.net/projects/sevenzip/files/7-Zip/

And the x86 and x64 binaries no longer resolve.  The arm64 does, but given that the folder is entirely removed on the origin server, it seems likely that it is a mistake.  I can't find why it was removed, but given that we already have 21.06 and 21.07, it seems like the correct thing to do is to simply remove the whole version in case it was a security issue.  Also, according to the origin server, 9 downloads per week of 21.07 ARM64 and no one is even downloading 21.06 ARM64, so I don't think much will be lost here.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/55498)


OfficialEsco:
Duplicate of #55497 👀

JohnMcPMS:
Beat me by 12 minutes!  If only I hadn't tried to update through the auto-update PR and then tried to see if we had intentionally not enabled that PR feature.

PR #55850, opened 2022-03-26
New version: Microsoft.WindowsTerminal 1.12.10733.0

OfficialEsco:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/55850)

### DIFF
--- a/manifests/m/Microsoft/WindowsTerminal/1.12.10733.0/Microsoft.WindowsTerminal.installer.yaml
+++ b/manifests/m/Microsoft/WindowsTerminal/1.12.10733.0/Microsoft.WindowsTerminal.installer.yaml
@@ -1,0 +1,29 @@
+# Created with YamlCreate.ps1 v2.1.2 $debug=QUSU.7-2-2
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.1.0.schema.json
+
+PackageIdentifier: Microsoft.WindowsTerminal
+PackageVersion: 1.12.10733.0
+Platform:
+- Windows.Desktop
+MinimumOSVersion: 10.0.18362.0
+InstallerType: msix
+InstallModes:
+- silent
+Commands:
+- wt
+PackageFamilyName: Microsoft.WindowsTerminal_8wekyb3d8bbwe
+Installers:
+- Architecture: x64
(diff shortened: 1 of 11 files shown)


ItzLevvie (approved):

manifests/m/Microsoft/WindowsTerminal/1.12.10733.0/Microsoft.WindowsTerminal.installer.yaml
```diff
@@ -0,0 +1,32 @@
...
+InstallerType: msix
+InstallModes:
+- silent
+Commands:
+- wt
+PackageFamilyName: Microsoft.WindowsTerminal_8wekyb3d8bbwe
+ReleaseDate: 2022-03-25
+Installers:
+- Architecture: x64
+  MinimumOSVersion: 10.0.22000.0
```
OfficialEsco, line 17: @JohnMcPMS whatchu gonna do about this one 🤨

JohnMcPMS: > This version of Windows Terminal is now distributed in two bundles, one of which works on Windows 10-11 and the
other of which only works on Windows 11. The Windows 11 version is much smaller because we no longer need to work around
a platform issue related to our dependencies.

Although I think you know this considering you already got the Win10 one in 😉 .  If we accept this Win 11 version its going to expose the issue that `install` doesn't do any fall back like `upgrade` does, meaning that new installs on Win10 will fail rather than just install the latest version that supports Win10.

@denelon , another validation that fails because the server isn't Win11 yet.  We probably want to bump the priority on getting it upgraded, while also fixing the client as I stated above.

OfficialEsco: Yeap i kinda wonder why its not the same version.. would make it so much easier for us

JohnMcPMS: MSIX requires that packages with the same version have the same contents (and Terminal knows this for other reasons).

PR #57668, opened 2022-04-13
change package ID

tushev:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/57668)


tushev:
Pipeline error: `(Addition and deletion only permitted in the same package directory.)`

Should I create separate pull requests for deletion of the old manifest and for the new one?

ItzLevvie (changes requested):
`PackageName` and `Publisher` are intentional because WinGet requires this to match Control Panel. If you would like to change this then you should do it in the application level and then make the changes in the manifest.

`PackageIdentifier` follows the publisher.application format but you can shorten the application name in the PackageIdentifier if you think it's too long.

`Moniker` can be adjusted to your liking but I believe it's only limited to **one** so the old one will need to be removed.

Trenly:
> Pipeline error: `(Addition and deletion only permitted in the same package directory.)`
> 
> Should I create separate pull requests for deletion of the old manifest and for the new one?

Yes, this is exactly what that error means

PR #58746, opened 2022-04-25
Uninstall before VC++2015-2022 before upgrading

felipecrs:
Otherwise winget upgrade will always flag as outdated, because the older package will be still there.

-----

Otherwise, the old version is kept installed (since multiple versions of it can be installed at the same time), and therefore, `winget upgrade` always flags the package as outdated unless the user manually uninstalls it.

I can confirm and reproduce this behavior for the VC++2015-2022 packages, and although I believe this can also be the case for other variants of it, I didn't try to reproduce it.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/58746)

### DIFF
--- a/manifests/m/Microsoft/VC++2015-2022Redist-x64/14.30.30704.0/Microsoft.VC++2015-2022Redist-x64.installer.yaml
+++ b/manifests/m/Microsoft/VC++2015-2022Redist-x64/14.30.30704.0/Microsoft.VC++2015-2022Redist-x64.installer.yaml
@@ -22,7 +22,7 @@ Installers:
   InstallerSwitches:
     Silent: /quiet /install /norestart
     SilentWithProgress: /passive /install /norestart
-  UpgradeBehavior: install
+  UpgradeBehavior: uninstallPrevious
   ProductCode: '{57a73df6-4ba9-4c1d-bbbb-517289ff6c13}'
 ManifestType: installer
 ManifestVersion: 1.0.0
--- a/manifests/m/Microsoft/VC++2015-2022Redist-x64/14.31.30818.0/Microsoft.VC++2015-2022Redist-x64.installer.yaml
+++ b/manifests/m/Microsoft/VC++2015-2022Redist-x64/14.31.30818.0/Microsoft.VC++2015-2022Redist-x64.installer.yaml
@@ -22,7 +22,7 @@ Installers:
   InstallerSwitches:
     Silent: /quiet /install /norestart
     SilentWithProgress: /passive /install /norestart
-  UpgradeBehavior: install
+  UpgradeBehavior: uninstallPrevious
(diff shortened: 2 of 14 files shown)


OfficialEsco:
I don't understand this issue, i've checked and regularly updated VC++2019 and 2022 via WinGet and it always removes the old version (pre 2015 does not uninstall old version tho)
But maybe its safer to do it this way since we do have complains about it all the time.. although i believe most are caused by the regex matching

felipecrs:
@OfficialEsco you are actually very right!

The issue is because winget is detecting my x86 VC++ 2015-2022 package as the x64 one during upgrade.

However, I cannot reproduce it in Sandbox.

That must be the reason why the old version never gets uninstalled/upgraded, because Winget is trying to install x64 variant while it should install the x86 one.

https://user-images.githubusercontent.com/29582865/165140994-f5b28aa3-84b9-49ae-b9c4-aaf94c7211da.mp4

OfficialEsco:
Oke good so i'm not that crazy yet, what version of winget-cli are you running on your host and in sandbox?
I've not experienced this issue in a long time, so i believe its fixed in winget-cli version 1.3, it will also be even more improved once it accepts all the ARP fields
https://github.com/microsoft/winget-pkgs/blob/d65be50b12a55401eb9195b69ceb88808dabbe9a/manifests/m/Microsoft/VC%2B%2B2015-2022Redist-x64/14.32.31326.0/Microsoft.VC%2B%2B2015-2022Redist-x64.installer.yaml#L24-L29

felipecrs:
Understood. I will check if I'm running Winget 1.3 in the affected environment, but, sadly, I installed the latest version of x86 variant by myself and now I cannot reproduce the issue anymore.

I tried to uninstall the latest version and install the outdated one, and now it detects as the proper package.

Anyway, if this issue comes back, I will let you with more meaningful information. Thanks a lot so far!

denelon:
@felipecrs 
It looks like you're not the only one with that experience:
https://github.com/microsoft/winget-cli/issues/1413#issuecomment-1097982141

felipecrs:
Hi @denelon, thanks for following up. But I don't believe it's the same situation. In my environment, I didn't have any broken or corrupted installation, nor I had any duplicated version installed. What I had was:

- VC++ 2015-2022 **x86** version 14.30.x.x (outdated)
- VC++ 2015-2022 **x64** version 14.32.x.x (up to date)

And still Winget detected the **x64** variant as outdated with the version from the **x86** one.

I unfortunately didn't show all the versions installed in my quick demo, but I remember that well. Anyway, if I ever get able to reproduce the issue again, I will be happy to provide more details.

Issue #54476, opened 2022-03-15
[Package Issue]: Lyx.Lyx Outdated


wigster:
### Please confirm these before moving forward

- [X] I have searched for my issue and not found a work-in-progress/duplicate/resolved issue.
- [X] I have not been informed if the issue is resolved in a preview version of the winget client.

### Category of the issue

No applicable update found.

### Brief description of your issue

The version of the LyX package is outdated by about 4 months.

Latest available on winget: 2.5.2
Latest available: 2.6.1

There was a new 64-bit installer introduced in version 2.6, so maybe this is the reason for the delay.

(https://ftp.lip6.fr/pub/lyx/bin/2.3.6/LyX-2361-Installer-3-x64.exe)

### Steps to reproduce

winget install lyx.lyx

### Actual behavior

installs 2.5.2

### Expected behavior

installs 2.6.1

### Environment

```shell
Windows Package Manager (Preview) v1.2.3411-preview
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.22000.556
Package: Microsoft.DesktopAppInstaller v1.17.3411.0
```

wigster:
I've noticed that this works fine in chocolatey, with no changes to the install script since 2.5. I'm setting up a new machine, trying to move from chocolatey to winget.

https://github.com/chtof/chocolatey-packages/blob/master/automatic/lyx/tools/chocolateyinstall.ps1

Edoardo66:
MiKTeX is a prerequisite

noamzaks:
> MiKTeX is a prerequisite

wdym?
MiKTeX is available, isn't it? https://github.com/microsoft/winget-pkgs/tree/master/manifests/m/MiKTeX/MiKTeX

cc @ItaiGani

Dragon1573:
> [!WARNING]
> 
> Manually verified. Even if I'm behind a proxy (proxy node in France), installer files can't be fetched from the server ...
> 
> Maybe they should be removed from the repository?

```text
17:19:25 D:\...\winget-pkgs  [master ≡] 1ms pwsh
$ komac list-versions LyX.LyX |
> ForEach-Object {
>   Write-Output ""
>   winget download --skip-dependencies -e --id LyX.LyX -v $_
> }

已找到 LyX [LyX.LyX] 版本 2.3.4.4
此应用程序由其所有者授权给你。
[...]

PR #62657, opened 2022-06-02
LyX.LyX version 2.3.6.1

Shengyis:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/62657)


ItzLevvie (changes requested):
Duplicate of https://github.com/microsoft/winget-pkgs/pull/7811.

We're going to have to wait until support for interactive-only installers are implemented.

Shengyis:
> 
Here is my test, no need the user interaction. 

no_user_i

And this is with user interaction.

with_user_i

The architecture is x64 instead of x86. I did not test the x86 one. But it seems the x64 one works well.

ItzLevvie (changes requested):
At least on a clean installation of Windows (automated tests happen on VMs), silent install seems broken for this app on WinGet: https://github.com/microsoft/winget-pkgs/issues/54476#issuecomment-1068110105

Shengyis:
oh...this is because lyx needs Texlive or Miktex installed in the system.....sad..

ItzLevvie (dismissed):

Trenly:
@itzlevvie - Just curious as I've seen a few times lately where you've suggested changes, and then later approved the PR with no changes??

If you have automation in place or a function to approve, perhaps a filter to exclude PR's with `Needs-Author-Feedback` would be important?

Shengyis:
@Trenly I don't know what to do next. Lyx needs to locate latex.exe during installation. So if there is Miktex/Texshop installed previously,  Lyx can be installed silently and with no error. But in VMs, the clean installation will report an error.

Trenly:
> @Trenly I don't know what to do next. Lyx needs to locate latex.exe during installation. So if there is Miktex/Texlive installed previously, Lyx can be installed silently and with no error. But in VMs, the clean installation will report an error.

This PR is blocked by microsoft/winget-cli#163 and the PR cannot be merged until the winget cli supports installing dependencies

Shengyis:
thx. good to know.

Shengyis:
but, the current version 2.3.5.2 should have the same problem. it will locate latex.exe when installing as well

Trenly:
> but, the current version 2.3.5.2 should have the same problem. it will locate latex.exe when installing as well

Actually, 2.3.5.2 doesn't require latex.exe to complete the installation. It may cause some issues if it isn't present, but it isn't a dependency until newer versions
image

FlorianPfaff:
The GUI interface has an option "Do not use LaTeX", maybe this could help to install without any dependency on miktex or some other LaTeX distribution. However, I do not know if it can be "checked" using command line commands.

Shengyis:
> 

it seems lyx has no available options to set "do not use latex" when installing using command line. lyx-xxx.exe /? just starts GUI installation.

FlorianPfaff:
Thanks for trying! Inow reported setting this via command line as a feature request now on the LyX bug tracker
[lyx.org/trac/ticket/12559](url)
Maybe they will consider it, which would help solve the problem (though it would be best if winget supported dependencies...).

ItzLevvie (approved):

ItzLevvie (changes requested):
You can set this pull request to DRAFT.

ItzLevvie:
@msftbot Interactive-Only-Installer

ItzLevvie:
@msftbot Blocking-Issue

ItzLevvie:
[Policy] Dependencies

PR #65274, opened 2022-07-06
Change VideoLAN.VLC version to match control panel.

JasonCook599:
- [ x] Have you checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/change?
- [ x] Have you [validated](https://github.com/microsoft/winget-pkgs/blob/master/AUTHORING_MANIFESTS.md#validation) your manifest locally with `winget validate --manifest <path>`? 
- [ x] Have you tested your manifest locally with `winget install --manifest <path>`?
- [ x] Does your manifest conform to the [1.1 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.1.0)?

-----
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/65274)

### DIFF
--- a/manifests/v/VideoLAN/VLC/3.0.17.0/VideoLAN.VLC.installer.yaml
+++ b/manifests/v/VideoLAN/VLC/3.0.17.0/VideoLAN.VLC.installer.yaml
@@ -2,7 +2,7 @@
 # yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.1.0.schema.json
 
 PackageIdentifier: VideoLAN.VLC
-PackageVersion: 3.0.17.4
+PackageVersion: 3.0.17.0
 Platform:
 - Windows.Desktop
 MinimumOSVersion: 10.0.0.0
--- a/manifests/v/VideoLAN/VLC/3.0.17.0/VideoLAN.VLC.locale.en-US.yaml
+++ b/manifests/v/VideoLAN/VLC/3.0.17.0/VideoLAN.VLC.locale.en-US.yaml
@@ -2,7 +2,7 @@
 # yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.1.0.schema.json
 
 PackageIdentifier: VideoLAN.VLC
-PackageVersion: 3.0.17.4
+PackageVersion: 3.0.17.0
 PackageLocale: en-US
 Publisher: VideoLAN
 PublisherUrl: https://www.videolan.org/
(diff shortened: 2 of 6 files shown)


ItzLevvie (changes requested):
Oops, I wrote about this issue here: https://github.com/microsoft/winget-pkgs/pull/61421#issuecomment-113299475

EXE has the correct version (3.17.4) so changing the PackageVersion to 3.17.0 won't cause any upgrade loops because the Control Panel version is larger than the PackageVersion.

MSI has the wrong version (3.17.0) so changing the PackageVersion to 3.17.0 won't cause any upgrade loops because the Control Panel version is the same as the PackageVersion.

Since a few months has already passed I think a lot of people have already upgraded to 3.17.4 to get the latest bugfixes so technically it's possible to merge this pull request but maybe I will wait for other mods to chime in.

Trenly:
I downloaded v3.0.17.0 from their website and installed it, but it appears their built-in Update Check functionality still detects it as latest without upgrading to v3.0.17.4; So, while winget users who upgrade regularly will have updated software, I would think that a majority of VLC users who are on the msi version are still vulnerable

image

I don't have a preference either way as to whether it gets changed or not, but I would say that the version is Technically 3.0.17.4 and not 3.0.17.0, so for accuracy it may be best to leave the upgrade loop in place until a new stable version is released.

ItzLevvie (changes requested):
https://github.com/microsoft/winget-pkgs/pull/65274#issuecomment-1176663791

JasonCook599:
> so for accuracy it may be best to leave the upgrade loop in place until a new stable version is released. 

I think we are in agreement that leaving the update loop in place is best. 

> Since the EXE installer has the correct version, could we remove the MSI installer for this version? The next time someone tries to update, the EXE version would be installed and the correct version listed. Are there any other side effects to removing the MSI installer?

What are your thoughts on removing the MSI installer? Would that have unintended consequences? Or should we consider the issue WON'T FIX and wait for a new release?

Trenly:
> Since the EXE installer has the correct version, could we remove the MSI installer for this version? The next time someone tries to update, the EXE version would be installed and the correct version listed. Are there any other side effects to removing the MSI installer?

This isn’t accurate. Winget will detect the install technologies are different and show an error that “the technologies are different, so users should uninstall and reinstall” (paraphrasing). The app would still show in upgrade since 3.0.17.4 is technically available - just as an EXE instead

> What are your thoughts on removing the MSI installer? Would that have unintended consequences? Or should we consider the issue WON'T FIX and wait for a new release?

You’re just trading one error for another; At this point I think we should wait for ARP version range matching in the winget client (1.5 release?) or a new release of VLC

Trenly (dismissed):

JasonCook599:
I have added a comment to the installer manifest about this issue. That should prevent someone else from trying to fix the issue again :)

ItzLevvie (approved):

ghost (approved):

PR #65410, opened 2022-07-07
Disable Duplicate(?) Task

### LINKED ISSUE #34408: [Bug]: Needs-Author-Feedback is removed on pipeline success
### Brief description of your issue

Msft-bot removes the “Needs-Author-Feedback” on Wingetbot’s PR’s when the pipeline was a success

### Steps to reproduce

Request Changes on Wingetbot’s PR while the pipeline is still running

### Expected behavior

The “Needs-Author-Feedback” tag is applied and persists

### Actual behavior

The label is removed

https://github.com/microsoft/winget-pkgs/pull/34404

### Environment

```shell
N/A
```

Trenly:
* Resolves #34408

I believe the task on L315-368 has incorrect triggers and is causing the Needs-Author-Feedback label to be removed as it is detecting @wingetbot adding the labels as an event, and thus triggering the action.

I also believe this is duplicate, as the task on L369-415 removes the Needs-Author-Feedback label and replaces it with Needs-Attention when the PR Author comments and the one on L416-462 does the same when responding to Review feedback.

@denelon - I'm not sure if this is the right way of disabling tasks, I just went based on what was in the config for other tasks; I'm  also not sure if it would be better to remove this task entirely; Obviously I can't test this myself, but I don't know if you do either?

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/65410)

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -364,7 +364,8 @@
           "project_card"
         ]
       },
-      "id": "xn8_sQh_33"
+      "id": "xn8_sQh_33",
+      "disabled": true
     },
     {
       "taskType": "trigger",


ItzLevvie (approved):

msftrubengu:
I don't believe it's a duplicate because PR events and PR comments events are treated differently, hence the two tasks. I do agree that task L315-368 is not needed. We should only remove the label for comments. 

We should probably also ignore L369-415 for wingetbot. It won't probably happen in practice, but if there's a PR from wingetbot with Needs-Author-Feedback and a non-related push happens because some other future automated update to the PR, the task will trigger and remove the label.

We've been talking internally about how to handle better wingetbot PRs that require human intervention and I think a similar task that is something like "if setting Needs-Author-Feedback and the author is wingetbot then assign it to one of the devs" would suffice.

msftrubengu (approved):

PR #67898, opened 2022-07-29
New package: GoLang.Go.1.18 version 1.18.4

russellbanks:
- [X] Does your manifest conform to the [1.1 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.1.0)?

Go has two stable channels that recieve updates currently: `1.17`, `1.18` and `1.19`, so the packages should reflect this.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/67898)

### DIFF
--- a/manifests/g/GoLang/Go/1/18/1.18.4/GoLang.Go.1.18.locale.en-US.yaml
+++ b/manifests/g/GoLang/Go/1/18/1.18.4/GoLang.Go.1.18.locale.en-US.yaml
@@ -1,0 +1,29 @@
+# Created with YamlCreate.ps1 v2.1.2 $debug=NVS1.7-2-5
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.1.0.schema.json
+
+PackageIdentifier: GoLang.Go.1.18
+PackageVersion: 1.18.4
+PackageLocale: en-US
+Publisher: GoLang
+PublisherUrl: https://go.dev/
+PublisherSupportUrl: https://go.dev/help/
+PrivacyUrl: https://policies.google.com/privacy
+Author: The Go Authors
+PackageName: Go 1.18
+PackageUrl: https://go.dev/
+License: BSD 3-Clause License
+LicenseUrl: https://github.com/golang/go/blob/master/LICENSE
+Copyright: Copyright (c) 2009 The Go Authors. All rights reserved.
+CopyrightUrl: https://go.dev/copyright
+ShortDescription: Go is an open source programming language that makes it easy to build simple, reliable, and efficient software.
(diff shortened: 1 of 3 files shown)


manifests/g/GoLang/Go/1/18/1.18.4/GoLang.Go.1.18.locale.en-US.yaml
```diff
@@ -0,0 +1,29 @@
...
+Author: The Go Authors
+PackageName: Go 1.18
+PackageUrl: https://go.dev/
+License: BSD 3-Clause License
+LicenseUrl: https://github.com/golang/go/blob/master/LICENSE
+Copyright: Copyright (c) 2009 The Go Authors. All rights reserved.
+CopyrightUrl: https://go.dev/copyright
+ShortDescription: Go is an open source programming language that makes it easy to build simple, reliable, and efficient software.
+Description: The Go programming language is an open source project to make programmers more productive. Go is expressive, concise, clean, and efficient. Its concurrency mechanisms make it easy to write programs that get the most out of multicore and networked machines, while its novel type system enables flexible and modular program construction. Go compiles quickly to machine code yet has the convenience of garbage collection and the power of run-time reflection. It's a fast, statically typed, compiled language that feels like a dynamically typed, interpreted language.
+Moniker: go-1-18
```
jedieaston, line 20: Should the latest version just have the moniker `go` or `golang` or something? I think `winget install golang` should do something. If someone else disagrees, I'm all ears.

jedieaston (changes requested):
Should the latest version just have the moniker `go` or `golang` or something? I think `winget install golang` should do something. If someone else disagrees, I'm all ears.

russellbanks:
> Should the latest version just have the moniker `go` or `golang` or something? I think `winget install golang` should do something. If someone else disagrees, I'm all ears.

Could do, but it would just mean having to change it every time there's a new channel so when Go 1.19 comes out, these packages would need updating to have the moniker of `go-1-18` again.

jedieaston:
@Trenly, @OfficialEsco, @ItzLevvie, thoughts?

OfficialEsco:
Is it that important atm? How far away is the feature to point to the latest version for a parent folder? How many people would prefer to just do `winget install golang` instead of just updating to the latest version via `winget upgrade`?

jedieaston:
Updating to the latest version won't be a thing if we split the PackageIdentifiers up by minor though, right? You'll get the latest patch but not the latest major/minor (which you could get around with `winget install golang`).

I'm guessing pointer packages are a while away, I haven't seen anything with a timeframe (although since it would probably be mostly work on the winget-pkgs service and not the CLI, the timeframe is not very public).

OfficialEsco:
From my point of view most people don't want to update to the latest major/minor version since they are working with a specific major/minor version, also we're causing more issues with installing side-by-side applications from the same package.

But if you want to keep an eye on the moniker for the versions so people can do `winget install golang`, go ahead, however i doubt  we'll get any complains about not getting the latest version by doing `winget install golang`, if we do then something that feature should be prioritized =)

russellbanks:
I think it's better to keep it as-is because this is how a lot of other packages are currently (`Groovy.2.5`, `Groovy.3`, `Groovy.4`, `RubyInstallerTeam.3.1`, etc) and as OfficialEsco has said, the feature to point to the latest version is coming

jedieaston (approved):

PR #68036, opened 2022-07-30
Audient.Sono version 4.2.1.0

hansmbakker:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/68036)

### DIFF
--- a/manifests/a/Audient/Sono/4.2.1.0/Audient.Sono.installer.yaml
+++ b/manifests/a/Audient/Sono/4.2.1.0/Audient.Sono.installer.yaml
@@ -1,0 +1,13 @@
+# Created using wingetcreate 1.0.4.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.1.0.schema.json
+
+PackageIdentifier: Audient.Sono
+PackageVersion: 4.2.1.0
+Installers:
+- Architecture: neutral
+  InstallerType: nullsoft
+  InstallerUrl: https://d9w4fhj63j193.cloudfront.net/2021/Sono/1.%20Drivers/Sono-v4.2.1a.exe
+  InstallerSha256: 427D0EB1174EF765B40762EF1FD6802E5D29E4CE409812FC8F831B95834E6EBA
+ManifestType: installer
+ManifestVersion: 1.1.0
+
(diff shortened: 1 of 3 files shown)


hansmbakker:
Audient calls the version `4.2.1a` instead of `4.2.1.0` but I believe winget expects numbers for the version?

hansmbakker:
I added the requirement for interactive installation in my installer manifest PR but

```
winget install --manifest .\manifests\a\Audient\Sono\4.2.1.0\
```
does not trigger an interactive installation and instead fails with exit code 2 - is that related to https://github.com/microsoft/winget-cli/issues/910?

The executable seems not to respond to command line parameters so I do not know how to make it unattended.

hansmbakker:
The link comes from https://audient.com/products/audio-interfaces/sono/downloads/. It is similar to https://github.com/microsoft/winget-pkgs/pull/60122.

ItzLevvie (approved):

hansmbakker:
@ItzLevvie thank you for the review!

Could you please help me with my question about the interactive mode required to make the installer succeed?

I put `interactive` as the only supported mode but winget still tries to install it silently unless I specify `-i`.

In the case it tries to install silently, it will install the drivers without the accompanying tools, and it will exit with exitcode 2 (I did not see additional info).

ItzLevvie:
`InstallModes:` currently doesn't do anything in WinGet.

When I installed this application on a clean install of Windows 11 it didn't exit with %ERRORLEVEL% of 2 and installed fine.

image

hansmbakker:
I found out that likely it crashed because of `Controlled folder access` - Windows blocked the installer from creating a folder in the `Documents` folder:
image

After allowing that action, the installation succeeded without an errorcode. ✅

hansmbakker:
I had some questions left (sorry - I'm new to creating these manifests)

As for the output of your terminal - I see it recognizes the three components it installed:
1. the driver
1. the control app
1. another app from Two Notes (Torpedo Remote)

I do not get that output on my console - do you have a newer version of winget maybe? Or do you use a parameter I don't know about?
Is there a way to tell winget how to deal with those components? Can I give those components a specific `Package Identifier`?

The reason I'm asking this is because I would like to package the 3rd component (Torpedo Remote) separately as updated versions are offered separately from https://www.two-notes.com/en/torpedo-series/torpedo-remote/.

Trenly:
> I do not get that output on my console - do you have a newer version of winget maybe? Or do you use a parameter I don't know about?

Levvie uses a custom script for testing pull requests. The best equivalent which is available right now is the [SandboxTest.ps1 script](https://github.com/microsoft/winget-pkgs/blob/master/doc/tools/SandboxTest.md)

> Is there a way to tell winget how to deal with those components? Can I give those components a specific Package Identifier?

The best way is to create a manifest for those additional packages so that they can be matched and upgraded as their own package

hansmbakker:
@Trenly thanks!

> The best way is to create a manifest for those additional packages so that they can be matched and upgraded as their own package
Created a manifest for Torpedo Remote at https://github.com/microsoft/winget-pkgs/pull/68221 and updated it using the manifest.

It installed correctly but `winget list` still shows the incorrect Package Name and Id (it lists Package Id as `Remote5`). Maybe that will be fixed when the manifest is merged in the repo and the winget index is updated?

PR #68356, opened 2022-08-01
Kiwix.KiwixJS version 2.1.0.0

Jaifroid:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/68356)

### DIFF
--- a/manifests/k/Kiwix/KiwixJS/2.1.0.0/Kiwix.KiwixJS.installer.yaml
+++ b/manifests/k/Kiwix/KiwixJS/2.1.0.0/Kiwix.KiwixJS.installer.yaml
@@ -1,0 +1,18 @@
+# Created using wingetcreate 1.0.4.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.1.0.schema.json
+
+PackageIdentifier: Kiwix.KiwixJS
+PackageVersion: 2.1.0.0
+Installers:
+- Platform:
+  - Windows.Universal
+  MinimumOSVersion: 10.0.10240.0
+  Architecture: neutral
+  InstallerType: appx
+  InstallerUrl: https://github.com/kiwix/kiwix-js-windows/releases/download/v2.1.0/KiwixWebApp_2.1.0.0_AnyCPU.appxbundle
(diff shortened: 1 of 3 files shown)


Jaifroid:
@yao-msft This updated HTML/JS UWP package is running into the same mysterious validation error as we had last month with the three packages Kiwix.KiwixJS, Kiwix.Wikivoyage and Kiwix.WikiMed. I have tested it locally, and it installs fine (see screenshot below) and certificate is valid.

Do I need to request a manual validation each time I upload a new package? This would be three times a month at least. I thought the validation waiver would carry over. Is there any other solution? Many thanks.

image

ItzLevvie (approved):

Jaifroid:
@denelon , @AmelBawa-msft This UWP HTML/JS appxbundle is running into the same validation issue as happened last month, despite the fact that you added a waiver last month. As a reminder, validating this type of msix/appxbundle app used to work fine with the Azure validation scripts, but it stopped validating just over a month ago despite the fact that winget has no problem installing the bundle, and the certificate checks out fine.

Is there any other solution than my having to request a manual validation every time I update these apps? There are three: Kiwix.KiwixJS, Kiwix.WikiMed, Kiwix.Wikivoyage. I update each one on a monthly cycle.

ryfu-msft:
@Jaifroid, I've waived this error again. I do expect this issue to be fixed in the near future as we are planning to update the validation to start handling portable packages soon. Once that update happens, I believe this package should no longer have issues. I'll keep an eye out if this problem occurs again.

Jaifroid:
@ryfu-msft Thank you very much. It's not supposed to be a portable package, it's fully installable as a UWP app, and goes in the UWP hidden installation folder when installed.

Be that as it may, the issue we're facing now is that the current versions of winget (latest 1.3 stable and the 1.4 preview) are both unable to deal with a 1.3GB installation file. I suspect a timeout because it never gets past the hash check and errors out with
```
An unexpected error occurred while executing the command:
0x80072eff : unknown error
```
This is nothing to do with package type, it's due to the size of the package. A timeout seems to abandon the hash check before the package can be downloaded. It's the only explanation I can come up with. See https://github.com/microsoft/winget-cli/issues/1987 .

PR #68461, opened 2022-08-02
New version: dail8859.NotepadNext version 0.5.4

dail8859:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/68461)

### DIFF
--- a/manifests/d/dail8859/NotepadNext/0.5.4/dail8859.NotepadNext.installer.yaml
+++ b/manifests/d/dail8859/NotepadNext/0.5.4/dail8859.NotepadNext.installer.yaml
@@ -1,0 +1,23 @@
+# Created with YamlCreate.ps1 v2.0.7 using InputObject 🤖 $debug=QUSU-7-2-5
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.1.0.schema.json
+
+PackageIdentifier: dail8859.NotepadNext
+PackageVersion: 0.5.4
+MinimumOSVersion: 10.0.0.0
+InstallerType: nullsoft
+UpgradeBehavior: install
+Installers:
+- Architecture: x64
+  Scope: user
+  InstallerUrl: https://github.com/dail8859/NotepadNext/releases/download/v0.5.4/NotepadNext-v0.5.4-Installer.exe
+  InstallerSha256: A45822E2777D4A6848538A0D3102784578FD95815D34F45B3A9DE139C04EB2A6
+  InstallerSwitches:
+    Custom: /CURRENTUSER
+- Architecture: x64
+  Scope: machine
+  InstallerUrl: https://github.com/dail8859/NotepadNext/releases/download/v0.5.4/NotepadNext-v0.5.4-Installer.exe
+  InstallerSha256: A45822E2777D4A6848538A0D3102784578FD95815D34F45B3A9DE139C04EB2A6
+  InstallerSwitches:
+    Custom: /ALLUSERS
+ManifestType: installer
+ManifestVersion: 1.1.0
--- a/manifests/d/dail8859/NotepadNext/0.5.4/dail8859.NotepadNext.installer.yaml
+++ b/manifests/d/dail8859/NotepadNext/0.5.4/dail8859.NotepadNext.installer.yaml
@@ -22,6 +22,6 @@ Installers:
   InstallerSwitches:
     Custom: /ALLUSERS
   AppsAndFeaturesEntries: 
-    - DisplayName: Notepad Next v0.5.4
+    - DisplayName: Notepad Next
 ManifestType: installer
 ManifestVersion: 1.1.0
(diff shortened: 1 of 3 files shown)


manifests/d/dail8859/NotepadNext/0.5.4/dail8859.NotepadNext.installer.yaml
```diff
@@ -0,0 +1,23 @@
...
+  InstallerUrl: https://github.com/dail8859/NotepadNext/releases/download/v0.5.4/NotepadNext-v0.5.4-Installer.exe
+  InstallerSha256: A45822E2777D4A6848538A0D3102784578FD95815D34F45B3A9DE139C04EB2A6
+  InstallerSwitches:
+    Custom: /CURRENTUSER
+- Architecture: x64
+  Scope: machine
+  InstallerUrl: https://github.com/dail8859/NotepadNext/releases/download/v0.5.4/NotepadNext-v0.5.4-Installer.exe
+  InstallerSha256: A45822E2777D4A6848538A0D3102784578FD95815D34F45B3A9DE139C04EB2A6
+  InstallerSwitches:
+    Custom: /ALLUSERS
```
Trenly, line 21: Display Names are recommended here to help with scope
```suggestion
Installers:
- Architecture: x64
  Scope: user
  InstallerUrl: https://github.com/dail8859/NotepadNext/releases/download/v0.5.4/NotepadNext-v0.5.4-Installer.exe
  InstallerSha256: A45822E2777D4A6848538A0D3102784578FD95815D34F45B3A9DE139C04EB2A6
  InstallerSwitches:
    Custom: /CURRENTUSER
  AppsAndFeaturesEntries: 
    - DisplayName: Notepad Next v0.5.4 (current user)
- Architecture: x64
  Scope: machine
  InstallerUrl: https://github.com/dail8859/NotepadNext/releases/download/v0.5.4/NotepadNext-v0.5.4-Installer.exe
  InstallerSha256: A45822E2777D4A6848538A0D3102784578FD95815D34F45B3A9DE139C04EB2A6
  InstallerSwitches:
    Custom: /ALLUSERS
  AppsAndFeaturesEntries: 
    - DisplayName: Notepad Next v0.5.4
```

Trenly (changes requested):

manifests/d/dail8859/NotepadNext/0.5.4/dail8859.NotepadNext.installer.yaml
dail8859: @sitiom Thoughts on this suggestion? Will the WinGet Releaser workflow update the display name appropriately in this case?

dail8859: Or would it be better to not include the version number in the DisplayName, e.g. `Notepad Next (current user)` and `Notepad Next`? Not sure what the best-practice.

Trenly: Vedant's action won't update the AppsAndFeaturesEntries, so leaving the version out of the DisplayName would be helpful, especially since the Control Panel will show the version a few columns over anyways.

dail8859: @Trenly Agreed.

Trenly:
Unfortunately, that isn't quite what I meant. The `DisplayName` field, if included, must match *exactly* what is written to control panel. When I said that it would be best to leave the version out of the DisplayName, I meant that the build script which creates the exe file should be updated to not include the version in the DisplayName registry key. 

Sorry for not being clear

dail8859:
@Trenly Oooooooh. I see now!

So the root issue is that since this support different installation scopes (current user vs. all users) that it needs some way to distinguish between the two? I can easily take the version string out of the display name, but I would need to add some extra installer logic to append something to it (e.g. `current user`) based on the install version.

I see a few possible options until the installer can be adjusted (since a new release would need created anyway):

1. Accept the original PR without the `AppsAndFeaturesEntries` block (since that's how it existed until now)
2. Just keep the machine scope for now.
3. Skip publishing this version of Notepad Next.

dail8859:
Actually, the changes look very easy on the NSIS side of things. But again, it would require publishing a new release.

sitiom:
> 1. Accept the original PR without the `AppsAndFeaturesEntries` block (since that's how it existed until now)

I would recommend this for now.

ItzLevvie (changes requested):
https://github.com/microsoft/winget-pkgs/pull/68461#issuecomment-1203374555

dail8859:
As suggested by @sitiom and @ItzLevvie, `AppsAndFeaturesEntries` block has been removed. This should be equivalent to the original PR. The `DisplayName` will be updated for the next release of Notepad Next.

ItzLevvie (approved):

dail8859:
Display Name is fixed in the next version:

https://github.com/dail8859/NotepadNext/commit/70a96eff294ca8ffa3c311da257ffb5497d9ed07

This will either display `Notepad Next` or `Notepad Next (current user)` in the control panel.

PR #69382, opened 2022-08-11
Add Machine Scope for Eloston.Ungoogled-Chromium

### LINKED ISSUE #69381: [New Feature]: Add per-machine installer for Eloston Ungoogled-Chromium (Regression)
### Description of the new feature/enhancement

Please reopen #39848.  
This was fixed for eloston ungoogled-chromium 96.0.4664.110, but has regressed in the current version.

Fixed version:
https://github.com/microsoft/winget-pkgs/blob/master/manifests/e/eloston/ungoogled-chromium/96.0.4664.110/eloston.ungoogled-chromium.installer.yaml

It looks like this change didn't make it into the next version.  Is there a template somewhere that needs to be updated?

### Proposed technical implementation details (optional)

_No response_

Trenly:
* Resolves #69381

 @Nifury - I checked the version before this and it did contain scopes for user and machine. If possible, when updating this package in the future, be sure to keep both the user scope and machine scope as it causes issues for people upgrading if both are not present in new versions. (Ref #68863)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/69382)

### DIFF
--- a/manifests/e/eloston/ungoogled-chromium/104.0.5112.81/eloston.ungoogled-chromium.installer.yaml
+++ b/manifests/e/eloston/ungoogled-chromium/104.0.5112.81/eloston.ungoogled-chromium.installer.yaml
@@ -6,10 +6,6 @@ PackageVersion: 104.0.5112.81
 InstallerLocale: en-US
 MinimumOSVersion: 10.0.0.0
 InstallerType: exe
-Scope: user
-InstallerSwitches:
-  Silent: /silent /install
-  SilentWithProgress: /silent /install
 UpgradeBehavior: install
 Protocols:
 - http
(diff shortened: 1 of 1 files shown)


ItzLevvie (approved):

Nifury:
Oh sorry I didn't know the machine scope was actually requested by a user. I removed it only because winget-create doesn't support multiple scopes for the same installer. I'll try other ways to update this package automatically.

Trenly:
@denelon - Do you know if there are any open issues over at create regarding same scope for multiple installers?

Nifury:
Ah this issue occurs when I update the manifest. https://github.com/microsoft/winget-create/issues/291#issuecomment-1198459372

Trenly:
Ohhh, I see. Makes sense! Thanks for your help looking into it!!

denelon:
@ryfu-msft can you check wingetcreate for this scenario?

PR #69431, opened 2022-08-11
Update: Microsoft.WingetCreate version 1.1.2.0

russellbanks:
- [X] Does your manifest conform to the [1.1 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.1.0)?

This adds the portable wingetcreate exe

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/69431)

### DIFF
--- a/manifests/m/Microsoft/WingetCreate/1.1.2.0/Microsoft.WingetCreate.locale.en-US.yaml
+++ b/manifests/m/Microsoft/WingetCreate/1.1.2.0/Microsoft.WingetCreate.locale.en-US.yaml
@@ -1,13 +1,13 @@
-# Created using wingetcreate 1.0.4.0
-# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.1.0.schema.json
+# Created with YamlCreate.ps1 v2.1.3 $debug=NVS1.7-2-5
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.2.0.schema.json
 
 PackageIdentifier: Microsoft.WingetCreate
 PackageVersion: 1.1.2.0
 PackageLocale: en-US
 Publisher: Microsoft Corporation
-PublisherUrl: https://www.microsoft.com/en-US/
+PublisherUrl: https://www.microsoft.com
 PublisherSupportUrl: https://github.com/microsoft/winget-create/issues
-PrivacyUrl: https://privacy.microsoft.com/en-US/privacystatement
+PrivacyUrl: https://github.com/microsoft/winget-create/blob/main/PRIVACY.md
 Author: https://github.com/microsoft/winget-create/graphs/contributors
 PackageName: Windows Package Manager Manifest Creator
 PackageUrl: https://github.com/microsoft/winget-create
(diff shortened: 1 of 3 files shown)


ItzLevvie (dismissed):

manifests/m/Microsoft/WingetCreate/1.1.2.0/Microsoft.WingetCreate.locale.en-US.yaml
```diff
@@ -1,13 +1,13 @@
...
 
 PackageIdentifier: Microsoft.WingetCreate
 PackageVersion: 1.1.2.0
 PackageLocale: en-US
 Publisher: Microsoft Corporation
-PublisherUrl: https://www.microsoft.com/en-US/
+PublisherUrl: https://www.microsoft.com
 PublisherSupportUrl: https://github.com/microsoft/winget-create/issues
-PrivacyUrl: https://privacy.microsoft.com/en-US/privacystatement
+PrivacyUrl: https://github.com/microsoft/winget-create/blob/main/PRIVACY.md
```
Trenly, line 10:
```suggestion
PrivacyUrl: https://privacy.microsoft.com/en-US/privacystatement
```
The original URL is correct, as the GitHub one just tells users to go to the microsoft.com one

Trenly (changes requested):

Trenly:
As I've been thinking. . . @ryfu-msft / @denelon - Given that this package is managed by the winget team, do you want the portable exe added here?

denelon:
I've been thinking about this one since we have the MSIX version and the portable versions with and without dependencies to support CI/CD flows. We don't have anything like `winget install <package> --installer-type [MSIX, MSI, .exe, portable, zip]`. This is likely to set a precedent so we should be intentional.

denelon:
We should look at:
* https://github.com/microsoft/winget-cli/issues/1166

It might make sense to have Microsoft.WingetCreate.Portable and break that down into the versions with and without dependencies. The manifest for the .exe without dependencies should declare them so if they are on the system they don't get downloaded.

russellbanks:
> It might make sense to have Microsoft.WingetCreate.Portable and break that down into the versions with and without dependencies. The manifest for the .exe without dependencies should declare them so if they are on the system they don't get downloaded.

Haven't the self-contained portable .exe's been removed since v1.1.2.0:
> Note: Self-contained exes have been removed since WS2022 now supports AppExecutionAliases. Check out the [README](https://github.com/microsoft/winget-create#using-the-msixbundle) for instructions on how to use the msixbundle in your CI/CD pipeline.

denelon:
I think I'm OK having both the MSIX and the portable .exe in the same manifest. It's going to take feature work to allow someone to install the portable though. It might be best in the short term to have a Microsoft.WingetCreate.Portable manifest for now so people can use winget to install the portable version.

denelon:
@ryfu-msft let me know if you have a different opinion 😊

ItzLevvie (approved):

PR #72164, opened 2022-08-26
First addition for SK Data AS into Winget, ORM Core. Our ORM programs combined into one SUIT. 

skdataas:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/72164)


skdataas (commented):
This Zumoshi company is not SKData AS. someone it was added when I cloned the master. is it possible to only merge SKDATA AS or otherwise cancel the pull request? As I already updated to a version that doesn't contain a difference like this.

PR #72174, opened 2022-08-26
Here is the new version,  it should only contain the SKDATA AS and ORM manifests. 

skdataas:
Please only check the manifist in ./manifests/s/skdataas/ormcore/202208261/

Thank you. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/72174)


Trenly:
Only one package can be submitted in a PR. You will need to use two different PR's - one for AS and one for ORM

PR #72874, opened 2022-08-27
Normalize Crazybump.Crazybump 1.2.2

Trenly:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/72874)


ItzLevvie (approved):

OfficialEsco:
VirtualBoxVM_QZQc1qnLxF

Madhusudhan-MSFT:
This seems to install fine but launching it would result in a failure as follows:

image


this looks .net dependency it looks for is missing.

Madhusudhan-MSFT:
Given this is just only normalization change and this package was submitted long ago. Our assumption is this package likely used by people already have dependencies. So, applying waiver for this one.

Trenly:
> Given this is just only normalization change and this package was submitted long ago. Our assumption is this package likely used by people already have dependencies. So, applying waiver for this one.

Thanks for the context!

PR #64058, opened 2022-06-21
New version: Baidu.BaiduTranslate version 1.5.4

vedantmgoyal9:
### Result: Installation Successful
|             | Manifest                 | Add/Remove Programs                   |
| ----------- | ------------------------ | ------------------------------------- |
| Name        | 百度翻译 |     |
| Version     | 1.5.4     |  |
| Publisher   | Baidu Translate   |       |
| ProductCode | b34156d3-fe5f-5fd2-a095-3f35d6dfc923, b34156d3-fe5f-5fd2-a095-3f35d6dfc923, b34156d3-fe5f-5fd2-a095-3f35d6dfc923, b34156d3-fe5f-5fd2-a095-3f35d6dfc923          |     |
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/64058)

PR #62517, opened 2022-06-02
New version: runlet.runlet version 1.0.8 [FP-D]

vedantmgoyal9:
### Result: Installation Failed
|             | Manifest                 | Add/Remove Programs                   |
| ----------- | ------------------------ | ------------------------------------- |
| Name        | Runlet | KDevelop    |
| Version     | 1.0.8     | 5.6.2 |
| Publisher   | Runlet   | KDE e.V.      |
| ProductCode |           | KDevelop    |
#### Auto-updated by [vedantmgoyal2009/winget-pkgs-automation](https://bittu.eu.org/docs/wpa-intro) in workflow run [3338](<https://github.com/vedantmgoyal2009/vedantmgoyal2009/actions/runs/3164927831>)

PR #76500, opened 2022-08-31
Don't remove Blocking Issue on changes pushed

vedantmgoyal9:
PRs such as:

* https://github.com/microsoft/winget-pkgs/pull/64058
* https://github.com/microsoft/winget-pkgs/pull/62517

have blocking issues and they're removed on every synchronize event. Its hard to mention @denelon, again and again, to add blocking issues, and sometimes to reopen if they're closed with No-Recent-Activity.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/76500)

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -3262,12 +3262,6 @@
           {
             "name": "removeLabel",
             "parameters": {
-              "label": "Blocking-Issue"
-            }
-          },
-          {
-            "name": "removeLabel",
-            "parameters": {
               "label": "Needs-SmartScreen-Investigation"
             }
           },


denelon:
@Trenly does this look like a good change to you? What's the blast radius you perceive?

Trenly:
> @Trenly does this look like a good change to you? What's the blast radius you perceive?

Any PR labeled blocking-issue will have to have it manually removed when the blocking issue is resolved.

Honestly, I think that keeping the behavior as it is probably a good thing. There shouldn’t need to be changes pushed to branches with "Blocking-Issue" anyways, so it would make sense that when changes are pushed, it should be re-evaluated as to whether or not it is still blocking

vedantmgoyal2009:
> Any PR labeled blocking-issue will have to have it manually removed when the blocking issue is resolved.

@denelon

denelon:
We're going to leave the behavior as it is currently implemented. I believe it to be the better choice since the "Blocking-Issue" label may be put in place intentionally for a manual decision, and automatically removing it forces someone to re-evaluate it again. We don't want these to get missed when we have a high PR volume.

PR #78795, opened 2022-09-06
Ensure YamlCreate renormalizes on Commit

Trenly:
@denelon @jedieaston @OfficialEsco 

This PR makes it so YamlCreate always renormalizes the files it creates before committing them - which should always result in CRLF endings

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/78795)

### DIFF
--- a/Tools/YamlCreate.ps1
+++ b/Tools/YamlCreate.ps1
@@ -91,7 +91,7 @@ if (Get-Command 'git.exe' -ErrorAction SilentlyContinue) {
                 DefaultString = ''
             }
             switch (Invoke-KeypressMenu -Prompt $_menu['Prompt'] -Entries $_menu['Entries'] -DefaultString $_menu['DefaultString'] -HelpText $_menu['HelpText']) {
-                'Y' { 
+                'Y' {
                     Write-Host
                     try {
                         winget upgrade --id Git.Git --exact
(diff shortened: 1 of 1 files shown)


ItzLevvie (approved):

russellbanks:
Couldn't this mean that the debug information would technically be incorrect? If a user were to create a manifest in an environment that produces LF line endings, the debug information would say LF but the line ending would be CRLF because it's renormalized on commit, at leasts from what I imagine would happen.

Trenly:
> Couldn't this mean that the debug information would technically be incorrect? If a user were to create a manifest in an environment that produces LF line endings, the debug information would say LF but the line ending would be CRLF because it's renormalized on commit, at leasts from what I imagine would happen.

Depends on your perspective. I look at it as the debug information tells the default line ending of the system it was created on, not necessarily the line endings of the file

jedieaston (approved):

russellbanks (approved):

Trenly:
Closing this as there is a bug in the logic of using the VerisonManifestFolder which will cause autoupgrades to not function, as well as a bug on WSL and Linux systems where `--renormalize` will cause an error when attempting to renormalize files that have been deleted

PR #79404, opened 2022-09-12
Workflow to check line endings on PRs

vedantmgoyal9:
@denelon this workflow checks line endings of manifests on pull requests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/79404)

### DIFF
--- a/.github/workflows/lineEndings.yml
+++ b/.github/workflows/lineEndings.yml
@@ -1,0 +1,75 @@
+name: Line Endings
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+    types:
+      - opened
+      - synchronize
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/github-script@v6
+        with:
+          script: |
+            // Get changed files
+            switch (context.eventName) {
+              case 'push':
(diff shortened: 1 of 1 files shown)


OfficialEsco (changes requested):
This will stop automated PRs from passing the pipeline and is not the correct way to deal with the issue.

Trenly (changes requested):
Because you're making Github calls, you're also using the default token, which in this case is the Application Token. That is still subject to rate limiting, even though the limits are higher. 

While a workflow could potentially work for this, I would strongly discourage it at this time, as the root causes of wingetbot and wingetcreate using LF endings have not been resolved, and it hasn't been verified that YamlCreate always uses CRLF. It's more important to fix the root causes, as implementing this would just cause PR's to be blocked. 

Ultimately, having LF line endings doesn't break anything, so there isn't any value in breaking PR's

vedantmgoyal2009:
> This will stop automated PRs from passing the pipeline and is not the correct way to deal with the issue.

Instead of requesting changes on PR, will a comment on the PR be fine? If the author can fix it, well and good, and if not, the PR would still be merged.

Another thing we can do is auto-renormalize if LF line endings are detected. @denelon what are your thoughts?

Trenly:
Close with reason: Not Planned for implementation;

PR #80182, opened 2022-09-20
New version: tibbo.ioninja version 5.2.0

vovkos:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/80182)

### DIFF
--- a/manifests/t/tibbo/ioninja/5.2.0/tibbo.ioninja.installer.yaml
+++ b/manifests/t/tibbo/ioninja/5.2.0/tibbo.ioninja.installer.yaml
@@ -1,0 +1,24 @@
+# Created using YamlCreate.ps1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.0.0.schema.json
+
+PackageIdentifier: tibbo.ioninja
+PackageVersion: 5.2.0
+Platform:
+- Windows.Desktop
+MinimumOSVersion: 10.0.0.0
+InstallerType: wix
+Scope: machine
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+Installers:
+- InstallerLocale: en-US
+  Architecture: x64
+  InstallerType: wix
+  InstallerUrl: https://ioninja.com/downloads/winget/ioninja-5.2.0-windows-amd64-core.msi
+  InstallerSha256: BA827AD91158ED6F2232E2033DF0F0F31993518B554D03362516174C35AF6883
+  ProductCode: '{BBE58E28-F44B-4891-8A74-CC9147FB25B2}'
+  UpgradeBehavior: install
+ManifestType: installer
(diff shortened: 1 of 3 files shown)


vovkos:
I need some help in resolving the hash validation failure, hope you can push me in the right direction.

What I do:

1. Download the installer from:

```
https://tibbo.com/downloads/archive/ioninja/ioninja-5.2.0/ioninja-5.2.0-windows-amd64-core.msi
```

This is a direct file link, no redirections whatsoever.

2. Verify hash:

```
winget hash -f ioninja-5.2.0-windows-amd64-core.msi
Sha256: ba827ad91158ed6f2232e2033df0f0f31993518b554d03362516174c35af6883 
```

i.e., the same as specified in `tibbo.ioninja.installer.yaml`

Yet, the pull request is rejected with:

```
I am sorry to report that the Sha256 Hash does not match the installer. This may be caused by using a vanity URL rather than a URL directly to the binary.

Please check the HASH and submit again. Sorry for the inconvenience.
```

3. I tried to run `winget` locally, but it also fails the hash validation: 

```
Installer hash does not match; this cannot be overridden when running as admin 
```

The hash is correct, so I'm quite confused; could you help please?

Thank you!

vovkos:
Just in case, I'm running the latest stable version of `winget`:

```
winget --version
v1.3.2091    
```

ItzLevvie:
No idea. It seems to be an issue that only happens on Delivery Optimization.

Delivery Optimization (DO):
image

WinINet:
image

The Download URL used to work with Delivery Optimization because this package used to exist before it was removed but it no longer works now.

Maybe the publisher changed something server-side that prevents it from downloading on Delivery Optimization or it's a Delivery Optimization bug.

cc @denelon

JohnMcPMS:
Thanks @ItzLevvie , that binary comparison shows the issue I was about to propose as the problem.

DO opens multiple connections to download the file.  They download 1 MB chunks using range requests.  If the URL is a redirection, and that points at different locations that actually have different files, then you will get 1 MB chunks of different files stitched together.  DO will never work for a redirection like this, because even without the hash check, you have a frankenfile that likely won't run.

WinINet is our own, very simple download code that does one connection and downloads whatever is behind that.  I would expect it to work most of the time, until you get redirected to the one different file host, then it will fail too (but you might have a functional file).

I agree that this URL does not appear to have a redirect, but I have no idea if the server is itself a pass through to different files.  We can't know how the server is working, just that the symptom very much looks like the known pitfall.

vovkos:
Thank you for your research. Not quite sure what's wrong with the original download URL; I will try to figure that out with our sysadmin. Meanwhile, I moved the `.msi` file to another server and it passes the hash check now.

ItzLevvie (approved):

PR #80300, opened 2022-09-21
[Zip][Portable] New package: WindowsPostInstallWizard.UniversalSilentSwitchFinder version 1.5.0.0

Trenly:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/80300)

### DIFF
--- a/manifests/w/WindowsPostInstallWizard/UniversalSilentSwitchFinder/1.5.0.0/WindowsPostInstallWizard.UniversalSilentSwitchFinder.installer.yaml
+++ b/manifests/w/WindowsPostInstallWizard/UniversalSilentSwitchFinder/1.5.0.0/WindowsPostInstallWizard.UniversalSilentSwitchFinder.installer.yaml
@@ -1,0 +1,28 @@
+# Created with YamlCreate.ps1 v2.2.0 $debug=NVS1.CRLF.5-1-19041-1682.Win32NT
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.4.0.schema.json
+
+PackageIdentifier: WindowsPostInstallWizard.UniversalSilentSwitchFinder
+PackageVersion: 1.5.0.0
+InstallerType: zip
+FileExtensions:
+- exe
+ReleaseDate: 2011-02-04
+Installers:
+- Architecture: x64
+  NestedInstallerType: portable
+  NestedInstallerFiles:
+  - RelativeFilePath: ./ussf.exe
+    PortableCommandAlias: ussf
+  # InstallerUrl: http://www.capstanservices.com/s/ussf-5jre.zip
+  InstallerUrl: https://static1.squarespace.com/static/555a4c09e4b01adc78467a1c/t/5ac5401eaa4a99de7a8b9155/1522876447578/ussf.zip
+  InstallerSha256: 2EF53A3238061F95AC4B72CDBFA9AB67FD5923AF746AE148B8FC4A02902F646F
+- Architecture: x86
+  NestedInstallerType: portable
+  NestedInstallerFiles:
+  - RelativeFilePath: ./ussf.exe
+    PortableCommandAlias: ussf
+  # InstallerUrl: http://www.capstanservices.com/s/ussf-5jre.zip
+  InstallerUrl: https://static1.squarespace.com/static/555a4c09e4b01adc78467a1c/t/5ac5401eaa4a99de7a8b9155/1522876447578/ussf.zip
+  InstallerSha256: 2EF53A3238061F95AC4B72CDBFA9AB67FD5923AF746AE148B8FC4A02902F646F
+ManifestType: installer
+ManifestVersion: 1.4.0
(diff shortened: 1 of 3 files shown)


vedantmgoyal9 (changes requested):
https://github.com/microsoft/winget-pkgs/discussions/31163#discussioncomment-1548738

@jedieaston

Trenly:
> [#31163 (reply in thread)](https://github.com/microsoft/winget-pkgs/discussions/31163#discussioncomment-1548738)
> 
> @jedieaston

Isn’t that the entire purpose of the pipeline and the malware scans though, is to catch when potentially malicious software links are used?

USSF has been around since 2011, and I know that myself and others who use winget are familiar enough with the package to know that it is safe. And, if it ever were to be compromised, that is why there is hash validations

vedantmgoyal2009:
USSF is great, but SoftPedia isn't trustable.

Trenly:
> USSF is great, but SoftPedia isn't trustable.

I’m not saying that SoftPedia is trustable as a whole. I'm saying that in certain cases where we know the package is safe, there are checks in place that would prevent bad updates. In fact, no publiser website is truly "trustable", not even Microsoft's. All websites are vulnerable to compromise. Certainly SoftPedia being open to all increases the chances of the site hosting malware, but that doesn’t mean all packages are malware.

We know USSF is safe, and if the hash were ever to change then it would be blocked from installing without using the force parameter.

ItzLevvie (dismissed):

JohnMcPMS:
I can say that the publisher of that on Softpedia, wpiw.net, is blocked by MSIT as a malware site.  Doesn't make me feel a warm and fuzzy inside.

Calling for an adult.  @denelon !

Trenly:
> I can say that the publisher of that on Softpedia, wpiw.net, is blocked by MSIT as a malware site. Doesn't make me feel a warm and fuzzy inside.
> 
> Calling for an adult. @denelon !

I will also add on to that statement that wpiw.net as a website is no longer a legitimate website (it likely was when the software came out, but alas, no longer). However, the software itself, being hosted on SoftPedia is used by community members and is known to be safe. 

Perhaps changing the publisher name in the metadata to be WPIW may also help keep this package's users safe (if accepted)

denelon:
Are there any other "main sites" besides WIPW and [SoftPedia](https://www.softpedia.com/get/System/Launchers-Shutdown-Tools/Universal-Silent-Switch-Finder.shtml)? I'm guessing our user agent or network is blocked by SoftPedia. I've taken a look at the URLs failing in validation and I don't have any problems with them. I also didn't see any licensing concerns, so I'm OK with a domain waiver on this package.

Trenly:
Looking at this more, Softpedia doesn't use a static download url (api changes it per user) so we can't use them anyways. I did find the [github repo](https://github.com/alexandruavadanii/USSF) which directs users to [MSFN](http://www.msfn.org/board/) which directs most users back to Softpedia. Most other sites I've found also direct back to Softpedia.

There is a download on [SoftSea](http://www.softsea.com/download/Universal-Silent-Switch-Finder.html) which may work

There is a download on [Capstanservices](http://www.capstanservices.com/tools-blog/2018/4/4/the-ultimate-silent-switch-finder-ussf) which distributes it inside a Zip (Extracted the zip, hash is the same). 

I'll try the SoftSea download to see if that remains static

denelon:
If this isn't something we think will be highy downloaded, it could be a great .zip to add. (mentioning @ryfu-msft)

I would guess most users who are downloading this are likely to be using it for winget-pkgs submissions, so they would also be good candidates for using the experimental .zip feature as well.

Trenly:
> If this isn't something we think will be highy downloaded, it could be a great .zip to add. (mentioning @ryfu-msft)
> 
> I would guess most users who are downloading this are likely to be using it for winget-pkgs submissions, so they would also be good candidates for using the experimental .zip feature as well.

I think it may be a good package to add the Zip installer for, but if portable works then I don't see a reason to delay

denelon:
Maybe we could create a WIPWnet.UniversalSilentSwitchFinder.zip package. Users could still install the portable with 1.3 and we could use the other as a "temporary" placeholder for the .zip version. I don't necessarily want to maintain that convention, so I would expect to delete it once .zip is GA. We also need to look at the argument and filters for installer type so users can have multiple different installer types in the same manifest and be able to choose the one they want.

* microsoft/winget-cli/issues/1166

Thoughts?

It might help avoid sad 1.3 users 😊.

ItzLevvie (dismissed):

Trenly:
> Maybe we could create a WIPWnet.UniversalSilentSwitchFinder.zip package. Users could still install the portable with 1.3 and we could use the other as a "temporary" placeholder for the .zip version. I don't necessarily want to maintain that convention, so I would expect to delete it once .zip is GA. We also need to look at the argument and filters for installer type so users can have multiple different installer types in the same manifest and be able to choose the one they want.
> 
> * [Specify InstallerType winget-cli#1166](https://github.com/microsoft/winget-cli/issues/1166)
> 
> Thoughts?
> 
> It might help avoid sad 1.3 users 😊.

Might have to be just the Zip installer anyways; Seems SoftSea is returning this as the download now-

```
<script type="text/javascript" language="javascript">alert('Too much download times on this program, please try later.');location.href='http://www.softsea.com/review/Universal-Silent-Switch-Finder.html';</script>
```

Trenly:
@denelon - Blocking Issue on .zip please

ItzLevvie (approved):

ryfu-msft:
This package will be used as a test example for the portables in zip feature.  This manifest will be checked in manually as zip manifests are not yet supported in validation. This also means that the latest stable 1.3 client will not be able to install this package as .zip is not yet supported in that version.

PR #85235, opened 2022-10-19
Add version: Google.AndroidStudio version 4.2.1.0

russellbanks:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/85235)

### DIFF
--- a/manifests/g/Google/AndroidStudio/4.2.1.0/Google.AndroidStudio.installer.yaml
+++ b/manifests/g/Google/AndroidStudio/4.2.1.0/Google.AndroidStudio.installer.yaml
@@ -1,0 +1,25 @@
+# Created with YamlCreate.ps1 v2.2.0 $debug=QUSU.CRLF.7-2-6.Win32NT
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.2.0.schema.json
+
+PackageIdentifier: Google.AndroidStudio
+PackageVersion: 4.2.1.0
+Platform:
+- Windows.Desktop
+MinimumOSVersion: 10.0.0.0
+InstallerType: nullsoft
+Scope: machine
+InstallModes:
+- interactive
+- silent
+InstallerSuccessCodes:
+- 1223
+UpgradeBehavior: install
+AppsAndFeaturesEntries:
+- ProductCode: Android Studio
+Installers:
+- Architecture: x64
+  InstallerUrl: https://redirector.gvt1.com/edgedl/android/studio/install/4.2.1.0/android-studio-ide-202.7351085-windows.exe
+  InstallerSha256: 1DB6AEF39055D4CD69DE26026AA0D3F17D10DD2B66051004580EB5175039934F
+  ProductCode: Android Studio
+ManifestType: installer
+ManifestVersion: 1.2.0
--- a/manifests/g/Google/AndroidStudio/4.2.1.0/Google.AndroidStudio.locale.en-US.yaml
+++ b/manifests/g/Google/AndroidStudio/4.2.1.0/Google.AndroidStudio.locale.en-US.yaml
@@ -1,0 +1,37 @@
+# Created with YamlCreate.ps1 v2.2.0 $debug=QUSU.CRLF.7-2-6.Win32NT
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.2.0.schema.json
+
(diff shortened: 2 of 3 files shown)


ItzLevvie (approved):

justinkb:
@ItzLevvie there are serious problems with these updates russell keeps making pull requests for https://github.com/microsoft/winget-pkgs/commit/0f890c837c9af77d0e6497122a1714e5d8c45d88

this version was already in the tree, under "4.2" to work around the bug that androidstudio version numbers don't get specified correctly in registry.

the same issue exists with https://github.com/microsoft/winget-pkgs/pull/85083 - winget keeps trying to update forever

```
Name                             Id                       Version      Available    Source
------------------------------------------------------------------------------------------
Android Studio                   Google.AndroidStudio     2021.3.1.16  2021.3.1.17  winget
```

not sure if russellbanks is a bot doing these pull requests?

russellbanks:
@justinkb If two different versions have the same ARP Entry, the ARP Entry should be removed from the newer versions where that is the case (see https://github.com/jedieaston/Add-ARPEntries/issues/9). Therefore, if the version is `2021.3.1.16` but what is written to registry is `2021.1`, then we can set `AppsAndFeaturesEntries` on that version, and ommit it on later versions:
```yaml
AppsAndFeaturesEntries:
- DisplayVersion: "2021.3"
```
I'm unsure why this isn't working for the latest versions though, as I am getting the same issue too. I don't believe saying there are serious problems with my PRs and insinuating that I am a bot is the right way to go about this; an issue would have been more appropriate. There is https://github.com/microsoft/winget-cli/issues/2630 open currently.

justinkb:
Sincerely sorry, I meant no offense. I just noticed there were a lot of commits from the account, so thought they were automated!

justinkb:
Now that I'm on my pc again, I'll respond in-depth to your comment

>  If two different versions have the same ARP Entry, the ARP Entry should be removed from the newer versions where that is the case

how is winget supposed to know then the latest update has already been installed? seems like there is no mechanism at all to ensure it does know that? you say "not sure why it doesn't work here", I ask "how would it ever work like that?"

russellbanks:
> how is winget supposed to know then the latest update has already been installed?

It would should check if what is written to the registry matches the most recent ARP Entry, and thus wouldn't actually prompt you to update (unless it was to a version that had a different ARP Entry).

To illustrate this better:

A user has `Major.Minor.1` with `Major.Minor` having an DisplayVersion of `1.2.3`. When `Major.Minor.2` comes out (which has the same DisplayVersion of 1.2.3), winget shouldn't prompt you to update because although they are separate versions, the ARP Entries are the same. It should prompt an update when `Major.NextMinorVersion.Patch` comes out with a different DisplayVersion.

You might say why doesn't *only* the latest version just get added then if winget shouldn't prompt you to update anyway unless the DisplayVersion is changed. The reason that all versions are added is because users can benefit from installing a specific version. For example, if the newest Canary version (with it updating inside the program) had a bug that meant you were unable to work, essentially deleting the previous version to replace it with the newer version would mean that you could not revert via winget.

> I just noticed there were a lot of commits from the account

I'm just very active, like the other top contributors are. The only commits I do that are automated will be marked with `ARPEntries:` in the title and `Pull request has been automatically created using Add-ARPEntries` in the body.

justinkb:
ah I understand, so winget wouldn't inform me at all that the "2021.3" version I'm running is possibly out of date? that also seems less than ideal. for now, is there any winget command to ignore a specific package when doing --all? again, really didn't mean to offend with the bot comment. keep up the good work

russellbanks:
> ah I understand, so winget wouldn't inform me at all that the "2021.3" version I'm running is possibly out of date? that also seems less than ideal.

Ideally and I appreciate that it isn't a perfect solution but that is essentially down to WinGet's nature to use registry entries to check for update and what's installed, etc, but of course the benefit for that is program's don't have to be installed by WinGet in order for them to be indentified as installed.

I'll raise this in the Gitter/Element chat with the mods and see if we can come to a solution/fix.

russellbanks:
https://github.com/microsoft/winget-cli/issues/2366

russellbanks:
I've outlined a possible fix in your [issue](https://github.com/microsoft/winget-pkgs/issues/86933).

Issue #85695, opened 2022-10-20
[Package Request]: Python.Python.3.Latest


MarcoPeraza:
Hi, why are there only packages for specific versions of Python3 and no package that just installs the latest version? I recognize that Python2 and Python3 are different enough to not be interchangeable. But when I install Python3, I normally want the latest version of Python3, just like I want the latest version of Powershell when I install Powershell.

TheBossMagnus:
all older python 3 are simply named "python 3". I think they should be renamed with also the .x (python 3.0/3.1/.....) like in the id.
The only "python 3" should be the latest (3.10 now).
Also, the 3.11 RC package doesn't exist.
But before doing any pr I'd like to listen to other opinions and possible breaking changes...

Trenly:
The `PackageName` should stay as `Python 3` because that is what is written as the display name in Control Panel. Besides, even if it were changed, winget would still find all the subversions when searching for `Python3`. The best way to get `winget install Python3` would be to change the moniker on the latest subpackage of Python3 to be `python3`, since winget treats an exact match with a moniker the same as an exact match for a package id.  This does have the downfall of needing to be changed when new minor versions are released (e.g 3.10 -> 3.11) but it should work.

Regarding 3.11 not existing yet, I'm believe this is because it is not yet released to stable and is only a release candidate build.

TheBossMagnus:
> Regarding 3.11 not existing yet, I'm believe this is because it is not yet released to stable and is only a release candidate build.
It's a good idea to make a package anyway(python 3.11 RC)?

Trenly:
`winget install python3` now should install the latest version

ppvnf:
I've a question: when python 3.11 is released, `winget upgrade --all` is gonna be able to update python 3.10 to 3.11?

MarcoPeraza:
I'm not sure a moniker solves the problem. Packages are not guaranteed to have unique monikers, but they are guaranteed to have unique IDs.

NJT145:
At this moment, there is a problem about `--id=Python.Python.3`. it seems that Python.Python.3 is not pointed to the latest version of Python3, and latest Python3 is with `--id=Python.Python.3.11`.
I tried to download Python3 with that cade and it doesn't work:
`winget install -e -i --id=Python.Python.3 --source=winget --scope=machine`
It says `No package found matching input criteria.`.

Trenly:
[...]

PR #85817, opened 2022-10-20
Change Python3 Moniker for latest version

Trenly:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/85817)


russellbanks:
Will we have to change this (and other versions with this moniker) back to `python3.10` if/when a new version comes out and that new version's moniker to `python3`?

Trenly:
> Will we have to change this (and other versions with this moniker) back to `python3.10` if/when a new version comes out and that new version's moniker to `python3`?

Yes, we will. However, this change makes it so that users can just do `winget install python3` for the latest version. Ref #85695

ItzLevvie (approved):

PR #86190, opened 2022-10-25
Python.Python version 3.11.0

remyjette:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/86190)

### DIFF
--- a/manifests/p/Python/Python/3/11/3.11.0/Python.Python.3.11.installer.yaml
+++ b/manifests/p/Python/Python/3/11/3.11.0/Python.Python.3.11.installer.yaml
@@ -10,7 +10,7 @@ InstallModes:
 - interactive
 - silent
 - silentWithProgress
-UpgradeBehavior: install
+UpgradeBehavior: uninstallPrevious
 Commands:
 - py
 - python
--- a/manifests/p/Python/Python/3/11/3.11.0/Python.Python.3.11.installer.yaml
+++ b/manifests/p/Python/Python/3/11/3.11.0/Python.Python.3.11.installer.yaml
@@ -1,0 +1,73 @@
+# Created with YamlCreate.ps1 v2.2.0 $debug=QUSU.CRLF.7-2-6.Win32NT
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.2.0.schema.json
+
+PackageIdentifier: Python.Python.3.11
+PackageVersion: 3.11.0
+InstallerLocale: en-US
+MinimumOSVersion: 10.0.0.0
+InstallerType: burn
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+UpgradeBehavior: install
+Commands:
+- py
+- python
+FileExtensions:
+- py
+- pyc
+- pyo
+- pyw
+ReleaseDate: 2022-10-24
+Installers:
+- Architecture: x64
+  Scope: user
+  InstallerUrl: https://www.python.org/ftp/python/3.11.0/python-3.11.0-amd64.exe
+  InstallerSha256: B5884FA3F05F88BBB617D08584930770C00BBCF675F2865A9161C2358829B605
+  InstallerSwitches:
+    InstallLocation: DefaultJustForMeTargetDir=<INSTALLPATH>
+    Custom: InstallAllUsers=0 PrependPath=1
+  AppsAndFeaturesEntries:
+  - DisplayName: Python 3.11.0 (64-bit)
+    Publisher: Python Software Foundation
+    DisplayVersion: 3.11.150.0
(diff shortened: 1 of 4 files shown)


SpecterShell:
ARM64 version
https://www.python.org/ftp/python/3.11.0/python-3.11.0-arm64.exe

manifests/p/Python/Python/3/11/3.11.0/Python.Python.3.11.installer.yaml
```diff
@@ -0,0 +1,73 @@
...
+PackageIdentifier: Python.Python.3.11
+PackageVersion: 3.11.0
+InstallerLocale: en-US
+MinimumOSVersion: 10.0.0.0
+InstallerType: burn
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+UpgradeBehavior: install
```
fernandog, line 13: shouldn't this setting be `UpgradeBehavior: uninstallPrevious` ?

remyjette: My preference would be for all Python versions to have consistent behavior. I don't have an opinion on the correct setting for UpgradeBehavior, but if it were to change then it should change for all previous versions as well, so I decided to keep this to just adding support for the new version.

fernandog: @denelon I'm new to this. what your thought on this?

remyjette: Given this discussion I think it makes sense to switch to `uninstallPrevious` for Python.

Prior to each major.minor being split into a separate package this would not have been correct as an upgrade from 3.10 to 3.11 (for example) could break code.

However, patch releases are compatible with each other and just contain bugfixes, and more importantly, security fixes. So I don't think it makes any sense to leave the old version around.

This would also match behavior with other package managers and repositories, where you can specifically install a major.minor version of Python, but the patch version will always be latest.

I'll update this PR to switch to uninstallPrevious.

@denelon I can make a PR to change previous versions of Python to match. Can this be done in one PR, or do I need to make a separate PR for each version?

denelon: @remyjette unfortunately, it requires one PR per manifest (package/version). The code on the back end to run validation can't handle figuring out which one failed (or which of multiple failures). We're looking at options to make this better/easier to modify, but it's a messy problem.

remyjette: Okay, I'll make a bunch of PRs then :)

remyjette: Submitted PRs for every version that had `UpgradeBehavior: install` to switch to `UpgradeBehavior: uninstallPrevious` (except the .0 versions because that seemed like unnecessary churn)

SpecterShell: I don't think switching to uninstallprevious is necessary because for minor version upgrade the installer can override the existing installation instead of adding another version.
image

remyjette:
> ARM64 version https://www.python.org/ftp/python/3.11.0/python-3.11.0-arm64.exe

Thanks for pointing that 3.11.0 begins native support for ARM64. Took me a bit to get my ARM64 machine working again as it's been a bit since I last used it, but I was able to add arm64 to this manifest and validate that it works.

manifests/p/Python/Python/3/11/3.11.0/Python.Python.3.11.installer.yaml
fernandog: My 3.11 was installed side-by-side the 3.10

ItzLevvie (approved):

ItzLevvie (approved):

manifests/p/Python/Python/3/11/3.11.0/Python.Python.3.11.installer.yaml
SpecterShell: Oh it is the patch version (3.10.6 → 3.10.8) that can upgrade directly. Minor versions (3.10 and 3.11) can be installed side-by-side. Sorry for mistakes.
All Python versions in WinGet are categorized by minor versions (`Python.Python.3.10`, `Python.Python.3.9`, etc.) and each minor version can only be upgraded by patch version and doesn't affect each other.

PR #89316, opened 2022-11-21
Arturia.MIDIControlCenter version 1.14.1

dongle:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/89316)

### DIFF
--- a/manifests/a/Arturia/MIDIControlCenter/1.14.1/Arturia.MIDIControlCenter.installer.yaml
+++ b/manifests/a/Arturia/MIDIControlCenter/1.14.1/Arturia.MIDIControlCenter.installer.yaml
@@ -1,0 +1,13 @@
+# Created using wingetcreate 1.1.2.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.2.0.schema.json
+
+PackageIdentifier: Arturia.MIDIControlCenter
+PackageVersion: 1.14.1
+Installers:
+- Architecture: x86
+  InstallerType: inno
+  InstallerUrl: https://downloads.arturia.net/extra/mcc/MIDI_Control_Center_1_14_1_2.exe
+  InstallerSha256: D3F90AB2480B38B434E6109B77DE175FC6CE30FE41A14D9A7EE5DD4730FA9492
+ManifestType: installer
+ManifestVersion: 1.2.0
+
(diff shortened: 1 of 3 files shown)


stephengillie:
Hi @dongle,

Validation is catching the Arturia USB MIDI driver setup. This launches after the MIDI Control Center install completes. 
image

Is there an installer switch that would make this driver also silent? If not, the driver could possibly be added as a separate package, and then added as a dependency on this package.

dongle:
> Hi @dongle,
> 
> Validation is catching the Arturia USB MIDI driver setup. This launches after the MIDI Control Center install completes. image
> 
> Is there an installer switch that would make this driver also silent? If not, the driver could possibly be added as a separate package, and then added as a dependency in this package's manifest.

Hm, I see -- makes sense. Just to ensure I'm following, the gist is that the installer closes, then launches the driver setup, which would goof scripting winget since winget would return, passing control to the subsequent instructions in the script?

For future reference, is there a ready-made way to check for this locally so I can avoid repeating this?

As far as I know, there's no separate download for the Arturia USB MIDI drivers, and there's no documented switch to pass to the installer. I'll admit that I'm mostly a linux user (Windows for music & games) so I'm not well-informed about ways to inspect windows installers for potential options -- is there a resource I could read or a term I should search to get up to speed?

Thanks for your help!

stephengillie:
> Hm, I see -- makes sense. Just to ensure I'm following, the gist is that the installer closes, then launches the driver setup, which would goof scripting winget since winget would return, passing control to the subsequent instructions in the script?

Validation would wait for the driver installer, were it unattended. It times out, I believe, instead of providing an exit code, signaling a user prompt and not-quite-unattended installer. 

> For future reference, is there a ready-made way to check for this locally so I can avoid repeating this?

Validation happens inside a fresh install of Windows, which uncovers dependencies like this in many other PRs. For testing, we also use Windows Sandboxes which are similarly barren.

> As far as I know, there's no separate download for the Arturia USB MIDI drivers, and there's no documented switch to pass to the installer. I'll admit that I'm mostly a linux user (Windows for music & games) so I'm not well-informed about ways to inspect windows installers for potential options -- is there a resource I could read or a term I should search to get up to speed?

There are a few major installer flavors, including Nullsoft, Inno, Burn, and MSI. Winget has the default silent switches for many of these built-in. Independent installer and nonstandard switches can also be declared directly in the manifest. Most installers can be queried with switches including "/?" and "--help". It can be a guessing game to find what works, but between command-line trial-and-error and some internet searches, the necessary switches can often be unearthed.

A problem with the present situation is that Winget isn't launching the driver installer - the program installer does. So Winget can't directly set switches for the driver installer. Unfortunately, packages must have a completely unattended mode to be included in Winget. 

> Thanks for your help!

Anytime!

Trenly:
@msftbot interactive-only due to driver prompt

ItzLevvie (changes requested):

stephengillie:
Unfortunately, installers must have a fully unattended mode to be included in the Winget manifest.

PR #89511, opened 2022-11-23
Remove executable flag from manifests

me-and:
Manifest files are not meaningfully executable.  The vast majority of manifest files don't have Git's executable flag set, but some do.  Clear those flags for consistency with the majority of other manifests.

I suspect that most of these flags exist because of either (a) folks having misconfigured Git, or (b) from mixing Windows and either WSL or Cygwin tooling and not correcting for the different ways the two use executable flags.  Certainly (b) is what caught me out for the few files here that I originally submitted.

I've not done most of the checks above because I've not changed the content of any manifest files here, and I don't think any of the winget tooling cares about the executable flag. This is entirely down to my dislike of inconsistent file properties when there's no good reason for that.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/89511)


russellbanks (changes requested):
You can only change one manifest per pull request. An issue would be better for a big change like this, as others can have their say on it.

ItzLevvie (approved):

russellbanks:
@me-and is this something we could enforce with a .gitattributes? Otherwise, I can't see how we could maintain the flags of files being set a certain way.

me-and:
@russellbanks I'll raise an issue for this then :)

AIUI the way to enforce this sort of thing is with hooks or other automation; in practice, here, I suspect that means any checks would need to work the same way as, say, the spelling checks that already exist. It should be a pretty small thing to add to the existing GitHub Actions tests that are already done on this repo.

stephengillie:
Closing as discussion is proceeding in 
* https://github.com/microsoft/winget-pkgs/issues/90127

If these changes are required, then 11 individual PRs will need to be opened.

Issue #90249, opened 2022-12-01
[New Feature]: Enforce HTTPS URLs where possible


JamieMagee:
### Description of the new feature/enhancement

There are quite a few manifests which [contain HTTP URLs][1]. Even more alarming, there are many manifests which [have `InstallerUrl` pointing to HTTP URLs][2].

This is potentially a security risk, and falls under [CWE-319: Cleartext Transmission of Sensitive Information][3].

[1]: https://github.com/search?q=repo%3Amicrosoft%2Fwinget-pkgs+%2Fhttp%3A%2F&type=code
[2]: https://github.com/search?q=repo%3Amicrosoft%2Fwinget-pkgs+%2FInstallerUrl%3A+http%3A%2F&type=code&p=2
[3]: https://cwe.mitre.org/data/definitions/319.html

### Proposed technical implementation details (optional)

Create a GitHub Action to validate that all possible URLs are HTTPS for manifests changed in a PR. An exception list should be maintained for URLs or domains which aren't served over HTTPS.

A starting point might be the following shell script which:

1. Finds all instances of HTTP URLs
2. Replaces the `http:` with `https:`
3. Attempts to make a HEAD request to the HTTPS equivalent of the URL
4. Prints the HTTPS URL if it's valid

```sh
grep --no-filename --only-matching --recursive --include "*.yaml" "http://.*\$" manifests/ | sed "s/http:/https:/" | xargs -n 1 curl --head -
-silent --output /dev/null --write-out "%{url}\n"
```

JamieMagee:
It looks like the validation tracks this by adding the `URL-Validation-Error` label and adding a comment like:

> Error Message: A secure URL : ... is available. Please use a secure URL in place of the non-secure URL for enhanced security.

The new validation appears to have been in place since the 27th February 2025: https://github.com/search?q=repo%3Amicrosoft%2Fwinget-pkgs+%22A+secure+URL%22&type=pullrequests&s=updated&o=asc

PR #90273, opened 2022-12-02
Add `ensure-https` workflow

JamieMagee:

-----

This partially addresses #90249.

`EnsureHttps.ps1` will:
1. Get a list of changed `.yaml` files in the `manifests` directory for a PR
2. For each changed file:
	1. Check if it contains an `http://` URL
	2. If yes, for each `http://` URL:
		1. It will see if the `https://` equivalent is valid
		2. If yes, it will write out an [error command][2]

[Check out an example of this workflow in my fork][1]

I think this is a good starting point, that can be improved on with URL exclusions, and eventually be made a required workflow. To start with I'd like to run it as an optional workflow to get some feedback.

[1]: https://github.com/JamieMagee/winget-pkgs/actions/runs/3597956799/jobs/6060255513#step:3:5
[2]: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/90273)

### DIFF
--- a/.github/EnsureHttps.ps1
+++ b/.github/EnsureHttps.ps1
@@ -1,0 +1,24 @@
+#!/usr/bin/env pwsh
+
+Set-StrictMode -Version Latest
+$ErrorActionPreference = 'Stop'
+
+$changedFiles = git diff --name-only "origin/$env:GITHUB_BASE_REF" HEAD -- manifests
+
+foreach ($changedFile in $changedFiles) {
+  $http = Select-String -Path $changedFile `
+    -Pattern '(http://.*)$' `
+    -AllMatches
+  foreach ($match in $http.Matches) {
+    if ($match.Groups[0].Value) {
+      $https = $match.Groups[0].Value.Replace('http://', 'https://')
+      $res = Invoke-WebRequest -Uri $https `
+        -Method HEAD `
+        -TimeoutSec 10 `
+        -UseBasicParsing
+      if ($res.StatusCode -eq 200) {
+        Write-Output "::error file=$changedFile::Found HTTP link: $($match.Value)"
+      }
+    }
+  }
+}
--- a/.github/workflows/ensure-https.yaml
+++ b/.github/workflows/ensure-https.yaml
@@ -1,0 +1,18 @@
+name: Ensure HTTPS
+
+on:
+  pull_request:
+
+jobs:
+  https:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - run: .github/EnsureHttps.ps1
+        shell: pwsh
+


ItzLevvie (approved):

vedantmgoyal9 (changes requested):
It would be better if @github-actions[bot] will comment on the PR.

JamieMagee:
@denelon is it possible to get a review of this by someone with merge rights?

denelon:
@JamieMagee I think we want to look at something other than a GitHub action to do this work. Let me see about getting one of the engineers to comment here and see what the best path forward looks like.

JamieMagee:
Thanks!

My intention with a GitHub Action was to prevent regressions by re-introducing an HTTP URL, and allow for incremental upgrades from HTTP to HTTPS as files were edited.

denelon:
Yes, we've also been thinking about the other potential scenarios where we might need to make immediate changes on the back end of things, and ensuring consistency. We don't want to accidentally introduce something breaking that would be blocked, or cause other problems for authors submitting PRs. We've got work in progress to deal with consistent line endings for example.

ItzLevvie:
[Policy] Blocking-Issue

.github/EnsureHttps.ps1
```diff
@@ -0,0 +1,24 @@
...
+    -AllMatches
+  foreach ($match in $http.Matches) {
+    if ($match.Groups[0].Value) {
+      $https = $match.Groups[0].Value.Replace('http://', 'https://')
+      $res = Invoke-WebRequest -Uri $https `
+        -Method HEAD `
+        -TimeoutSec 10 `
+        -UseBasicParsing
+      if ($res.StatusCode -eq 200) {
+        Write-Output "::error file=$changedFile::Found HTTP link: $($match.Value)"
```
stephengillie, line 20:
```suggestion
        Write-Output "::error file=$changedFile::Found HTTPS link: $($match.Value)"
```

stephengillie:
In general, this is a good idea, as encrypting sites through HTTPS has many benefits. Insofar as process steps, this check could more easily be made upstream, in the manifest creation step. It could be easily integrated into [YamlCreate](https://github.com/Trenly/winget-pkgs/blob/master/Tools/YamlCreate.ps1), [WinGetCreate](https://github.com/microsoft/winget-create), and [Komac](https://github.com/russellbanks/Komac). And this would allow the manifest creator to decide if they want strict HTTPS checking, or want to leave HTTP links bare. (I'm not a fan of mandating HTTPS - I believe that every site should offer both if possible, and that users should be educated and free to choose.)

This implementation might be missed by many users, especially those who aren't used to checking into the PR after creating it. Also, users tend to want the validation process to output a binary yes/no answer, and this might be interpreted as a confusing "yes, but..." answer to this step.

.github/EnsureHttps.ps1
JamieMagee: I'm not sure this makes sense as an error message.

If we are trying to prompt users to upgrade their links from HTTP to HTTPS, an error message that says `Found HTTPS link http://example.com` would be confusing.

JamieMagee:
@stephengillie I agree that adding this to the manifest creation tools would help 'shift-left' on where HTTPS is enforced. But adding this as a required workflow to PRs will prevent any new HTTP links being added, and will force PR authors to fix an HTTP URL if they otherwise edit a manifest.

I disagree that HTTPS should be optional if the server supports it. It would open up the possibility for manifest authors to perform downgrade attacks on users.

As for making it visible to PR authors, I can update the script with `exit 1` if an HTTP link is found (with an HTTPS link available) which will prevent the PR from being merged.

.github/EnsureHttps.ps1
stephengillie: I had misread, thanks for catching.

denelon:
I'm going to go ahead and close this PR.

We should be doing this check inside the current validation pipeline.

I put the following issue into the next release milestone:
* https://github.com/microsoft/winget-pkgs/issues/90249

JamieMagee:
@denelon I'm not fussed _how_ HTTPS is enforced, only that it is enforced. I would appreciate being kept in the loop on this. Feel free to reach out to me internally.

PR #92397, opened 2022-12-27
Add Dependencies to LeawoSoftware.Blu-ray 3.0.0.2

mdanish-kh:
Add dependencies as requested in https://github.com/microsoft/winget-pkgs/pull/92282#discussion_r1057870465

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/92397)


russellbanks (approved):

elisimpson:
There's a pretty big bug with this dependency add. The Microsoft.VCRedist.2010.x64 dep doesn't make sense, as it's a x86 app and it ends up installing an older version of the 2008 VCRedist x86, but installs it to the root of the C: drive. The app runs fine after uninstalling the corrupted VCDist... not sure it's really necessary at all? But in any case it needs to be fixed by either removing the dep or redoing it so it installs the correct version in the correct place (not the root of C:)

mdanish-kh:
> There's a pretty big bug with this dependency add. The Microsoft.VCRedist.2010.x64 dep doesn't make sense, as it's a x86 app and it ends up installing an older version of the 2008 VCRedist x86, but installs it to the root of the C: drive. The app runs fine after uninstalling the corrupted VCDist... not sure it's really necessary at all? But in any case it needs to be fixed by either removing the dep or redoing it so it installs the correct version in the correct place (not the root of C:)

@stephengillie - Any thoughts?

PR #93519, opened 2023-01-11
FontForge.FontForge - remove old version 08-03-2022

krystofkrticka:
Remove old package version

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/93519)


krystofkrticka:
IDK why but winget sees this version (08-03-2022) instead of new version (01-01-2023)

stephengillie:
> IDK why but winget sees this version as latest (08-03-2022) instead of new version (01-01-2023)

The dashes might be causing versions to be sorted alphabetically instead of temporally.

stephengillie:
The package is still available - why remove it?

krystofkrticka:
@stephengillie I know, I also would like to leave it in repo but this, right now I don't know how to fix this quickly without removing the package from repo.

krystofkrticka:
Should I report it as bug in winget development repo?

stephengillie:
> @stephengillie I know, I also would like to leave it in repo but this, right now I don't know how to fix this quickly without removing the package from repo.

@krystofkrticka, If I'm understanding correctly, this has 2 impacts:
- Update always available
- Installing the default version does not install the latest version

> Should I report it as bug in winget development repo?

Yes, to track the version number sorting bug. This might be the only package in our manifests with this format of temporal versioning. 

As a workaround, until the version sorting is updated, please try with the version specified:
`winget install fontforge.fontforge --version "01-01-2023"`

krystofkrticka:
> As a workaround, until the version sorting is updated, please try with the version specified: `winget install fontforge.fontforge --version "01-01-2023"`

I already have newest version because I updated before fixing version number to be same as in appwiz.cpl

krystofkrticka:
> @krystofkrticka, If I'm understanding correctly, this has 2 impacts:
> 
> * Update always available
> * Installing the default version does not install the latest version

Yes, update is always available and not matching latest version.

krystofkrticka:
> Yes, to track the version number sorting bug. This might be the only package in our manifests with this format of temporal versioning.

@stephengillie OK, I'll do it as soon as I have little bit of time to write it.

krystofkrticka:
@stephengillie I see you are trying to change version in repo but version will colide with version number in appwiz.cpl 
image

krystofkrticka:
https://github.com/microsoft/winget-cli/issues/2840

PR #93941, opened 2023-01-17
Update: Prusa3D.PrusaSlicer version 2.3.2

Trenly:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/93941)

### DIFF
--- a/manifests/p/Prusa3D/PrusaSlicer/2.3.2/Prusa3D.PrusaSlicer.installer.yaml
+++ b/manifests/p/Prusa3D/PrusaSlicer/2.3.2/Prusa3D.PrusaSlicer.installer.yaml
@@ -12,8 +12,8 @@ InstallModes:
 - silent
 - silentWithProgress
 InstallerSwitches:
-  Silent: /silent
-  SilentWithProgress: /passive
+  Silent: /exenoupdates /silent
+  SilentWithProgress: /exenoupdates /passive
 UpgradeBehavior: install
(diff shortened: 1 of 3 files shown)


stephengillie:
This package version also prompts to install a newer version.
image

Is there an installer switch to bypass this prompt?

stephengillie:
Here is the installer's output when queried: 
image

denelon:
Does it do the same thing when installed with "--silent"?

As long as the latest version honors our intent, and earlier versions would have done so, they would have been grandfathered in.

stephengillie:
Found it!
`/exenoupdates`

stephengillie:
> Found it! `/exenoupdates`

After running with both switches `/exenoupdates /passive`, the install runs silently. But then it prompts to install a driver: 
image

Trenly:
> Found it! `/exenoupdates`

Perfect! Doesn't work on my end when using the `Custom:` installer switches; Probably they need to be in a certain order. Putting them both into the `SilentWithProgress` worked. I think that it would still pass validation since the process controlling the driver prompt is a sub-process and not tied to the installer completing

> As long as the latest version honors our intent, and earlier versions would have done so

Can confirm that 2.5.0 (latest) does install silently

stephengillie (approved):

PR #94480, opened 2023-01-24
traviskunkel.MealPicker version 1.1.0.0

HabbisSkunkel:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/94480)

### DIFF
--- a/manifests/t/traviskunkel/MealPicker/1.1.0.0/traviskunkel.MealPicker.installer.yaml
+++ b/manifests/t/traviskunkel/MealPicker/1.1.0.0/traviskunkel.MealPicker.installer.yaml
@@ -1,0 +1,19 @@
+# Created using wingetcreate 1.1.2.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.2.0.schema.json
+
+PackageIdentifier: traviskunkel.MealPicker
+PackageVersion: 1.1.0.0
+Installers:
+- Platform:
+  - Windows.Universal
+  - Windows.Desktop
+  MinimumOSVersion: 10.0.17763.0
+  Architecture: x86
+  InstallerType: msix
+  InstallerUrl: https://github.com/HabbisSkunkel/WinGetTest/releases/latest/download/MealPicker.msix
+  InstallerSha256: D33DEEB77FC6320A1239CAE8AEBE3402171A3B18AF4073B3F002510173D2153F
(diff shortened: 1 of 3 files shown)


HabbisSkunkel:
This uses a self-signed certificate and the default YAML produces by wingetcreate.

stephengillie:
Hi @HabbisSkunkel,

> This uses a self-signed certificate, and the default YAML produced by "wingetcreate new".

A self-signed certificate may not be sufficient. Validation is failing with this error:
> The root certificate and all intermediate certificates of the signature in the app package or bundle must be trusted.
> A certificate chain could not be built to a trusted root authority.

From https://github.com/microsoft/winget-pkgs/pull/93522#issuecomment-1379644451: [Sign a Windows 10 app package - MSIX | Microsoft Learn](https://learn.microsoft.com/windows/msix/package/signing-package-overview)

> To successfully install a Windows 10 application, the package doesn't just have to be signed but also trusted on the device. This means that the certificate has to chain to one of the trusted roots on the device.

HabbisSkunkel:
Hello @stephengillie, 

This is a self-signed certificate that I created and exported following the instructions at [Create a certificate for package signing](https://learn.microsoft.com/en-us/windows/msix/package/create-certificate-package-signing). The only change I made was adding the -NotAfter switch to set the certificate to expire a year from now.

I have exported the certificate to use on machines this app is intended to be put on (they will have the certificate as a trusted root authority).

Hope this helps!

stephengillie:
> I have exported the certificate to use on machines this app is intended to be put on (they will have the certificate as a trusted root authority).

Ah - the problem here is that the certificate hasn't been exported to the validation pipeline VMs. If the package isn't meant for public distribution, implied by the local certificates, then it might not be a good choice for inclusion in the public manifest. Instead, a local manifest could be used. Using these same files, the command `winget install --manifest c:\path\to\winget\manifest\files` would have Winget install the package described by the manifest files. And since these PCs have the certificate already installed, they would be able to validate and trust the application.

HabbisSkunkel:
> If the package isn't meant for public distribution, implied by the local certificates, then it might not be a good choice for inclusion in the public manifest.

Thanks for the information. I'll go ahead with the method you provided. Thanks for the help!

PR #94817, opened 2023-01-27
Fix netsurf version

aaronliu0130:

-----
The uninstall command exits before the user is able to click the uninstaller's confirmation button, though the confirmation dialog stays and confirming does successfully uninstall. Not sure why.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/94817)

### DIFF
--- a/manifests/n/NetSurf/NetSurf/3.10.0/NetSurf.NetSurf.installer.yaml
+++ b/manifests/n/NetSurf/NetSurf/3.10.0/NetSurf.NetSurf.installer.yaml
@@ -9,4 +9,4 @@ Installers:
   InstallerUrl: https://download.netsurf-browser.org/netsurf/releases/pre-built/windows/3.10-netsurf.exe
   InstallerSha256: 2AA12EB57198AFF9A216E764B9C8B7956CA98CD6F5544A491D228849094DAE6D
 ManifestType: installer
-ManifestVersion: 1.0.0
+ManifestVersion: 1.4.0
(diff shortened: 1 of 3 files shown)


aaronliu0130:
Weird, I ran winget validate, hold on

stephengillie:
For 1.4 manifests, we're monitoring for winget 1.4 to become the majority version among users. We're hopeful this will be soon, possibly in a few days. Sorry for any confusion.

aaronliu0130:
Ah, ok. So we'll restart the workflow later?

stephengillie:
> Ah, ok. So we'll restart the workflow later?

Yes. Several PRs are pending for this.

aaronliu0130:
> The uninstall command exits before the user is able to click the uninstaller's confirmation button, though the confirmation dialog stays and confirming does successfully uninstall. Not sure why.

Could somebody give me some pointers about this? Looks like a standard NSIS uninstaller.

stephengillie:
> > The uninstall command exits before the user is able to click the uninstaller's confirmation button, though the confirmation dialog stays and confirming does successfully uninstall. Not sure why.
> 
> Could somebody give me some pointers about this? Looks like a standard NSIS uninstaller.

Is this the Netsurf uninstaller's behavior? 

While we push for unattended installers, we're less strict about the uninstaller. Though we still prefer completely unattended uninstallation, if possible, a handful do require user interaction. I'm not sure, but Winget might use the installer switches on the uninstaller, expecting the same effect as with the installer. The uninstaller here might react in this unusual way if the switch combination were unexpected.

aaronliu0130:
> I'm not sure, but Winget might use the installer switches on the uninstaller, expecting the same effect as with the installer.
The silent switches should work across NSIS. Regardless of that, the winget uninstall process exits before the uninstall is finished, which is very confusing.

stephengillie (approved):

PR #95092, opened 2023-01-30
New package: Smallstep.step-ca version 0.23.1

tashian:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/95092)


stephengillie:
In a VM, this installs normally, but presents an error when launched:
> error opening C:\Users\user\.step\config\ca.json: open C:\Users\user\.step\config\ca.json: The system cannot find the path specified.

tashian:
Hi @stephengillie, this is the expected output when the server is started with no configuration file, however I can understand why it's confusing. I've opened https://github.com/smallstep/certificates/issues/1248 to make the output more helpful.
We cannot ship a default configuration file with the package, because it contains unique private keys, etc.

However, even once we have that, the command is designed to return an error status code when run by itself and when there is no configuration file.

Is there a way to mark this package as OK, under the circumstances?

Alternatively, we may be able to write a setup script that generates a default server configuration when the package is installed. Would that work?

tashian:
We have added a much more helpful message to the output, to cover this use case.
It will be in our next release.
I will close this now and open a new PR for that release soon.

I've also added `Smallstep.step` as a dependency to this project, so that folks can get set up immediately by running `step ca init`.

PR #95268, opened 2023-02-01
New version: pnpm.pnpm version 7.26.3

russellbanks:


sitiom:
Duplicate of #95215

russellbanks:
> Duplicate of #95215

Thanks for letting me know @sitiom - hadn't realised pnpm was being automated now. This PR does have more metadata (you can add that to the other PR if you want) but I'll close in favour of that one. WinGet Releaser intends to use Komac as a backend in the near future so it will have the extra metadata later on.

sitiom:
The only extra field I see here is the `ReleaseNotes`, but it would be great to have extra metadata

russellbanks:
> The only extra field I see here is the `ReleaseNotes`, but it would be great to have extra metadata

1.4.0 Schemas are being accepted now, but that's not mandatory to change; WinGet Releaser still creates 1.2.0 schemas currently. The Release Notes would be nice to add though :)

PR #95321, opened 2023-02-01
Microsoft.Git version 2.39.1.0.0

ldennington:

-----

### DIFF
--- a/manifests/m/Microsoft/Git/2.39.1.0.0/Microsoft.Git.locale.en-US.yaml
+++ b/manifests/m/Microsoft/Git/2.39.1.0.0/Microsoft.Git.locale.en-US.yaml
@@ -1,0 +1,26 @@
+# Created using wingetcreate 1.2.5.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.4.0.schema.json
+
+PackageIdentifier: Microsoft.Git
+PackageVersion: 2.39.1.0.0
+PackageLocale: en-US
+Publisher: The Git Fundamentals Team at GitHub
+PublisherUrl: https://github.com/microsoft/git
+PublisherSupportUrl: https://github.com/microsoft/git/issues
+Author: Microsoft
+PackageName: Microsoft Git
+PackageUrl: https://aka.ms/ms-git
+License: GPLv2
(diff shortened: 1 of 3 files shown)


manifests/m/Microsoft/Git/2.39.1.0.0/Microsoft.Git.locale.en-US.yaml
```diff
@@ -0,0 +1,26 @@
...
+PackageVersion: 2.39.1.0.0
+PackageLocale: en-US
+Publisher: The Git Fundamentals Team at GitHub
+PublisherUrl: https://github.com/microsoft/git
+PublisherSupportUrl: https://github.com/microsoft/git/issues
+Author: Microsoft
+PackageName: Microsoft Git
+PackageUrl: https://aka.ms/ms-git
+License: GPLv2
+LicenseUrl: https://github.com/microsoft/git/blob/vfs-2.34.0/COPYING
```
derrickstolee, line 14: This needs to be updated to not be `vfs-2.34.0`, instead `vfs-2.39.1`, right?

ldennington: This is a bit of a conundrum. We use the `winget-create` tool to automatically generate this manifest for us. This tool knows to update things like version but not to update urls like this one. Normally, we could just use the name of the default branch and be good to go, but since the microsoft/git default branch changes, we can't get away with that here.

derrickstolee: How can we update it? Because I think the commit distance between the vfs-2.34.0 branch and our vfs-2.39.1 branch is why the automated build failed, I think.

ldennington: My current thinking is to modify our process. Currently we generate the manifest and submit with a single `winget-create` command. I _think_ we can instead:

1. Generate the manifest and write to a file.
2. Use `sed` to correct the `LicenseUrl` and `CopyrightUrl` fields.
3. Submit the file.

PR #95344, opened 2023-02-01
Delete the manifests for /j/JackieLiu/NotepadsApp

0x7c13:
Remove winget managed version of Notepads app.

-----

I am the owner of Notepads app. This PR is to delete winget version of Notepads app since winget already supports store apps. When issuing "winget install notepads", two versions of the app will be listed, and it is causing confusion to users thus I am deleting the winget managed one here. Thanks!

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://api.github.com/repos/microsoft/winget-pkgs/pulls/95344)


0x7c13:
Another question I have is that with the removal of this winget managed version, will existing user still be getting automatic updates from store (If they install the managed version instead of the store one)? Or they have to re-install the store version to make sure it can be automatically updated?

Trenly:
Only one manifest can be changed per PR. Please submit separate PR's for each version you wish to remove

> Another question I have is that with the removal of this winget managed version, will existing user still be getting automatic updates from store (If they install the managed version instead of the store one)? Or they have to re-install the store version to make sure it can be automatically updated?

I'm not sure. It *should* get automatic updates, but you would have to test to be certain

Issue #88726, opened 2022-11-16
voidtools.Everything.Lite updated to latest version but winget keeps showing update available


jmaraujouy:
### Brief description of your issue

Winget shows that an update is available, even though the software is up to date with the latest version. (id="voidtools.Everything.Lite" version="1.4.1.1022")

`winget upgrade`
```
Name                             Id                        Version      Available  Source
-----------------------------------------------------------------------------------------
Everything 1.4.1.1022 Lite (x64) voidtools.Everything.Lite < 1.4.1.1015 1.4.1.1022 winget
1 upgrade available.
```

### Steps to reproduce

`winget install "voidtools.Everything.Lite"`
`winget upgrade`

### Expected behavior

No available updates should be displayed (with "_winget upgrade_") if the latest version is already installed.

### Actual behavior

Winget shows that an update is available, even though the software is up to date with the latest version.

### Environment

```shell
Administrador de paquetes de Windows v1.3.2691
Copyright (c) Microsoft Corporation. Todos los derechos reservados.

Windows: Windows.Desktop v10.0.22621.819
Arquitectura del sistema: X64
Paquete: Microsoft.DesktopAppInstaller v1.18.2691.0

Registros: %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir
```

Trenly:
@jmaraujouy - Can you provide the output of `winget list Everything` ?

jmaraujouy:
Yes, sure. ;)

`winget list Everything`

```
Nombre                           Id                        Versión      Disponible Origen
-----------------------------------------------------------------------------------------
Everything 1.4.1.1022 Lite (x64) voidtools.Everything.Lite < 1.4.1.1015 1.4.1.1022 winget
```

jmaraujouy:
I just noticed that the [manifest for v1.4.1.2022](https://github.com/microsoft/winget-pkgs/blob/master/manifests/v/voidtools/Everything/1.4.1.1022/voidtools.Everything.installer.yaml) shows "DisplayName: Everything 1.4.1.**1021**" six times.

It should be 1.4.1.1022.

jmaraujouy:
Hello, @Dragon1573 

Can you please fix the version number? (See my previous message)

I have no idea on how to make a pull request.

Dragon1573:
> Can you guys please fix the version number? (See my previous message)

Em ...

[...]

Issue #70469, opened 2022-08-17
[Package Request]: GPU-Z by TechPowerUp


suratovvlad:
### Package Requested

- [X] I would like someone else to build the manifest.
- [ ] I would like help so I can submit the manifest.
- [X] I have performed a search and couldn't find this package.
- [ ] I have checked that the installer for this package is not contained in a .zip file
- [ ] I think there is a new version available and I have provided the URL.

### Please provide the following information

Publisher: TechPowerUp 
Publisher Url: https://www.techpowerup.com/
Package Name: TechPowerUp.GPU-Z
Description: GPU-Z is a lightweight utility designed to give you all information about your video card and GPU.
Package Version: 2.47.0
Installer URL: https://nl2-dl.techpowerup.com/files/CX_GLRax-0knE2pKIgakyA/1660813451/GPU-Z.2.47.0.exe

Current Package Identifier (for new versions):

suratovvlad:
I see, there was this package https://github.com/microsoft/winget-pkgs/issues/22086, but now it is not available for some reason
![image](https://user-images.githubusercontent.com/3812290/185243369-93d761c7-dfc3-47c4-a2b9-b1075c79e648.png)

russellbanks:
The packages were removed because the installer URLs returned 404. #70481 will revert the latest version. However, it is still 404 so it will need updating once that PR has been accepted.

ItzLevvie:
> However, it is still 404 so it will need updating once that PR has been accepted.

This part is intentional because the developer only allows package managers to download TechPowerUp applications without any sort of expiration dates by taking advantage of WinGet's user agent.

ItzLevvie:
Close with reason: Resolved by the two pull requests above;

PR #95900, opened 2023-02-08
Add Postman 10.9.3

al-lac:

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/95900)

### DIFF
--- a/manifests/p/Postman/Postman/10.9.3/Postman.Postman.installer.yaml
+++ b/manifests/p/Postman/Postman/10.9.3/Postman.Postman.installer.yaml
@@ -20,10 +20,6 @@ Installers:
     SilentWithProgress: -s
   UpgradeBehavior: install
   ProductCode: Postman
-  AppsAndFeaturesEntries:
-  - DisplayName: Postman x86_64 10.9.3
-    DisplayVersion: 10.9.3
-    ProductCode: Postman
 ManifestType: installer
 ManifestVersion: 1.2.0
 
--- a/manifests/p/Postman/Postman/10.9.3/Postman.Postman.installer.yaml
+++ b/manifests/p/Postman/Postman/10.9.3/Postman.Postman.installer.yaml
@@ -1,0 +1,29 @@
+# Created using wingetcreate 1.1.2.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.2.0.schema.json
+
+PackageIdentifier: Postman.Postman
+PackageVersion: 10.9.3
+MinimumOSVersion: 10.0.0.0
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+Installers:
+- InstallerLocale: en-US
+  Architecture: x64
+  InstallerType: exe
+  Scope: user
+  InstallerUrl: https://dl.pstmn.io/download/version/10.9.3/win64
+  InstallerSha256: 9a85c5959449365ad2800bfc5dfa7b8ba0d350bdd7e1602bbab17f1ab01ecda7
+  InstallerSwitches:
+    Silent: -s
+    SilentWithProgress: -s
+  UpgradeBehavior: install
+  ProductCode: Postman
+  AppsAndFeaturesEntries:
+  - DisplayName: Postman x86_64 10.9.3
+    DisplayVersion: 10.9.3
+    ProductCode: Postman
+ManifestType: installer
+ManifestVersion: 1.2.0
+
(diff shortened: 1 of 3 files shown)


Trouffman:
The older version 10.9.0 is no longer available.
Should we consider using the https://dl.pstmn.io/download/latest/win64 in the installer URL instead and use the auto-updater take care of the update?

manifests/p/Postman/Postman/10.9.3/Postman.Postman.installer.yaml
```diff
@@ -0,0 +1,29 @@
...
+  InstallerSha256: 9a85c5959449365ad2800bfc5dfa7b8ba0d350bdd7e1602bbab17f1ab01ecda7
+  InstallerSwitches:
+    Silent: -s
+    SilentWithProgress: -s
+  UpgradeBehavior: install
+  ProductCode: Postman
+  AppsAndFeaturesEntries:
+  - DisplayName: Postman x86_64 10.9.3
+    DisplayVersion: 10.9.3
+    ProductCode: Postman
```
mdanish-kh, line 26:
```suggestion
```
ARP entries especially with DisplayVersion are not needed here as PackageVersion is already equal to DisplayVersion and upgrade matching should work without them as well.

@rodrigoborgesdeoliveira I believe you added the entries in the last two manifests. We shouldn't be using DisplayVersion here as it will cause unnecessary pre-checks (See https://github.com/microsoft/winget-pkgs/pull/65816#issue-1301301147) and here it will also lead to infinite upgrade scenarios for the user if previous manifests do not use this field (See https://github.com/microsoft/winget-pkgs/issues/88726#issuecomment-1398993435).
We should remove them from the previous two manifests as well as they would likely cause the above mentioned infinite upgrades for user.

mdanish-kh (changes requested):

manifests/p/Postman/Postman/10.9.3/Postman.Postman.installer.yaml
rodrigoborgesdeoliveira: Got it. Thanks for the info @mdanish-kh! I'll update the previous manifests later today and post a comment here once those are ready.

mdanish-kh:
> The older version 10.9.0 is no longer available. Should we consider using the https://dl.pstmn.io/download/latest/win64 in the installer URL instead and use the auto-updater take care of the update?

That's not a bad idea if the publisher removes previous versions regularly and retains only the latest one. However, I do see that many Urls for previous versions are working, so there may be value in keeping the previous versions which would not be possible if we were to start using a vanity Url as you suggested.

manifests/p/Postman/Postman/10.9.3/Postman.Postman.installer.yaml
rodrigoborgesdeoliveira: To confirm, is it recommended to not have the `AppsAndFeaturesEntries` in these situations at all or just not have the `DisplayVersion`?

rodrigoborgesdeoliveira:
> That's not a bad idea if the publisher removes previous versions regularly and retains only the latest one. However, I do see that many Urls for previous versions are working, so there may be value in keeping the previous versions which would not be possible if we were to start using a vanity Url as you suggested.

I agree. This is actually the first time I see them removing the previous version. Another interesting behavior from their links is that many patch releases are not directly accessible through a download link, e.g. https://dl.pstmn.io/download/version/10.8.1/win64, only the the minor releases, e.g. https://dl.pstmn.io/download/version/10.8.0/win64.

@mdanish-kh, for version 10.9.0, what would be the side-effect of deleting it from the repository?

mdanish-kh:
> @mdanish-kh, for version 10.9.0, what would be the side-effect of deleting it from the repository?

Not much really. If it's Url is not accessible, it should be removed from the repo. `winget` source will update and users would be getting the previous "latest" version. (in this case `10.8.0` until this PR is merged). 

However, I don't think `10.9.0` should be removed as I can still install it through `winget install postman`. This could be the case similar to https://github.com/microsoft/winget-pkgs/issues/70469#issuecomment-1218684399 where [this URL](https://us2-dl.techpowerup.com/files/GPU-Z.2.52.0.exe) would return 404 from the browser, but you could still get the package through WinGet (winget install gpu-z).

rodrigoborgesdeoliveira:
I can't install version 10.9.0 through winget anymore. Running `winget install Postman.Postman -v 10.9.0` gives me an error.

manifests/p/Postman/Postman/10.9.3/Postman.Postman.installer.yaml
rodrigoborgesdeoliveira: I've opened the PRs for versions 10.8.0 and 10.9.0:
- #95932
- #95934

@mdanish-kh, the URL for version 10.9.0 doesn't work for me anymore even through `winget install`. Please let me know if you can still download it after a few hours have passed. I'm wondering if it has anything to do with CDNs or something similar.

Trenly (approved):

PR #95856, opened 2023-02-08
Location required for AutoHotkey

Trenly:
* Resolves #95829
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/95856)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/95856)

PR #96799, opened 2023-02-15
AutoHotkey: remove elevation requirement for user-scoped installs

stkb:
Moves the ElevationRequirement key from the manifest root to just the machine installers, since it's not necessary for a user-scoped install.

I also removed /user and /elevate switches from the installer commands. I could be wrong and can't find full documentation on the installer switches; but from testing, they don't appear to make any difference if the /silent switch is present.

Also moved the ProductCode key to the manifest root since it's commmon on all installers.

Two other things I attempted but failed at: removing the installation location requirement and properly supporting interactive installs.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/96799)

### DIFF
--- a/manifests/a/AutoHotkey/AutoHotkey/2.0.2/AutoHotkey.AutoHotkey.installer.yaml
+++ b/manifests/a/AutoHotkey/AutoHotkey/2.0.2/AutoHotkey.AutoHotkey.installer.yaml
@@ -8,45 +8,33 @@ InstallModes:
 - interactive
 - silent
 InstallerSwitches:
+  Silent: /silent
+  SilentWithProgress: /silent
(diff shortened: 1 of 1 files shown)


stkb:
Pinging @denelon, @Trenly, @stephengillie 

I assume that, as with #95856, the problem is that pipeline can't enter a location for the install.

stephengillie (approved):

stephengillie:
> I assume that, as with #95856, the problem is that pipeline can't enter a location for the install.

Yes, and I believe it's on the list of features to add.

For (my own) future reference, the install root used here is `C:\ahk` without a trailing slash.

PR #97085, opened 2023-02-17
New version: Suwayomi.Tachidesk-JUI version 1.3.2

Syer10:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/97085)


Syer10:
@microsoft-github-policy-service agree company="Suwayomi"

stephengillie:
The application installs normally but gives a message when run: 
image

Syer10:
That happens when there is no Java 8+ installed on the machine, since the internal server needs it.

stephengillie:
> That usually happens when there is no Java 8+ installed on the machine, since the internal server needs it. It can be caused by other errors but missing Java is the most common issue.

This sounds as though this could use `Oracle.JavaRuntimeEnvironment` as a dependency.

stephengillie (approved):

Syer10:
> This sounds as though this could use `Oracle.JavaRuntimeEnvironment` as a dependency.

There isn't only one implementation of the java runtime, so I don't think I can add java as a dependency unless I am able to tell if they don't have another implementation of Java like Temurin or Azul Zulu.

PR #97312, opened 2023-02-20
New version: ElijahLopez.MusicCaster version 5.13.9

elibroftw:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/97312)


russellbanks (changes requested):
The `4.89.0` submitted with this is a mistake

elibroftw:
That has never happened before, must be because of the new winget releaser not solving the edge case of last major version? I'm surprised I even caught your comment in my emails since I usually auto-delete them.

russellbanks (approved):

Issue #94013, opened 2023-01-16
Not able to found roblox with cli


khewweifeng:
### Brief description of your issue

Not able to found roblox with cli

### Steps to reproduce

winget search roblox 

winget search 9NBLGGGZM6WM

### Expected behavior

able to install roblox with winget cli

### Actual behavior

No package found matching input criteria 

### Environment

```shell
winget --info
Windows Package Manager v1.3.2691
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.22621.1105
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.18.2691.0
```

khewweifeng:
> @stephengille - Roblox cannot be supported.
> 
> See [#47480 (review)](https://github.com/microsoft/winget-pkgs/pull/47480#pullrequestreview-888653322) and the littany of failed PRs - https://github.com/microsoft/winget-pkgs/issues?q=Roblox
> 
> > https://www.xbox.com/en-us/games/store/roblox/9nblgggzm6wm
> 
> Seems like this is a request for adding an Xbox store source, not just msstore

Another link but the id is the same - https://apps.microsoft.com/store/detail/roblox/9NBLGGGZM6WM

rollingmoai:
If you are eager to use Winget, try using https://github.com/pizzaboxer/bloxstrap instead, which is available here:
```	
winget install bloxstrap
```

PR #97854, opened 2023-02-26
Yandex.Browser version 23.1.3.952

Andrej730:

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/97854)
 
 
 Tried to use vanity url so that it will be keep itself up to date automatically.

### DIFF
--- a/manifests/y/Yandex/Browser/23.1.3.952/Yandex.Browser.installer.yaml
+++ b/manifests/y/Yandex/Browser/23.1.3.952/Yandex.Browser.installer.yaml
@@ -9,8 +9,8 @@ InstallerSuccessCodes:
 Installers:
 - Architecture: x64
   InstallerType: exe
-  InstallerUrl: https://browser.yandex.com/download/?full=1&bitness=64&statpromo=false&os=win&lang=en
-  InstallerSha256: 9F8D1DD5FCE61D716C1B92C92A505A3F02C9ACE4D0ABB92426B2DB2A94E95997
+  InstallerUrl: https://cachev2-ams01.cdn.yandex.net/download.cdn.yandex.net/browser/int/23_1_3_952_40699/en/Yandex.exe
+  InstallerSha256: 0f59109bcbea0899798caed9358f4fbec5a78399b9355dee233affc9438bae4a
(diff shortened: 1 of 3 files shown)


Andrej730:
@mdanish-kh hello! 👀 
Downloading installer from the same urls seem to produce different hash because they are adding info about the way it was downloaded at the end...

image

Is it possible to validate the installer some other way? I've found that there is so some signature from publisher that match between files.

image

mdanish-kh:
@Andrej730 Yes, I can confirm I'm seeing different hashes from the application from the InstallerUrl you used in the manifest as it redirects to couple other Urls and adds additional parameters at the end of the Url that are serving a different application?

However, if I use the direct Url from the CDN https://cachev2-ams02.cdn.yandex.net/download.cdn.yandex.net/browser/int/23_1_3_949_40687/en/Yandex.exe, I'm consistently getting the same hash and it doesn't appear to add additional params or redirect to any other Url



```pwsh
 ~/Downloads winget hash .\Yandex.exe
InstallerSha256: 0159a0da24910b6b9cbdf888f82862e76bdc334d467f0156c279a82fe55eb449
```
I'm not sure I follow the motivation behind this PR as you used the direct Url with the correct hash in PR  #97827 👀 The merged package seems to work just fine for me with no hash mismatch errors.

Andrej730:
@mdanish-kh the idea was to use vanity url so that wingetbot could have automatically update the manifest in the future

Is there way to use executable signature to check the match?

mdanish-kh:
Oh right, In that case I believe we're out of luck as WinGet only matches Installer256 hash to determine if it's the same downloaded application.

stephengillie:
This situation seems similar to the Roblox installer. https://github.com/microsoft/winget-pkgs/issues/94013#issuecomment-1386209074

Andrej730:
I guess there is no way to get consistent hash from this vanity url, I switch it to direct url.

mdanish-kh (approved):

PR #97870, opened 2023-02-27
New version: chrisant996.Clink version 1.4.20.236aba

vedantmgoyal9:
### Result: Installation Failed
|             | Manifest                 | Add/Remove Programs                   |
| ----------- | ------------------------ | ------------------------------------- |
| Name        | Clink |     |
| Version     | 1.4.20.236aba     |  |
| Publisher   | Christopher Antos   |       |
| ProductCode |           |     |
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/97870)

### DIFF
--- a/manifests/c/chrisant996/Clink/1.4.20.236aba/chrisant996.Clink.installer.yaml
+++ b/manifests/c/chrisant996/Clink/1.4.20.236aba/chrisant996.Clink.installer.yaml
@@ -1,0 +1,25 @@
+# Created with YamlCreate.ps1 v2.2.0 using InputObject 🤖 $debug=QUSU.CRLF..7-2-9
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.2.0.schema.json
+
+PackageIdentifier: chrisant996.Clink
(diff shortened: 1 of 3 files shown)


stephengillie (approved):

CrushBug:
Hey @vedantmgoyal2009 if clink is auto-updated by itself, it does **not** write the final Git hash to the registry in the above listed Version field. As such, future WinGet upgrade calls result in WinGet thinking there is a newer version when there is not. Here is a sample output of me checking WinGet after Clink updates itself.
```
Name            Id                  Version   Available
Clink v1.4.20   chrisant996.Clink   1.4.20    1.4.20.236aba
```
Is it possible for all future updates to only have the first 3 version number parts? I had originally filed this as a bug with chrisant996 directly, but he said this isn't something that is a problem on his side and he had no idea that Clink had a package in WinGet. I don't think he minds, but since it isn't working quite right I think this is the right place to fix it.

Thanks.

PR #98389, opened 2023-03-03
Tyrrrz.DiscordChatExporter.CLI version 2.39

CanePlayz:

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/98389)


CanePlayz:
Can anyone please confirm that this issue happened because the application doesn't actually have an installer but instead comes just as a .zip file with an executable in it and winget needs an actual installer?

stephengillie:
Hi @CanePlayz,

The package installs normally, but requires `Microsoft.VCRedist.2015+.x64` and apparently `Microsoft.DotNet.DesktopRuntime.7` or a similar package to run:
```
Your system is missing runtime components required by DiscordChatExporter.Cli:
  - Visual C++ Redistributable 2015-2019
  - .NET Runtime (Base) v7.0.0

To install the missing components automatically, set the environment variable DOTNET_ENABLE_BOOTSTRAPPER to true, and then run the application again:
  set DOTNET_ENABLE_BOOTSTRAPPER=true      (Command Prompt)
  $env:DOTNET_ENABLE_BOOTSTRAPPER="true"   (Powershell)
```

It appears that `$env:DOTNET_ENABLE_BOOTSTRAPPER="true"` should be run on first start, before the application, to have the application install these dependencies. After doing so the application runs normally:
```
DiscordChatExporter.Cli v2.39.0

USAGE
  DiscordChatExporter.Cli [options]
  DiscordChatExporter.Cli [command] [...]

OPTIONS
  -h|--help         Shows help text. 
  --version         Shows version information. 

COMMANDS
  exportall         Export all accessible channels. 
  export            Export one or multiple channels. 
  exportdm          Export all direct message channels. 
  exportguild       Export all channels within specified guild. 
  channels          Get the list of channels in a guild. 
  dm                Get the list of direct message channels. 
  guilds            Get the list of accessible guilds. 
  guide             Explains how to obtain token, guild or channel ID. 

You can run `DiscordChatExporter.Cli [command] --help` to show help on a specific command.

```

CanePlayz:
Ok, thanks. So I guess winget will just treat the program run .exe as the installer file and open up the program once as the "installation process".

In regard to the dependencies: winget doesn't support downloading other packages as dependencies, right?

CanePlayz:
> It appears that $env:DOTNET_ENABLE_BOOTSTRAPPER="true" should be run on first start, before the application, to have the application install these dependencies.

Is that something the application needs to change (so that the environment variable is set on the first launch) or something I can change here by adding something to the manifest that will be done on every install?

CanePlayz:
For clarification: Still the old issue, this rerun just happened because I reset some tests.

stephengillie:
> Ok, thanks. So I guess winget will just treat the program run .exe as the installer file and open up the program once as the "installation process".

Part of the validation process runs every executable added to the system. 

> In regard to the dependencies: winget doesn't support downloading other packages as dependencies, right?

Dependencies are not currently supported. Hopefully that will be fixed soon. 

> > It appears that $env:DOTNET_ENABLE_BOOTSTRAPPER="true" should be run on first start, before the application, to have the application install these dependencies.
> 
> Is that something the application needs to change (so that the environment variable is set on the first launch) or something I can change here by adding something to the manifest that will be done on every install?

Yes, it would be ideal to have this set before the program runs. I'm not completely sure, but I believe there might be a `Commands` field where something like this might be specified. If not, it would need to wait for multiple dependency support.

CanePlayz:
Alright, thanks. I don't think anything in the application will be changed so it's best for me to wait until there is dependency support in winget which will fix this problem easily I think.

stephengillie:
Hi @CanePlayz,

Unfortunately, the dependency hasn't been removed, just changed appearance:
image

CanePlayz:
> Hi @CanePlayz,
> 
> Unfortunately, the dependency hasn't been removed, just changed appearance: image

Ok, I'm a bit confused here. Just like in the GUI version, I've changed the installer type to portable, cause well, that's what it is. How does winget handle portables? I thought that it shouldn't have to install anything then, in case it's a portable wrapped inside a zip it just has to unzip the portable. So when does that dialogue pop up?

If that dialogues doesn't pop up during installation but only once the winget installation was successful and you try to use the program, then I don't understand why the GUI passed (it has the same behaviour) or why it should be winget's problem, as the program itself handles dependencies by letting the user know about them.

However, if that pops up during installation, I don't understand what winget does with portables.

stephengillie:
> If that dialogues doesn't pop up during installation but only once the winget installation was successful and you try to use the program, then I don't understand why the GUI passed (it has the same behaviour) or why it should be winget's problem, as the program itself handles dependencies by letting the user know about them.

The program installs as expected, but requires the dependency on running `DiscordChatExporter.Cli.exe`. The issue is that the application doesn't run when executed - instead it runs another installer, which isn't unattended. The program must run promptly when executed, and remain responsive to user input.

CanePlayz:
> > If that dialogues doesn't pop up during installation but only once the winget installation was successful and you try to use the program, then I don't understand why the GUI passed (it has the same behaviour) or why it should be winget's problem, as the program itself handles dependencies by letting the user know about them.
> 
> The program installs as expected, but requires the dependency on running `DiscordChatExporter.Cli.exe`. The issue is that the application doesn't run when executed - instead it runs another installer, which isn't unattended. The program must run promptly when executed, and remain responsive to user input. 

That makes sense, thanks for clarifying. The CLI version has to wait for dependency support then, I guess.

CanePlayz:
Hi @stephengillie, how would I specify the .NET Framework as a Windows feature in the manifest? I think I understood the general layout but I don't know what the identifier would be... (thought I'd pick this up again since dependencies are now supported in the 1.6 release candidate)

stephengillie:
> Hi @stephengillie, how would I specify the .NET Framework as a Windows feature in the manifest? I think I understood the general layout but I don't know what the identifier would be... (thought I'd pick this up again since dependencies are now supported in the 1.6 release candidate)

Hi @CanePlayz,

I believe the .NET Runtime in question is among the public manifests under `Microsoft.DotNet.Runtime.7`.

manifests/t/Tyrrrz/DiscordChatExporter/CLI/2.40.4/Tyrrrz.DiscordChatExporter.CLI.installer.yaml
```diff
@@ -0,0 +1,18 @@
+# Created using wingetcreate 1.2.5.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.5.0.schema.json
+
+PackageIdentifier: Tyrrrz.DiscordChatExporter.CLI
+PackageVersion: 2.40.4
```
stephengillie, line 5:
```suggestion
PackageVersion: 2.40.4
Dependencies:
  PackageDependencies:
  - PackageIdentifier: Microsoft.VCRedist.2015+.x64
  - PackageIdentifier: Microsoft.DotNet.DesktopRuntime.7
```

stephengillie:
Alternatively, if the application will handle installing the dependencies, we can consider this normal program operation:
image

CanePlayz:
> Alternatively, if the application will handle installing the dependencies, we can consider this normal program operation: image

So are you saying there is a way to add to the manifest that a "y" should be entered upon first launch?

stephengillie:
That step would be up to the user.

CanePlayz:
> That step would be up to the user.

I see. So what did you mean by "we can consider this normal program operation:"? What would we be doing differently than in case 1 where we add those components as dependencies and in case 2 where it leads to an executable validation error?

stephengillie:
> > That step would be up to the user.
> 
> I see. So what did you mean by "we can consider this normal program operation:"? What would we be doing differently than in case 1 where we add those components as dependencies and in case 2 where it leads to an executable validation error?

To put this in long form - The package depends on 2 other packages, `Microsoft.VCRedist.2015+.x64` and `Microsoft.DotNet.DesktopRuntime.7`. There are a few ways that these can be installed. 

First, or case 1, these can be installed through the WinGet package manager, as dependencies. This is largely what we have been attempting to accomplish in this thread. And we will accept dependencies again after the 1.6 release. This would be a traditional and accepted path to release the package. 

Second, the installer can install these dependencies. Many developers prefer to use this route, as it gives them the best control over the install process. But this option isn't available to users who are submitting another ISV's package. 

Third, the application can install these dependencies during first run. This is what I meant by "we can consider this normal program operation" - that in the course of normal program operation, it will prompt the user to allow the install, instead of erroring due to an unhandled situation. 

Case 2 would be the second or third options. The Validation-Executable-Error label is sometimes due to a missing dependency, and sometimes because the Automatic Validation Pipeline was designed for GUI applications, and has trouble confirming that CLI applications have launched correctly. 

If you were interested in case 2, then I would apply the waiver now. If you would instead prefer case 1, then I'd apply the waiver after the 1.6 release is ready. I didn't want to apply it prematurely and disrupt your plans.

CanePlayz:
Thanks for the detailed explanation. I think I prefer having the dependencies listed as dependencies in winget.

CanePlayz:
@stephengillie So what's the next step now?

stephengillie:
>  Manifest: The manifest version is not supported. Supported versions: 1.0.0,1.1.0,1.2.0,1.4.0,1.5.0.

manifests/t/Tyrrrz/DiscordChatExporter/CLI/2.41/Tyrrrz.DiscordChatExporter.CLI.locale.en-US.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+PackageIdentifier: Tyrrrz.DiscordChatExporter.CLI
+PackageVersion: 2.41
+PackageLocale: en-US
+Publisher: Tyrrrz
+PackageName: DiscordChatExporter.CLI
+License: GNU
+ShortDescription: Exports Discord chat logs to a file
+ManifestType: defaultLocale
+ManifestVersion: 1.6.0
```
stephengillie, line 12:
```suggestion
ManifestVersion: 1.5.0
```

stephengillie:
Hi @CanePlayz,

With the ManifestVersion set, hopefully this manifest will pass validation. At worst, we should be past any manifest issues and into diagnosing an installer or application issue.

manifests/t/Tyrrrz/DiscordChatExporter/CLI/2.41/Tyrrrz.DiscordChatExporter.CLI.installer.yaml
```diff
@@ -0,0 +1,19 @@
...
+Installers:
+- Architecture: neutral
+  InstallerType: zip
+  NestedInstallerType: portable
+  NestedInstallerFiles:
+  - RelativeFilePath: DiscordChatExporter.Cli.exe
+  InstallerUrl: https://github.com/Tyrrrz/DiscordChatExporter/releases/download/2.41/DiscordChatExporter.Cli.zip
+  InstallerSha256: D47C894C7EFCBAF9932D02482365DCE5AFF188B1D884821995E5DA42F1E718FA
+ManifestType: installer
+ManifestVersion: 1.6.0
```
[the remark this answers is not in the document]
CanePlayz: Thanks

CanePlayz:
How can there be a Validation-Executable-Error when all checks passed?

stephengillie:
Is another dependency needed?
image

CanePlayz:
> Is another dependency needed? This is what pops up when the application is launched.

This is why originally wanted to use WindowsFeatures instead of Dependencies. The application has a bootstrapper that automatically installs .NET 7 and the Visual C++ Redistributable. And the bootstrapper is only dependent on the .NET Framework which is good because that should be installed on every Windows machine. Is it not in the testing environment?

stephengillie:
> This is why I originally wanted to use WindowsFeatures instead of Dependencies. The application has a bootstrapper that automatically installs .NET 7 and the Visual C++ Redistributable. And the bootstrapper itself is only dependent on the .NET Framework which is good because that should be pre-installed on every Windows machine. 
> Edit: However, we should leave the dependencies, because the bootstrapper requires a confirmation to install the dependencies which is not allowed in winget.

As the developer, these are your choices to make. 

> Is [the .NET Framework] not in the testing environment?

Correct, the [Manual Validation environment](https://github.com/microsoft/winget-pkgs/issues/107639#issuecomment-1734683940) has few applications installed, as a fresh Windows install. I'm unsure about exactly what is installed in the Automated Validation VMs, but the same goal should yield a similar reduced list of applications.

CanePlayz:
> As the developer, these are your choices to make.

Unfortunately, I am not the developer.

I'll talk to the dev about the issue with .NET framework to find out what exact version it needs.

CanePlayz:
Okay, the bootstrapper is dependent on .NET Framework 3.5 because it is also supposed to work on Windows 7, which doesn't come with .NET Framework 4.x out of the box. That means we'll also have to wait for WindowsFeatures to enter the stable release.

stephengillie:
> Okay, the bootstrapper is dependent on .NET Framework 3.5 because it is also supposed to work on Windows 7, which doesn't come with .NET Framework 4.x out of the box. That means we'll also have to wait for WindowsFeatures to enter the stable release.
> 
> It's a bit of a bizarre situation because we only need the WIndowsFeature so that the bootstrapper can work and do nothing because the dependencies have already been installed, but it's our only way to make this work.

Hi @CanePlayz,

Sorry for the delayed reply. An issue here is that the .NET Framework 3.5 isn't really a package that can be downloaded and installed - it was meant to be installed through the Server Manager, often off the original Windows CD. (It uses the .NET Common Language Runtime (CLR) version 2.0, for what it's worth.) Unfortunately, this dependency is not currently available through the WinGet package manager. 

Windows 7 and above should be able to install the .NET Framework 4.5.1 and above, which is available through WinGet as `Microsoft.DotNet.Framework.DeveloperPack.4.5` (and the similar `Microsoft.DotNet.Framework.DeveloperPack.4` - these use .NET CLR 4.0.) If the developers are able to increase the version, it might allow this manifest to become accepted into the public manifests.

CanePlayz:
@stephengillie Unfortunately, the version of the bootstrapper won't be increased, since it's supposed to work out of the box even on Windows 7.

Regrading .NET Framework 3.5: Isn't that one of the WindowsFeatures that we will be able to enable via the new manifests? Or shouldn't it be pre-installed anyways on every Windows machine?

stephengillie:
> Regrading .NET Framework 3.5: Isn't that one of the WindowsFeatures that we will be able to enable via the new manifests? Or shouldn't it be pre-installed anyways on every Windows machine?

I just tested with a Windows 11 VM, and unfortunately the .NET Framework 3.5 prompt still appears. I know the `VCRedist` framework is included with Windows 11, but apparently .NET Framework 3.5 isn't. I'm unsure about using WindowsFeatures as Dependencies.

CanePlayz:
Just talked to the dev, and apparently the bootstrapper also works on 4.x. BUT shouldn't that be pre-installed and activated on Windows?

CanePlayz:
Let's sum this up:

We have a bootstrapper that can install the dependencies of this application for us. However, this won't work, because for the bootstrapper to install the packages, an environment variable would need to be set, or 'y' would have to be typed. That means we have to install the dependencies via the dependencies field.

However, there seems to be an issue with a certain version of the .NET Framework being needed. The cause of that has to be the bootstrapper, which will be launched to check for the dependencies even if they are already installed, as it relies on .NET Framework 3.x or 4.x. That prompt really confuses me as .NET Framework 4.x should be pre-installed on new Windows installs, enabling the bootstrapper to launch properly.

CanePlayz:
Okay, so this is the same error as before. Is this error indicating that our prompt popped up again?

manifests/t/Tyrrrz/DiscordChatExporter/CLI/2.42.6/Tyrrrz.DiscordChatExporter.CLI.locale.en-US.yaml
```diff
@@ -0,0 +1,12 @@
+# Created using wingetcreate 1.2.5.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.5.0.schema.json
+
+PackageIdentifier: Tyrrrz.DiscordChatExporter.CLI
+PackageVersion: 2.42.6
+PackageLocale: en-US
+Publisher: Tyrrrz
+PackageName: DiscordChatExporter.CLI
+License: GNU
```
Tyrrrz, line 9: The license is MIT now

CanePlayz: Thanks, this PR has been around quite a while now 🥲

manifests/t/Tyrrrz/DiscordChatExporter/CLI/2.42.6/Tyrrrz.DiscordChatExporter.CLI.installer.yaml
```diff
@@ -0,0 +1,20 @@
+# Created using wingetcreate 1.2.5.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.5.0.schema.json
+
+PackageIdentifier: Tyrrrz.DiscordChatExporter.CLI
+PackageVersion: 2.42.6
+Dependencies:
+  PackageDependencies:
+    - PackageIdentifier: Microsoft.VCRedist.2015+.x64
+    - PackageIdentifier: Microsoft.DotNet.DesktopRuntime.8
+    - PackageIdentifier: Microsoft.DotNet.Framework.DeveloperPack.4.5
```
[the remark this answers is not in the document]
CanePlayz: Alright thanks, I'll remove the VCRedist then. I added the .NET Framework 4.x because of this weird prompt appearing, I'll see which ones I need after some tests.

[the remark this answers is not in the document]
CanePlayz: Can confirm now that VCRedist is not needed explicitly. I got this info:

image

and installed the .NET 8 runtime and it's working now.

CanePlayz:
As my tests revealed and confirmed by @Tyrrrz, the package works completely fine after just installing the .NET 8 runtime, no need for any .NET framework version which is not included in Windows by default.

@stephengillie: Could we wave this through manually?

CanePlayz:
No way it finally passed 🤯 Thanks @stephengillie . I owe you one haha.

stephengillie (approved):

stephengillie:
I forgot to check the symlink before approving. This package works when launched from the full install location, but the symlink can't find the .NET Framework 3.5 files, and requests to install it.

PR #98648, opened 2023-03-06
New version: FlightGear.FlightGear version 2020.3.17

The-Fl3dd0x:
- [x] Have you [validated](https://github.com/microsoft/winget-pkgs/blob/master/AUTHORING_MANIFESTS.md#validation) your manifest locally with `winget validate --manifest manifests\f\FlightGear\FlightGear\2020.3.17`?
- [x] Have you tested your manifest locally with `winget install --manifest manifests\f\FlightGear\FlightGear\2020.3.17`?

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/98648)

### DIFF
--- a/manifests/f/FlightGear/FlightGear/2020.3.17/FlightGear.FlightGear.installer.yaml
+++ b/manifests/f/FlightGear/FlightGear/2020.3.17/FlightGear.FlightGear.installer.yaml
@@ -1,0 +1,16 @@
+# Created with YamlCreate.ps1 v2.2.3 $debug=NVS1.CRLF.5-1-22621-963.Win32NT
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.4.0.schema.json
+
+PackageIdentifier: FlightGear.FlightGear
+PackageVersion: 2020.3.17
+MinimumOSVersion: 10.0.0.0
+InstallerType: inno
+Scope: machine
+UpgradeBehavior: uninstallPrevious
+ReleaseDate: 2022-11-22
+Installers:
+- Architecture: x64
+  InstallerUrl: https://sourceforge.net/projects/flightgear/files/release-2020.3/FlightGear-2020.3.17.exe/download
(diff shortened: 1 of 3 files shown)


The-Fl3dd0x:

stephengillie:
This package is 1.77 GB and the download appears throttled, as it downloads significantly faster through Edge. These are complicating validation and causing it to take longer than usual.

The-Fl3dd0x:
I wonder why the other versions worked, but thanks for looking into it!

stephengillie:
> I wonder why the other versions worked, but thanks for looking into it!

Oh - I failed to note that this is an upgrade for a package already among the public manifests. Apologies.

The-Fl3dd0x:
No problem, it seems to an issue with the installer though, as other people are also missing the MSVC Runtimes. I'll try it with the next version, but test it in the sandbox beforehand.

The-Fl3dd0x:
Maybe providing a mirror instead of the main download link helps, with this one I get 22 MByte/s: https://deac-riga.dl.sourceforge.net/project/flightgear/release-2020.3/FlightGear-2020.3.17.exe

The-Fl3dd0x:
Alternatively, there also is a very small (60.5 MB) web installer

stephengillie:
The manifest would need to be updated with them.

stephengillie (approved):

stephengillie:
After preinstalling the dependency and leaving the download running overnight, I was able to validate the package. This is apparently just an update package and stub to download the main files - which is apparently normal package operation. 
image

The only remaining concern is that the dependency wouldn't automatically install for users installing this upgrade. Once dependency support is in place, this won't be a problem.

The-Fl3dd0x:
Great, is the dependency issue resolved?

The-Fl3dd0x:
I think this can be closed, as there is a newer version merged.

PR #101229, opened 2023-03-31
New version: SomePythonThings.WingetUIStore version 1.6.5

marticliment:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/101229)


stephengillie:
I'm not able to find the cause for this error. It installs and runs normally on a Windows 11 VM. Pipeline logs don't show an error.

marticliment:
I have filed a report regarding the false-positive installer

ppvnf:
is version 1.7.1 also flagged by smartscreen?

stephengillie:
> is version 1.7.1 also flagged by smartscreen?

The only way I know to test is to submit. I believe this usually only blocks specific packages, so other versions on the same host might be unaffected.

marticliment:
I think so, since sometimes I get the Unknown app popup.

I have submitted the installer twice, but they won't accept it as a false positive

Trenly:
Stale PR; Please update or close
[Policy] Needs-Author-Feedback

PR #101378, opened 2023-04-02
New version: Taiko2k.TauonMusicBox version 7.6.3

Taiko2k:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/101378)

### DIFF
--- a/manifests/t/Taiko2k/TauonMusicBox/7.6.3/Taiko2k.TauonMusicBox.installer.yaml
+++ b/manifests/t/Taiko2k/TauonMusicBox/7.6.3/Taiko2k.TauonMusicBox.installer.yaml
@@ -1,0 +1,26 @@
+# Created with YamlCreate.ps1 v2.2.1 using InputObject 🤖 $debug=QUSU.CRLF.7-2-10
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.2.0.schema.json
+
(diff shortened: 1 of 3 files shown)


BrandonWanHuanSheng:
Does it have x86?

Taiko2k:
@BrandonWanHuanSheng I only build a 64 bit build.

BrandonWanHuanSheng:
It does not have x86. I visited that page "https://github.com/Taiko2k/TauonMusicBox/releases"

Taiko2k:
@BrandonWanHuanSheng Its x64 as in x86-64

BrandonWanHuanSheng:
@Taiko2k But winget will not show up x86 into x64. If you do like that. This one most likely choose Neutral.

BrandonWanHuanSheng:
This is a problem with winget client! ⚠️ Winget does not show up x86 instead x64 Installer.
Screenshot (127)
When there is x64 installer only.

Taiko2k:
I can rename it to use x86-64, I assume I will need to re-run the action?

BrandonWanHuanSheng:
If you rename it, I think it might unable on x64. But I don't recommend to do that.

stephengillie (approved):

PR #97850, opened 2023-02-26
Update `PortableCommandAlias` to an optional field

sitiom:
`PortableCommandAlias` is actually an optional field, and the manifest successfully validates without it.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/97850)

PR #101383, opened 2023-04-02
Make `PortableCommandAlias` optional in YamlCreate

mdanish-kh:
This PR is a follow-up of #97850 that updates YamlCreate to treat PortableCommandAlias as an optional field.

cc @Trenly, @denelon

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/101383)

### DIFF
--- a/Tools/YamlCreate.ps1
+++ b/Tools/YamlCreate.ps1
@@ -757,7 +757,6 @@ Function Read-NestedInstaller {
           if (Test-String -not "$($_InstallerFile['PortableCommandAlias'])" -IsNull) { Write-Host -ForegroundColor 'DarkGray' "Old Variable: $($_InstallerFile['PortableCommandAlias'])" }
           $_Alias = Read-Host -Prompt 'PortableCommandAlias' | TrimString
           if (Test-String -not $_Alias -IsNull) { $_InstallerFile['PortableCommandAlias'] = $_Alias }
-          elseif (Test-String -not $_InstallerFile['PortableCommandAlias'] -IsNull) { $_InstallerFile['PortableCommandAlias'] = "$($_InstallerFile['PortableCommandAlias'])" }
 
           if (Test-String $_InstallerFile['PortableCommandAlias'] -MinLength $Patterns.PortableCommandAliasMinLength -MaxLength $Patterns.PortableCommandAliasMaxLength -AllowNull) {
             $script:_returnValue = [ReturnValue]::Success()
--- a/Tools/YamlCreate.ps1
+++ b/Tools/YamlCreate.ps1
@@ -753,17 +753,18 @@ Function Read-NestedInstaller {
       if ($_EffectiveType -eq 'portable') {
         do {
           Write-Host -ForegroundColor 'Red' $script:_returnValue.ErrorString()
-          Write-Host -ForegroundColor 'Green' -Object '[Required] Enter the portable command alias'
-          if (Test-String -not $_Alias -IsNull) { Write-Host -ForegroundColor 'DarkGray' "Old Variable: $_Alias" }
+          Write-Host -ForegroundColor 'Yellow' -Object '[Optional] Enter the portable command alias'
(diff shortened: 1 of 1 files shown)


Tools/YamlCreate.ps1
```diff
@@ -753,17 +753,18 @@ Function Read-NestedInstaller {
       if ($_EffectiveType -eq 'portable') {
         do {
           Write-Host -ForegroundColor 'Red' $script:_returnValue.ErrorString()
-          Write-Host -ForegroundColor 'Green' -Object '[Required] Enter the portable command alias'
-          if (Test-String -not $_Alias -IsNull) { Write-Host -ForegroundColor 'DarkGray' "Old Variable: $_Alias" }
+          Write-Host -ForegroundColor 'Yellow' -Object '[Optional] Enter the portable command alias'
+          if (Test-String -not "$($_InstallerFile['PortableCommandAlias'])" -IsNull) { Write-Host -ForegroundColor 'DarkGray' "Old Variable: $($_InstallerFile['PortableCommandAlias'])" }
           $_Alias = Read-Host -Prompt 'PortableCommandAlias' | TrimString
           if (Test-String -not $_Alias -IsNull) { $_InstallerFile['PortableCommandAlias'] = $_Alias }
+          elseif (Test-String -not $_InstallerFile['PortableCommandAlias'] -IsNull) { $_InstallerFile['PortableCommandAlias'] = "$($_InstallerFile['PortableCommandAlias'])" }
```
Trenly, line 760: This line doesn't make sense to me. Reading through the logic . . .
If the `InstallerFile.PortableCommandAlias` is not null, set `InstallerFile.PortableCommandAlias` to `InstallerFile.PortableCommandAlias`

Tools/YamlCreate.ps1
```diff
@@ -753,17 +753,18 @@ Function Read-NestedInstaller {
...
 
-          if (Test-String $_Alias -MinLength $Patterns.PortableCommandAliasMinLength -MaxLength $Patterns.PortableCommandAliasMaxLength) {
+          if (Test-String $_InstallerFile['PortableCommandAlias'] -MinLength $Patterns.PortableCommandAliasMinLength -MaxLength $Patterns.PortableCommandAliasMaxLength -AllowNull) {
             $script:_returnValue = [ReturnValue]::Success()
           } else {
             $script:_returnValue = [ReturnValue]::LengthError($Patterns.PortableCommandAliasMinLength, $Patterns.PortableCommandAliasMaxLength)
           }
-          if ($_Alias -in @($_NestedInstallerFiles.PortableCommandAlias)) {
+          if ("$($_InstallerFile['PortableCommandAlias'])" -in @($_NestedInstallerFiles.PortableCommandAlias)) {
```
Trenly, line 767: This may throw an error if the alias exists and isn't changed, since the `InstallerFile.PortableCommandAlias` could be existing in the portable command aliases already in the `NestedInstallerFiles`.  The validation was on the user input to specifically avoid any possibility of a collision with existing data

mdanish-kh: I'm not sure I understand the concern. Don't we want to stop the user from entering a duplicate in that case?
I'm thinking about the following scenario:
1) User types a duplicate portable alias (one that already is in NestedInstallerFiles from previous portable entry) e.g., `test`. This gets written into InstallerFile as it's not a null input.
2) Alias Collision error thrown. User presses <kbd>Enter</kbd> and *null** value is passed which doesn't get written to InstallerFile. If $_Alias check is used, InstallerFile would hold the value `test`, the script would return success as $_Alias being null is not in NestedInstallerFiles and the value would get written into the manifest. This would result in a manifest that fails `winget validate` with duplicate portable aliases. The check currently used in the PR forces the user to enter a non-duplicate alias.

stephengillie (approved):

PR #101429, opened 2023-04-03
Strem.App version 0.7.226

grofit:

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/101429)


grofit:

stephengillie:
This application is very picky about its launch location, and errors unless the current location is:
> C:\Users\User\AppData\Local\Microsoft\WinGet\Packages\Strem.App__DefaultSource

This means that:
- In PowerShell and Windows Explorer, launching by path does not work, and instead one has to change directory into this directory before launching the application.
- Symlinks don't work because this is like launching by path.
- The application adds both a "Plugins" and a "logs" folder to the current location. It then fails to launch, logging that `strem.ico` is missing, with no pop-up error message. 
- Shortcuts only work when their `Start in` path matches the install location. (above)

This behavior might be mistaken for non-functionality by a novice user.

grofit:
Yeah I am seeing what is possible as I just used the tool originally to make this assuming it would just take the exe and run that as a shortcut or something but alas it just puts it in a dir, I think I may have to hold out until the issue around start shortcuts from portable apps progresses (on mobile client so can't get issue number right now).

The app assumed that cwd would be same as exe location, so it's probably a simple fix to change logic to use assembly location in dotnet to solve the immediate issue, but if the user experience for launching would be to use command line I don't think it's going to offer much over existing deployment model.

Will still try to get it working thoughto improve upon later when/if start menu item support is added.

grofit:
This should now work as expected as the underlying lib has has the path handling changed to ensure it looks at the app directory not the `cwd` for running, so using `strem` on command line locally runs as expected for this build now.

stephengillie (approved):

PR #102611, opened 2023-04-16
New version: pizzaboxer.Bloxstrap version 2.1.1

pizzaboxer:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/102611)

### DIFF
--- a/manifests/p/pizzaboxer/Bloxstrap/2.1.1/pizzaboxer.Bloxstrap.installer.yaml
+++ b/manifests/p/pizzaboxer/Bloxstrap/2.1.1/pizzaboxer.Bloxstrap.installer.yaml
@@ -1,0 +1,32 @@
+# Created with WinGet Releaser v2 using Komac v1.4.1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.4.0.schema.json
+
+PackageIdentifier: pizzaboxer.Bloxstrap
+PackageVersion: 2.1.1
+InstallerType: exe
+Scope: user
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+InstallerSwitches:
+  Silent: -quiet -nolaunch
+  SilentWithProgress: -quiet -nolaunch
+  Upgrade: -upgrade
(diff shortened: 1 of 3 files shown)


pizzaboxer:
Requires .NET 6 desktop runtime as a dependency, see https://github.com/microsoft/winget-pkgs/issues/92827#issuecomment-1369993178.

image

stephengillie:
Automatic Validation failed with:
InstallationVerification-x86-Scope_Undefined-Locale_Undefined-ErrorScreenshot_20230416021120

pizzaboxer:
> Even with the dependency preinstalled, the installer asks to install it. image

It doesn't really make sense that the installer asks to install it when it's preinstalled, since in my screenshot above I had the runtime preinstalled, and installing with Winget was completely fine. I even tried it in a VM to be sure, and it definitely works. Besides, the runtime check is done by .NET itself and not by me - if it failed to detect itself when it's really installed then I think there'd be a wider problem than just this. Can I see which version of the desktop runtime is installed?

image

pizzaboxer:
> x86 Automatic Validation failed with: InstallationVerification-x86-Scope_Undefined-Locale_Undefined-ErrorScreenshot_20230416021120
> 
> x64 validation failed for the `x64` architecture version of the dependency.


I noticed this myself when testing the install in a VM, but Winget doesn't even try to install the runtime even though it's explicitly specified as a dependency in the installer YAML for `Microsoft.DotNet.DesktopRuntime.6`. I know it was mentioned that dependency support is an experimental feature, but has it always been like this? I'm just wondering why this is only a problem now as nothing relating to this has changed between the previous version and now.

sitiom:
This happens every time. Is validation not installing the correct .NET Runtime arch?

pizzaboxer:
> This happens every time. Is validation not installing the correct .NET Runtime arch?

See https://github.com/microsoft/winget-pkgs/pull/103098#issuecomment-1518421875, seems like it's being worked out.

Issue #94840, opened 2023-01-27
[Package Issue]: Electrum.Electrum


W3LLD0NE:
### Please confirm these before moving forward

- [X] I have searched for my issue and not found a work-in-progress/duplicate/resolved issue.
- [X] I have not been informed if the issue is resolved in a preview version of the winget client.

### Category of the issue

No applicable update found.

### Brief description of your issue

Problem:
![image](https://user-images.githubusercontent.com/14851973/215186348-5305960a-373b-40b2-b8f9-626c0608c724.png)

Please note that Electrum was initially installed via their site, not via WinGet.

### Steps to reproduce

Install Electrum v4.3.1 from [their site](https://download.electrum.org/4.3.1/electrum-4.3.1-setup.exe) and try to update it via WinGet.

### Actual behavior

"No applicable upgrade found." message displayed.

### Expected behavior

Package has been updated.

### Environment

```shell
Windows Package Manager v1.4.10173
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.19045.2486
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.19.10173.0
```

### Screenshots and Logs

2023-01-27 22:10:02.842 [CORE] WinGet, version [1.4.10173], activity [{7C9341D2-AE8A-4A82-A5DB-9882988C2B66}]
2023-01-27 22:10:02.843 [CORE] OS: Windows.Desktop v10.0.19045.2486
2023-01-27 22:10:02.843 [CORE] Command line Args: "C:\Users\user\AppData\Local\Microsoft\WindowsApps\winget.exe" update Electrum.Electrum
2023-01-27 22:10:02.843 [CORE] Package: Microsoft.DesktopAppInstaller v1.19.10173.0
2023-01-27 22:10:02.843 [CORE] IsCOMCall:0; Caller: winget-cli
2023-01-27 22:10:02.845 [CLI ] WinGet invoked with arguments: 'update' 'Electrum.Electrum'
2023-01-27 22:10:02.845 [CLI ] Found subcommand: update
2023-01-27 22:10:02.845 [CLI ] Leaf command to execute: root:upgrade
2023-01-27 22:10:02.848 [CLI ] Executing command: upgrade
2023-01-27 22:10:02.849 [REPO] GetCurrentSourceRefs: Source named 'microsoft.builtin.desktop.frameworks' from origin Default is hidden and is dropped.
[...]

PR #103057, opened 2023-04-21
Add scope and ElevationRequirement: Codeblocks.Codeblocks

stephengillie:

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/103057)

### DIFF
--- a/manifests/c/Codeblocks/Codeblocks/20.03/Codeblocks.Codeblocks.installer.yaml
+++ b/manifests/c/Codeblocks/Codeblocks/20.03/Codeblocks.Codeblocks.installer.yaml
@@ -7,7 +7,9 @@ MinimumOSVersion: 10.0.0.0
 InstallerType: nullsoft
 Installers:
 - Architecture: x64
+  Scope: machine
   InstallerUrl: https://sourceforge.net/projects/codeblocks/files/Binaries/20.03/Windows/codeblocks-20.03mingw-setup.exe/download
   InstallerSha256: 862479A6D45B8A36FFC09FC4C63CD0850155088E33BD8CAD9075FB6FCB1F43C2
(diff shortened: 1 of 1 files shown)


manifests/c/Codeblocks/Codeblocks/20.03/Codeblocks.Codeblocks.installer.yaml
```diff
@@ -7,7 +7,9 @@ MinimumOSVersion: 10.0.0.0
 InstallerType: nullsoft
 Installers:
 - Architecture: x64
+  Scope: machine
```
mdanish-kh, line 10:
```suggestion
```

The installation is machine-wide as it installs to `C:\Program Files`, but the installer writes its uninstall entries to `HKEY_CURRENT_USER` which will make WinGet think it's a user scoped installation in the upgrade scenario and subsequent upgrades of this package will error out with `No applicable upgrade found` if we add `Scope: machine` in the manifest.

This is a similar scenario to https://github.com/microsoft/winget-pkgs/issues/94840 and the proper way to resolve this would be to have the publishers fix the installer to write the uninstall entries in the correct place.

mdanish-kh (changes requested):

mdanish-kh:
cc @stephengillie Does this need to be re-opened?

stephengillie:
> cc @stephengillie Does this need to be re-opened?

Probably not. This was part of a project to prevent UAC prompts from interfering with Dev Home, by adding the `elevatesSelf` parameter, and specifying scope where necessary. Since scope is incompatible with this application currently, the need for this individual change seems to have disappeared.

PR #103098, opened 2023-04-21
New version: pizzaboxer.Bloxstrap version 2.2.0

pizzaboxer:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/103098)

### DIFF
--- a/manifests/p/pizzaboxer/Bloxstrap/2.2.0/pizzaboxer.Bloxstrap.installer.yaml
+++ b/manifests/p/pizzaboxer/Bloxstrap/2.2.0/pizzaboxer.Bloxstrap.installer.yaml
@@ -1,0 +1,32 @@
+# Created with WinGet Releaser v2 using Komac v1.4.1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.4.0.schema.json
+
+PackageIdentifier: pizzaboxer.Bloxstrap
+PackageVersion: 2.2.0
+InstallerType: exe
+Scope: user
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+InstallerSwitches:
+  Silent: -quiet -nolaunch
+  SilentWithProgress: -quiet -nolaunch
+  Upgrade: -upgrade
+UpgradeBehavior: install
+Protocols:
+- roblox
+- roblox-player
+Dependencies:
+  PackageDependencies:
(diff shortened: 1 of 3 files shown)


pizzaboxer:
Defer this PR until #102611 has some resolution.

stephengillie:
The issue in previous PRs is that the dependency needs to be `x86`, but the `x64` version is installed by default. Not sure how to specify dependency architecture. Would it be possible to upgrade the package to use the `x64` dependency?

pizzaboxer:
> The issue in previous PRs is that the dependency needs to be `x86`, but the `x64` version is installed by default. Not sure how to specify dependency architecture. Would it be possible to upgrade the package to use the `x64` dependency?

Oh, I didn't realize that. The problem is that I publish both x64 and x86 versions of the application (for now), so I guess I could choose the option of not publishing x86 versions to Winget?

stephengillie (approved):

stephengillie:
> > The issue in previous PRs is that the dependency needs to be `x86`, but the `x64` version is installed by default. Not sure how to specify dependency architecture. Would it be possible to upgrade the package to use the `x64` dependency?
> 
> Oh, I didn't realize that. The problem is that I publish both x64 and x86 versions of the application (for now), so I guess I could choose the option of not publishing x86 versions to Winget?

Thinking on this again, the x86 version needs to use the x86 dependency to work on 32-bit OSes. This is an issue with our validating, not your package. I'll work to solve this in my processes.

pizzaboxer:
> > > The issue in previous PRs is that the dependency needs to be `x86`, but the `x64` version is installed by default. Not sure how to specify dependency architecture. Would it be possible to upgrade the package to use the `x64` dependency?
> > 
> > 
> > Oh, I didn't realize that. The problem is that I publish both x64 and x86 versions of the application (for now), so I guess I could choose the option of not publishing x86 versions to Winget?
> 
> Thinking on this again, the x86 version needs to use the x86 dependency to work on 32-bit OSes. This is an issue with our validating, not your package. I'll work to solve this in my processes.

Thank you. I appreciate the response, though I wouldn't mind only publishing the x64 version since x86 usage is very small (about 4%) from my stats, and it's due to be removed very soon anyway.

PR #103105, opened 2023-04-21
Trigger a bot reply for Zip-Binary label

mdanish-kh:
As per @Trenly's request to add an autoresponder trigger event for https://github.com/microsoft/winget-pkgs/labels/zip-binary

cc @denelon 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/103105)

PR #101250, opened 2023-03-31
Respond with comments when moderators add Dependencies, Interactive, or Hardware Labels

Trenly:
Currently, when moderators add the dependencies label to a PR, the label doesn't trigger the autoresponder to add the message describing what the dependencies label means. This change should allow fabricbot to process the event label triggers even if they were kicked off by fabricbot

cc @denelon
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/101250)

PR #103148, opened 2023-04-22
Trigger label description replies in issues

mdanish-kh:
The reply triggers added in https://github.com/microsoft/winget-pkgs/pull/101250 and https://github.com/microsoft/winget-pkgs/pull/103105 only work when the label is added in a PR. This PR (hopefully) updates the behavior to reply with label description in issues as well.

cc @Trenly

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/103148)

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -6098,7 +6098,7 @@
           "issues",
           "project_card"
         ],
-        "taskName": "Blocked by Dependencies",
+        "taskName": "PR blocked by Dependencies",
         "dangerZone": {
           "respondToBotActions": true,
           "acceptRespondToBotActions": true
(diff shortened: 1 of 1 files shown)


Trenly (approved):

Trenly:
@denelon please

denelon (approved):

denelon:
Please test :)

mdanish-kh:
Looks like it's not working in https://github.com/microsoft/winget-pkgs/issues/103209, will have to revert this commit 😔

mdanish-kh:
FWIW - instead of binding the reply to an `addLabel` event, we could make it based off of a comment trigger, like how `Duplicate of #<>` works (as this does work for both PRs and Issues). This would work for the [Policy] comments but not if an admin manually attaches the label.

But I guess this is a low ask (and I don't want to keep disturbing denelon for the uncertain PRs xD)

PR #103157, opened 2023-04-22
Fixing ARP for Winamp.Winamp

krystofkrticka:
According to appwiz.cpl in Windows version number in AppsAndFeaturesEntries should be 5.91.0 
image

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/103157)

### DIFF
--- a/manifests/w/Winamp/Winamp/5.91.0.10029/Winamp.Winamp.installer.yaml
+++ b/manifests/w/Winamp/Winamp/5.91.0.10029/Winamp.Winamp.installer.yaml
@@ -18,7 +18,7 @@ Installers:
   AppsAndFeaturesEntries:
     - DisplayName: "Winamp"
       Publisher: "Winamp SA"
-      DisplayVersion: "5.9.1"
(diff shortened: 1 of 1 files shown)


krystofkrticka:
For some reason dependency VC++ waits until you click on install.
https://imgur.com/a/OC4Ye16

stephengillie:
This is due to Winamp having a dependency.

krystofkrticka:
@stephengillie I know but weird thing is that dependency suddenly fails to install silently even though it installs just fine by itself. I thought and expected that it would use install parameters stated in manifest of dependency.

stephengillie:
Oh, dependencies aren't currently supported, pending:
* https://github.com/microsoft/winget-cli/issues/163

krystofkrticka:
I've known about this, but I was little bit surprised that stable version of winget tried to installed dependencies which are still in works.

stephengillie (approved):

PR #103468, opened 2023-04-25
Fix CCleaner Pro Trial issue

Masamune3210:
Currently, there is a issue with winget trying to upgrade the normal ccleaner to the pro trial due to the ProTrial folder being in the ccleaner folder, so I moved it out into the folder with all the other programs and added the real 6.11 manifest for ccleaner

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/103468)


Masamune3210:
Do I need to split these out or can this be manually merged since all I did was move the ProTrial folder, copy the 6.10 manifest for ccleaner and then put them in the 6.11 folder with the correct version and hash for the new update

BrandonWanHuanSheng:
It look like you not need to delete my file. It look like the manifests isn't created on  "manifests\p\Piriform\CCleaner\6.11" So, why the update is not available.

BrandonWanHuanSheng:
Did you noticed?
Screenshot (208)
Do not delete Piriform.CCleaner.ProTrial because Piriform.CCleaner manifest version was not up to date.

Masamune3210:
Part of this pr is correctly adding the 6.11 installer for the non-pro version of CCleaner. Having the trial in the main CCleaner folder is confusing winget into thinking that it is the 6.11 manifest. You can check yourself if you want, currently, trying to update 6.10 or below to what winget thinks is 6.11 but is actually the pro trial will fail.

Masamune3210:
Also, didn't delete the pro trial manifest. If you look at the change log, it was moved out of the main folder, not deleted

BrandonWanHuanSheng:
If u add the new manifest, it will overlap with Piriform.CCleaner.

BrandonWanHuanSheng:
> Part of this pr is correctly adding the 6.11 installer for the non-pro version of CCleaner. Having the trial in the main CCleaner folder is confusing winget into thinking that it is the 6.11 manifest. You can check yourself if you want, currently, trying to update 6.10 or below to what winget thinks is 6.11 but is actually the pro trial will fail. 

It will not failed but it enable pro version. I think that pro version might need to put them together.

Masamune3210:
It very much does. What prompted me to do this pr in the first place was the fact that I tried to use winget to update to 6.11. Winget sees 6.11 as available due to the protrial manifest but 
A. The pro trial is not the correct package to offer to someone who isn't wanting to use pro, the correct package is the 6.11 installer for the normal version of CCleaner
B. It doesn't matter anyway because if you try to do "winget upgrade piriform.ccleaner", it just comes back with no available updates could be found.

BrandonWanHuanSheng:
> Also, didn't delete the pro trial manifest. If you look at the change log, it was moved out of the main folder, not deleted

I think, I seeing you as deleted.

PR #103590, opened 2023-04-26
Remove: GoLang.Go.1.20 version 1.20.0

russellbanks:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/103590)


stephengillie (approved):

staticfrost:
@stephengillie why was this commit allowed, it changed the whole structure of the Go versioning?

stephengillie:
> @stephengillie why was this commit allowed, it changed the whole structure of the Go versioning?

From the first commit by @russellbanks: 
> ## Replaced by `GoLang.Go` as splitting is no longer necessary with package pinning

[Package Pinning](https://github.com/microsoft/winget-cli/blob/master/doc/specs/%23476%20-%20Package%20Pinning.md) prevents a package from being upgraded unless it's explicitly specified, or unless `--include-pinned` is specified.

russellbanks:
> @stephengillie why was this commit allowed, it changed the whole structure of the Go versioning?

Package pinning never used to exist so we split packages by major versions if older major versions were still getting updates (think Go 1.19 still getting updates), to emulate update channels. The downside to this is that it violates the good-practice of packages being able to be installed side-by-side and it also means that users on Go 1.19, for example, won't get an upgrade to Go 1.20 unless they explicitly upgrade.

The solution is package pinning where a package can be pinned to a version such as `1.19.x`; the user won't get updates greater than that. This means that this temporary solution is no longer needed.

PR #106757, opened 2023-05-14
Ferdium.Ferdium.Nightly version 6.3.0-nightly.5

Alphrag:

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/106757)


Alphrag:
If I understand correctly the validation fails on the analysis of the application which is linked to some antivirus, is that right?
Where could I find a detail of the problem to solve it, since I can't find much information in the logs to explain why the `arm64` version fails when both the `x64` and `x86` pass the analysis (all being built from the same code and same process)?

stephengillie:
> If I understand correctly the validation fails on the analysis of the application which is linked to some antivirus, is that right? Where could I find a detail of the problem to solve it, since I can't find much information in the logs to explain why the `arm64` version fails when both the `x64` and `x86` pass the analysis (all being built from the same code and same process)?

This was due to a pipeline issue. Sorry for the false alarm.

stephengillie (approved):

Alphrag:
> This was due to a pipeline issue. Sorry for the false alarm. 

No worries, thanks for checking it.

PR #107340, opened 2023-05-19
(SpencerSoft.Chess) Point to latest version with bug fixes and new features.

spencerparkin:
This updated version fixes two major bugs...
1) A player could castle to get out of check, which is illegal, and...
2) The computer would fail to find check-mate in some cases.

A few other fixes and features have been added as well.

Thank you for considering this pull request.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/107340)


manifests/s/SpencerSoft/Chess/1.0.0/SpencerSoft.Chess.installer.yaml
```diff
@@ -7,8 +7,8 @@ Installers:
 - InstallerLocale: en-US
   Architecture: x86
   InstallerType: msi
-  InstallerUrl: https://github.com/spencerparkin/ChessEngine/releases/download/v1.2/ChessInstaller.msi
-  InstallerSha256: A9183F4E5F4A4AB104A1182B248768AD1E598C92D61AC084F635FB3800273E32
+  InstallerUrl: https://github.com/spencerparkin/ChessEngine/releases/download/v1.3/ChessInstaller.msi
```
stephengillie, line 10: Shouldn't the PackageVersion be updated from `1.0.0` to `1.3` to match the URL version? This would allow upgrading to work correctly - performed this way, I believe the WinGet client will see these as the same version, so users who already have it installed wouldn't see an upgrade from `WinGet upgrade`. 

(Please note that updating the PackageVersion has to be performed in all 3(?) manifest files, and their paths have to be updated to match as well.)

spencerparkin:
Hi.  I tried to fix my .yaml files.  There should now be a package for every release that there's been.  Regardless, I suspect that the automation will still fail on something.  It just doesn't like my .yaml files for some reason, or maybe the installer binary.  There always seems to be an internal error.  Thank you for your help.

spencerparkin:
Okay, I had to undo some of my changes, because the automation told me that I can update/add one and only one manifest per pull request.  What I tried to do is make a manifest file for each release I had, but I guess we're beyond that now.  My apologies.

stephengillie (approved):

PR #107661, opened 2023-05-21
Remove user scope from filips.FirefoxPWA 2.5.0

mdanish-kh:
`user` scope for WiX installers generally does not work well because of a bug/design issue with WiX installers in that the installer installs per-user but incorrectly writes the uninstall entries to HKEY_LOCAL_MACHINE that makes WinGet think it's a per-machine install. This causes users not being able to upgrade the package after installation through WinGet. For best UX, it's best to remove the user scope from the manifest

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/107661)

### DIFF
--- a/manifests/f/filips/FirefoxPWA/2.5.0/filips.FirefoxPWA.installer.yaml
+++ b/manifests/f/filips/FirefoxPWA/2.5.0/filips.FirefoxPWA.installer.yaml
@@ -5,49 +5,25 @@ PackageIdentifier: filips.FirefoxPWA
 PackageVersion: 2.5.0
 InstallerLocale: en-US
 MinimumOSVersion: 10.0.0.0
+ElevationRequirement: elevatesSelf
+Scope: machine
 UpgradeBehavior: uninstallPrevious
 Installers:
 - Architecture: x86
   InstallerType: wix
-  Scope: user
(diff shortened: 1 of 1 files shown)


russellbanks (approved):

filips123:
Is there any way to fix the installer, so it works properly both in user and machine scope?

mdanish-kh:
This is an issue with almost every MSI installer that is made using WiX. You can find related discussion over at https://github.com/microsoft/winget-cli/issues/3011

filips123:
So, this means that WiX-made installers cannot be installed in a user scope with `winget`, unless this issue is fixed in WiX Toolset or `winget-cli` gets some other way of detecting installation scope?

mdanish-kh:
> So, this means that WiX-made installers cannot be installed in a user scope with `winget`, unless this issue is fixed in WiX Toolset or `winget-cli` gets some other way of detecting installation scope?

Unfortunately yes. WinGet detects scope from where the installer writes its uninstall entries in Registry (HCKU corresponds to `user`, HKLM -> `machine`). The issue is that if a user installs a WiX installer in user scope, subsequent updates would fail because WinGet looks for a compatible installers in the upgrade scenario that matches the architecture and scope of the current installation. Since the current installation would be detected as "machine-scoped", WinGet would fail with `No applicable upgrade found`

PR #107722, opened 2023-05-22
Netfree.Anywhere_2 version 2.0.0

rocketinventor:

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/107722)

### DIFF
--- a/manifests/n/Netfree/Anywhere_2/2.0.0/Netfree.Anywhere_2.installer.yaml
+++ b/manifests/n/Netfree/Anywhere_2/2.0.0/Netfree.Anywhere_2.installer.yaml
@@ -1,0 +1,12 @@
+# Created using wingetcreate 1.2.5.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.4.0.schema.json
+
+PackageIdentifier: Netfree.Anywhere_2
+PackageVersion: 2.0.0
+Installers:
+- InstallerUrl: https://netfree.link/pc/windows/NetFreeAW2.exe
+  Architecture: x86
+  InstallerType: nullsoft
+  InstallerSha256: E381A9038BB23FE2BAFB2BC7AC1B9C1BA9B9FC8F5DDEC22EFE7BE6294BC80244
+ManifestType: installer
+ManifestVersion: 1.4.0
--- a/manifests/n/Netfree/Anywhere_2/2.0.0/Netfree.Anywhere_2.locale.he-IL.yaml
+++ b/manifests/n/Netfree/Anywhere_2/2.0.0/Netfree.Anywhere_2.locale.he-IL.yaml
@@ -1,0 +1,12 @@
+# Created using wingetcreate 1.2.5.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.4.0.schema.json
+
+PackageIdentifier: Netfree.Anywhere_2
+PackageVersion: 2.0.0
+PackageLocale: he-IL
+Publisher: Net Free
+PackageName: NetFree Anywhere
+License: All Rights Reserved
+ShortDescription: Install Netfree on your device, and be able to use it anywhere!
+ManifestType: defaultLocale
+ManifestVersion: 1.4.0
--- a/manifests/n/Netfree/Anywhere_2/2.0.0/Netfree.Anywhere_2.yaml
+++ b/manifests/n/Netfree/Anywhere_2/2.0.0/Netfree.Anywhere_2.yaml
@@ -1,0 +1,8 @@
+# Created using wingetcreate 1.2.5.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.version.1.4.0.schema.json
+
+PackageIdentifier: Netfree.Anywhere_2
+PackageVersion: 2.0.0
+DefaultLocale: he-IL
+ManifestType: version
+ManifestVersion: 1.4.0


stephengillie:
image

rocketinventor:
@stephengillie Thank you for your quick review of this PR. In regards to the screenshot that you shared, once the "Install" button is pressed, the full process goes back to being completely automatic - nothing else needs to be done at all since the installer is also bundled.

Is there any way just to bypass this popup, or even to simulate pressing the "install" button?

Trenly:
> Is there any way just to bypass this popup, or even to simulate pressing the "install" button? 

If the publisher of Anywhere has a command line switch to not install the drivers then it would, but there isn't a way to simulate pressing the button. Currently all packages in winget must be able to install without user interaction until support for filtering by install type is implemented. I don't believe this is blocked by dependencies since drivers aren't able to be managed by winget at the moment, instead I would call this an "Interactive Only Installer"

stephengillie:
Unfortunately, this issue is common with packet capture and network tapping programs, as many rely on a special network driver to enable their functionality.

rocketinventor:
Hello all! After reviewing this issue, I don't think that it is accurate to call it a "user interaction" issue, _or_ a dependency issue. It is actually just an issue of certificates, and getting them to be trusted by the system: If you look closely at the pop-up, you can see that it says "Windows Security" on it. The popup is not actually from the installer - it is from Windows. This means that there should be a way to fix it.

It seems that the solution is just to add the cert to the trusted store before the installation step is run. I will include some more information because I think that it will be useful for others.

I found that it can be done from the command line: [`certutil -addstore "TrustedPublisher" cert_name.cer`](https://www.itninja.com/question/disabling-would-you-like-to-install-this-device-software#:~:text=certutil%20%2Daddstore%20%22TrustedPublisher%22%20cert_name.cer). I tested out this method and it seems to work. Attached below is the cert inside of **Windows Cert Manager** `certmgr.msc` after the software was installed. It is possible to get this same result from the command line. I tried it myself, and it seems to work.

2023-05-24 12_36_04-certlm -  Certificates - Local Computer_Trusted Publishers_Certificates


Running:
_certutil -addstore "TrustedPublisher" open_vpn_cert.cer_
Produces this output:
```
TrustedPublisher "Trusted Publishers"
Certificate "OpenVPN Technologies, Inc." added to store.
CertUtil: -addstore command completed successfully.
```

And does indeed add it to the certificate store.

---

I am still wholly unfamiliar with WinGet's distribution, etc, etc. My question for the community moderators (ie @Trenly), is on the implementation. Two questions: How do I bundle this certificate, and how do I make it run an operation with that file on the command line _before_ anything else is done? (Also, how do I mark that it has to be installed with "system" scope (ie admin)?

Thanks!

stephengillie:
> Hello all! After reviewing this issue, I don't think that it is accurate to call it a "user interaction" issue, _or_ a dependency issue. It is actually just an issue of certificates, and getting them to be trusted by the system: If you look closely at the pop-up, you can see that it says "Windows Security" on it. The popup is not actually from the installer - it is from Windows. This means that there should be a way to fix it.
> 
> It seems that the solution is just to add the cert to the trusted store before the installation step is run. I will include some more information because I think that it will be useful for others.
> 
> I found that it can be done from the command line: [`certutil -addstore "TrustedPublisher" cert_name.cer`](https://www.itninja.com/question/disabling-would-you-like-to-install-this-device-software#:~:text=certutil%20%2Daddstore%20%22TrustedPublisher%22%20cert_name.cer). I tested out this method and it seems to work. Attached below is the cert inside of **Windows Cert Manager** `certmgr.msc` after the software was installed. It is possible to get this same result from the command line. I tried it myself, and it seems to work.
> 
> 2023-05-24 12_36_04-certlm - Certificates - Local Computer_Trusted Publishers_Certificates
> Running: _certutil -addstore "TrustedPublisher" open_vpn_cert.cer_ Produces this output:
> 
> ```
> TrustedPublisher "Trusted Publishers"
> Certificate "OpenVPN Technologies, Inc." added to store.
> CertUtil: -addstore command completed successfully.
> ```
> 
> And does indeed add it to the certificate store.
> 
> I am still wholly unfamiliar with WinGet's distribution, etc, etc. My question for the community moderators (ie @Trenly), is on the implementation. Two questions: How do I bundle this certificate, and how do I make it run an operation with that file on the command line _before_ anything else is done? (Also, how do I mark that it has to be installed with "system" scope (ie admin)?
> 
> Thanks!

Hi @rocketinventor,

Wow, that's some good research. Thank you for putting together this solution. 

This certificate would bypass the security check on all software from "OpenVPN Technologies, Inc." in the future, even after this package were removed. Using this to allow the driver to install might create unexpected security issues that some users at publicly owned corporations might not be able to justify to their auditors, and also might be used in nefarious ways against individual users. Such as someone spoofing the publisher name through a vulnerability that might not yet exist. 

A preferrable path would be to specify the driver as a dependency, so its permissions can be handled independently and automatically. 
- This package, `TAP-Windows Provider V9 Network adapters` (thus) doesn't appear to be available through the WInGet package manager yet.
- Is this package available from OpenVPN for separate download? 
- If another adapter is preinstalled, will this package use that instead? We might already have another suitable adapter in the public manifests. 

Regarding the actual goal, the WinGet package manager doesn't currently have the functionality to add or remove certificates directly, nor to perform scripted operations or run PowerShell scripts (including PowerShell installers). An installer would need to be built around performing the certificate install - I'm not aware of installers for installing certificates, and only speculating that some might offer the functionality. That installer could then be submitted into the WinGet public manifests and be available for public use once approved. This might work better as a [feature request](https://github.com/microsoft/winget-pkgs/issues).

Let me know what you think. I'd be happy to help to move this forward in any direction.

PR #108340, opened 2023-05-29
It look like it confused with winget on x86 machine. Removing x64 fro…

BrandonWanHuanSheng:
…m 64Gram.64Gram..

Reason: It confused with winget on x86. 

Screenshot (7)

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-pkgs/pull/108340)


mdanish-kh (changes requested):
I think that's expected behavior as the publisher stopped making x86 releases. If you explicitly run `winget upgrade 64gram.64gram`, it will tell you that a new upgrade is available but your system requirements do not match the new installer. That gives user a hint as to what's blocking their update. I do agree `winget uprade --all` should give a little more output as to why an upgrade fails but that is not the scope of this PR. Removing this package like this will block x64 users from receiving the update when they should be receiving it.

BrandonWanHuanSheng:
But actually I am turning into New Package Identifier.. Otherwise, it will have another issue.. It look like x86 version instead of x64 version.

Issue #94743, opened 2023-01-26
[Package Issue]: OpenRCT2.OpenRCT2 wrong version advertisement leads to update loop


markusdd:
### Please confirm these before moving forward

- [X] I have searched for my issue and not found a work-in-progress/duplicate/resolved issue.
- [X] I have not been informed if the issue is resolved in a preview version of the winget client.

### Category of the issue

Installation issue.

### Brief description of your issue

update loop caused by wrong version (0.4.3 is advertised here, but in reality 0.4.3 is followed by a hash and branch name)

### Steps to reproduce

install OpenRCT2, then run an update via winget

### Actual behavior

update is offered again and again

### Expected behavior

update happens and is not offered again

### Environment

```shell
> winget --info
Windows-Paket-Manager v1.4.10173
Copyright (c) Microsoft Corporation. Alle Rechte vorbehalten.

Windows: Windows.Desktop v10.0.22621.1105
Systemarchitektur: X64
Paket: Microsoft.DesktopAppInstaller v1.19.10173.0
```

### Screenshots and Logs

Here is the relted ticket at OpenRCT2, which rightfully redirected here:
https://github.com/OpenRCT2/OpenRCT2/issues/19072

denelon:
What does the output from `winget list` show for the installed version and the available version?

mdanish-kh:
```pwsh
~ winget list openrct2
Name                         Id                Version             Available Source
-----------------------------------------------------------------------------------
OpenRCT2 0.4.3master-285e0fc OpenRCT2.OpenRCT2 0.4.3master-285e0fc 0.4.3     winget
```
DisplayVersion is different from marketing version. I noticed something while going through the older manifests here [manifests/o/OpenRCT2](https://github.com/microsoft/winget-pkgs/tree/master/manifests/o/OpenRCT2/OpenRCT2). Some releases use the same DisplayVersion as Marketing versions while others use `MarketingVersion<branch_name><commit_hash>` as DisplayVersion. This behavior can be fixed by either changing the PackageVersion of the recent release to `0.4.3master-285e0fc` or use AppsAndFeaturesEntries.

I'm more tempted towards the latter to have all versions use the Marketing version but that would also require going through the previous releases and adding AppsAndFeaturesEntries to every manifest to avoid an upgrade loop again 😅 I'll open PRs to fix this behavior

markusdd:
exactlym thanks for providing this here.

[...]

PR #109585, opened 2023-06-12
​

ItzLevvie:
​

### DIFF
--- a/manifests/v/VideoLAN/VLC/3.0.18/VideoLAN.VLC.installer.yaml
+++ b/manifests/v/VideoLAN/VLC/3.0.18/VideoLAN.VLC.installer.yaml
@@ -53,21 +53,39 @@ ReleaseDate: 2022-11-08
 Installers:
 - Architecture: x64
   InstallerType: wix
-  InstallerUrl: https://download.videolan.org/pub/videolan/vlc/3.0.18/win64/vlc-3.0.18-win64.msi
+  InstallerUrl: https://download.videolan.org/videolan/vlc/3.0.18/win64/vlc-3.0.18-win64.msi
   InstallerSha256: A618D66B2F753343402200AD17D92EB6D4AB10DB0DD1A098402E2968AA3114C4
   ProductCode: '{B022B1C5-D067-42CB-98E2-D965E4D74CFE}'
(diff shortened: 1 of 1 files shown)


mdanish-kh:
Yep, I'm confident this will resolve it because of https://github.com/microsoft/winget-pkgs/issues/88726#issuecomment-1398993435

mdanish-kh (approved):

mdanish-kh (approved):

mdanish-kh:
@ItzLevvie On second thoughts, this may not be the fix; I re-call if one manifest contains `DisplayVersion`, then all manifests (including past ones) should contain this key in order to resolve the issue.

The solution may just well be to get rid of `3.0.17.4` since no other manifest requires DisplayVersion.
Ref: https://github.com/microsoft/winget-pkgs/issues/88726#issuecomment-1398993435
https://github.com/microsoft/winget-pkgs/issues/94743#issuecomment-1406868090

mdanish-kh:
@kriswilk I actually wrote that before, but the problem is that the current 3.0.18 does require `DisplayVersion` because the version is different between the MSI (`3.0.18.0`) and the EXE (`3.0.18`) installers. We use DisplayVersion key where we want to differentiate between a marketing version and a version that is used in Apps&Features / Registry. Since we cant have the same PackageVersion for both MSIs and EXEs as they write different versions to registry, we need to use `DisplayVersion` here for the MSIs

kriswilk:
> This discrepancy was introduced in 3.0.17.4 and we may just need to remove all manifests prior to that version for the latest ones to work correctly.

OK, now I understand the issue. Thanks for the clarification!

kriswilk:
BTW, I notice that this PR removes the `/pub/` from all the URLs. Note that previous manifests had even simpler URLs without the `/videolan/`. However, this URL path is buggy. For instance, the 32-bit EXE URL is broken:

`https://download.videolan.org/vlc/3.0.18/win32/vlc-3.0.18-win32.exe`

I specifically used the URLs with `/pub/videolan/vlc/` because this is the URL style you get when you navigate to the release archives from the application website. I presume these are the most likely to be maintained/tested.

That said, the ones without `pub` do seem to be OK for this release. 🤷‍♂️

PR #110649, opened 2023-06-25
Add label trigger for Validation-Missing-Dependency

mdanish-kh:
cc @Trenly @stephengillie

Related PR https://github.com/microsoft/winget-pkgs/pull/110640

-----

 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-pkgs/pull/110649&drop=dogfoodAlpha

### DIFF
--- a/.github/policies/labelAdded.validationMissingDependency.yml
+++ b/.github/policies/labelAdded.validationMissingDependency.yml
@@ -25,7 +25,7 @@ configuration:
                 Hello @${issueAuthor},
 
 
-                It seems that a specified package dependency does not exist in the repository. Please double-check the spellings and ensure that the dependency is not available under a different PackageIdentifier. If the dependency is indeed missing, you may consider contributing it to the repository.
+                The validation pipeline could not locate a specified dependency. Please review the spelling and check whether the dependency is associated with a different identifier.
 
 
                 Template: msftbot/validationError/missingDependency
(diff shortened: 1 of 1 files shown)


.github/policies/labelAdded.validationMissingDependency.yml
```diff
@@ -0,0 +1,43 @@
...
+          - payloadType: Pull_Request
+          - labelAdded:
+              label: Validation-Missing-Dependency
+        then:
+          - addReply:
+              reply: >-
+                Hello @${issueAuthor},
+
+
+                It seems that a specified package dependency does not exist in the repository. Please double-check the spellings and ensure that the dependency is not available under a different PackageIdentifier. If the dependency is indeed missing, you may consider contributing it to the repository.
```
Trenly, line 28: I know this is specific to PackageDependencies right now, but WindowsFeatures dependencies are in experimental. Maybe the message needs to consider both of them? Of course we could always update it later too, so I'm fine with this as-is, just something to think about.

Trenly (approved):

Trenly (approved):

.github/policies/labelAdded.validationMissingDependency.yml
mdanish-kh: Thanks for pointing this out, I forgot there were multiple types of dependencies. Would love to see specific labels and messages for each type of dependency but until then we can have a generic message.

denelon (approved):

mdanish-kh:
cc @denelon Can you tag `Validation-Missing-Dependency` as a test?

PR #111183, opened 2023-07-03
Add labels automatically when a hash mismatch is mentioned

Trenly:
cc @stephengillie / @denelon
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-pkgs/pull/111183&drop=dogfoodAlpha

### DIFF
--- a/.github/policies/labelManagement.issueOpened.yml
+++ b/.github/policies/labelManagement.issueOpened.yml
@@ -96,6 +96,7 @@ configuration:
       - description: >-
           When a pull request is opened, if the content contains "Hash Mismatch" or "Hash does not match" (case insensitive)
           * Add the Update-Request label
+          * Add the Error-Hash-Mismatch label
         if:
           - payloadType: Issues
           - isAction:
@@ -106,5 +107,7 @@ configuration:
         then:
           - addLabel:
               label: Update-Request
+          - addLabel:
+              label: Error-Hash-Mismatch # It's okay to add this label to issues since the labelAdded.errorHashMismatch triggers only respond to PRs
 onFailure:
 onSuccess:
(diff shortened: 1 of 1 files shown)


.github/policies/labelManagement.issueOpened.yml
```diff
@@ -93,5 +93,18 @@ configuration:
...
+        if:
+          - payloadType: Issues
+          - isAction:
+              action: Opened
+          - bodyContains:
+              pattern: (?i)hash[\s-](mismatch|does\snot\smatch)
+              isRegex: True
+        then:
+          - addLabel:
+              label: Update-Request
```
mdanish-kh, line 108: How about adding Error-Hash-Mismatch label too? Also this got me thinking, since hash mismatches are so common, maybe it's better we have another category for it in the issue templates that automatically applies that label. That'll help filtering out the "trivial" hash mismatch cases and the other non-obvious package issues..

Trenly: I think that adding the label automatically would work, since the Error-Hash-Mismatch trigger looks like it's only set to respond to PR's.

I don't think that having a separate issue template for it would really help at all, since many inexperienced users may simply classify it as a bug or a package issue. Also, the specific issue type for hash mismatches in the bug template would hit this trigger and apply the labels anyways

.github/policies/labelManagement.issueOpened.yml
```diff
@@ -96,6 +96,7 @@ configuration:
       - description: >-
           When a pull request is opened, if the content contains "Hash Mismatch" or "Hash does not match" (case insensitive)
```
mdanish-kh, line 97:
```suggestion
          When an issue is opened, if the content contains "Hash Mismatch" or "Hash does not match" (case insensitive)
```

mdanish-kh (approved):

denelon (approved):

Issue #111840, opened 2023-07-11
[Package Request]: Microsoft.UI.Xaml


denelon:
### How can we help?

I would like someone else to build the manifest.

### Please read and ensure the following

- [X] The installer meets the above requirements

### Please provide the following information

Download Page Url: https://github.com/microsoft/microsoft-ui-xaml
Publisher: Microsoft
Package Name: Windows UI Library
Description: WinUI is a user interface layer that contains modern controls and styles for building Windows apps.
Package Versions: 2.7.3 & 2.8.5 
Installer URLs:

https://github.com/microsoft/microsoft-ui-xaml/releases/tag/v2.7.3
https://github.com/microsoft/microsoft-ui-xaml/releases/tag/v2.8.5

I'd propose the ID: Microsoft.UI.Xaml
Moniker: winui

Related to:
* https://github.com/microsoft/winget-cli/issues/2134

denelon:
The upgrade behavior should be "install" so side-by-side installation can work correctly, or this should be treated as separate packages like other frameworks/libraries.

Trenly:
@denelon - Should be resolved now

```
winget install Microsoft.UI.Xaml.2.7
winget install winui-2.7

winget install Microsoft.UI.Xaml.2.8
winget install winui-2.8
```

cbr-github:
@denelon and @Trenly :

When installing

https://github.com/microsoft/winget-pkgs/tree/master/manifests/m/Microsoft/UI/Xaml/2/8

I'm getting the following error:
```
Installation wird ausgeführt...
-
███▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒  1024 KB / 8.44 MB
███████▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒  2.00 MB / 8.44 MB
██████████▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒  3.00 MB / 8.44 MB
██████████████▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒  4.00 MB / 8.44 MB
█████████████████▒▒▒▒▒▒▒▒▒▒▒▒▒  5.00 MB / 8.44 MB
█████████████████████▒▒▒▒▒▒▒▒▒  6.00 MB / 8.44 MB
[...]

Issue #99631, opened 2023-03-16
[Package Issue]: Microsoft.PowerShell.Preview 7.4.1.0 - wrong version


Andrej730:
### Please confirm these before moving forward

- [X] I have searched for my issue and not found a work-in-progress/duplicate/resolved issue.
- [X] I have not been informed if the issue is resolved in a preview version of the winget client.

### Brief description of your issue

Powershell Preview 7.4.0.1 added to winget packages as 7.4.1.0 which is a bit confusing - shouldn't it be 7.4.0.1?
https://github.com/microsoft/winget-pkgs/blob/master/manifests/m/Microsoft/PowerShell/Preview/7.4.1.0/Microsoft.PowerShell.Preview.installer.yaml

Similar thing with upcoming PR #99498

It could be an issue if 7.4.1 will actually come out

denelon:
@anamnavi is this intentional?

mdanish-kh:
The reason is that MSIX installers report a different version to registry than the MSI ones. The two versions are separated as:

- PackageVersion is set to `7.4.1.0` as that is what MSIX installers write to registry
- DisplayVersion field `7.4.0.1` is used for the MSI ones

Since the MSIX installers are a temporary solution for the ARM systems (the pwsh team is working on MSI installers for arm architecture), I would say we use PackageVersion field according to the MSI installers and the marketing version i.e., `7.4.0.1` and use DisplayVersion field for the MSIX ones.

mdanish-kh:
> Similar thing with upcoming PR https://github.com/microsoft/winget-pkgs/pull/99498

This was resolved in commit https://github.com/microsoft/winget-pkgs/pull/99498/commits/cd3ce1f949f36fe16f7b2f0c9eb7635e2e4e8367 by @anamnavi in the latest PR.

I've created PRs to fix the version in previous manifets. 
cc @anamnavi to please review these PRs:
- #99811
- #99812
- #99813
- #99814

anamnavi:
@mdanish-kh yes that captures our scenario accurately since we are temporarily specifying MSIX installers until we have msi installer for arm.

So we would use `7.4.0.1` as the PackageVersion field. Where/how would we specify `7.4.1.0` for the msix installer?

mdanish-kh:
[...]

PR #111849, opened 2023-07-11
New package: Microsoft.UI.Xaml.2.8 version 2.8.5

Trenly:

* Related to #111840

-----

 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-pkgs/pull/111849&drop=dogfoodAlpha

### DIFF
--- a/manifests/m/Microsoft/UI/Xaml/2/8/2.8.5/Microsoft.UI.Xaml.2.8.installer.yaml
+++ b/manifests/m/Microsoft/UI/Xaml/2/8/2.8.5/Microsoft.UI.Xaml.2.8.installer.yaml
@@ -5,6 +5,7 @@ PackageIdentifier: Microsoft.UI.Xaml.2.8
 PackageVersion: 2.8.5
 InstallerType: appx
 UpgradeBehavior: install
+MinimumOSVersion: 10.0.15063.0
 AppsAndFeaturesEntries:
   - DisplayVersion: 8.2306.22001.0
     DisplayName: Microsoft.UI.Xaml.2.8
--- a/manifests/m/Microsoft/UI/Xaml/2/8/2.8.5/Microsoft.UI.Xaml.2.8.installer.yaml
+++ b/manifests/m/Microsoft/UI/Xaml/2/8/2.8.5/Microsoft.UI.Xaml.2.8.installer.yaml
@@ -5,6 +5,11 @@ PackageIdentifier: Microsoft.UI.Xaml.2.8
 PackageVersion: 2.8.5
 InstallerType: appx
 UpgradeBehavior: install
+AppsAndFeaturesEntries:
+  - DisplayVersion: 8.2306.22001.0
+    DisplayName: Microsoft.UI.Xaml.2.8
+    Publisher: CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US
+    InstallerType: appx
 Installers:
 - Architecture: x64
   InstallerUrl: https://github.com/microsoft/microsoft-ui-xaml/releases/download/v2.8.5/Microsoft.UI.Xaml.2.8.x64.appx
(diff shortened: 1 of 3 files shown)


Trenly:
Looks like the validation pipeline is too strict on Appx  packages, saying that Winget Manifest PackageVersion must match the version in the Appx Manifest, even though the true version is specified in AppsAndFeaturesEntries

AmelBawa-msft:
@Trenly @denelon  Correct, today the validation does the following comparison:
Is the YAML `PackageVersion: 2.8.5` the same as the AppxManifest.xml `<Identity Name="Microsoft.UI.Xaml.2.8" ... Version="8.2306.22001.0" />`, which in this case it does not match.

@denelon, what should be the new behavior for this validation? Prioritizing `AppsAndFeaturesEntries.DisplayVersion`, checking if either of `PackageVersion` or `AppsAndFeaturesEntries.DisplayVersion` matches, or something else?

denelon:
The more critical thing to match is the AppsAndFeaturesEntries.DisplayVersion if it's present, then we could allow the match to comply with our policy rather than requiring that the PackageVersion matching.

If the DisplayVersion is not present, then the PackageVersion must match.
If the DisplayVersion is present, then it must match, and the PackageVersion is not required to match.

mdanish-kh (approved):

yao-msft:
This manifest will have winget upgrade related issues. Currently, AppsAndFeaturesEntries are for representing the ARP registry entries written by win32 apps. For msix, these entries are ignored and not used. We even have a validation that on individual installer level, if AppsAndFeaturesEntries exist with msix, a validation error will be returned.

If we do wish to let msix look at AppsAndFeaturesEntries, we need to go through the code and make sure no extra surprise would happen if we enable this new behavior. And for now I cannot think of a reason to enable this. Because msix always have a well defined version (no version mapping is actually needed). And we should implement the MarketingVersion (@trenly's proposal) if we want to allow showing a different version other than the appxmanifest version.

yao-msft:
I'll create 2 follow up prs to make the manifest correct (aligned with current implementation)

Trenly:
@yao-msft - You may want to also check the Microsoft Powershell, since I believe I recall someone mentioning it had the same issue

yao-msft:
> @yao-msft - You may want to also check the Microsoft Powershell, since I believe I recall someone mentioning it had the same issue

I looked at the latest powershell manifest and preview manifest, they do not declare AppsAndFeaturesEntries. Are you referring to their package version not matching appxmanifest version?

Trenly:
> Are you referring to their package version not matching appxmanifest version?

Yes

mdanish-kh:
The concerning package is actually `Microsoft.PowerShell.Preview` and its case is a little bit different than this one. 

The issue is that the Preview manifest has InstallerTypes msi and msix where they don't have the same DisplayVersion value. Take the current latest manifest [7.4.0.4](https://github.com/microsoft/winget-pkgs/blob/master/manifests/m/Microsoft/PowerShell/Preview/7.4.0.4/Microsoft.PowerShell.Preview.installer.yaml) as an example. The MSI DisplayVersion is `7.4.0.4` while the version the MSIX uses is `7.4.4.0`. Previously, what we were doing was to use the MSIX version as the PackageVersion in the manifest and use `AppsAndFeaturesEntries` for the MSI installers. The problem with this approach is that the marketing version is actually the MSI version and we had users complaining that they were not being shown the correct version in winget upgrade / list output where the PackageVersion (msix version) was being shown. (https://github.com/microsoft/winget-pkgs/issues/99631). I looked at the manifest and then removed the AppsAndFeaturesEntries -> used MSI version as the PackageVersion / user-facing version -> tried to use ARP entries for msix where I was quickly met with the issue with `winget validate` not passing. So then I removed the ARP Entries as it didn't seem to mess with correlation after installing the package but then again the issue arose with the pipelines not allowing the PR to pass through the validation pipelines since they require the PackageVersion to match the version retrieved from the MSIX. Seeing all this, and having discussions with the PowerShell team in https://github.com/microsoft/winget-pkgs/issues/99631 and in https://github.com/microsoft/winget-pkgs/pull/99498, it seemed that we should have manual merges for these PRs as it's a temp scenario. The temporary scenario is due to the fact the pwsh team is working for making MSIs for ARM arch but until they start releasing them as MSI, they are using MSIX as a temporary solution for ARM (ref https://github.com/microsoft/winget-pkgs/issues/99631#issuecomment-1474716847)

mdanish-kh:
@yao-msft What would you advise to do in this situation? If we were to use the MSIX version as the PackageVersion so that we don't need manual merges (and use AppsAndFeaturesEntries for MSI), we would have to go back and edit all previous manifests of PowerShell Preview as well to avoid the issue with version ranges and infinite upgrades for the user. And then again, the main problem would be that the user-facing version would not match the marketing version and imo it's more important that users feels they are getting the correct version from WinGet.

PR #112033, opened 2023-07-13
fix hash

CoolPlayLin:
Resolve #111954

-----

 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-pkgs/pull/112033&drop=dogfoodAlpha


mdanish-kh (changes requested):
Typically a hash mismatch indicates a newer release and is the case here; You'll have to set `PackageVersion: 10.10.58` in all yaml files and add the manifest in a newer directory `10.10.58`.

One way to do this would be here in GitHub Web by pressing the <kbd>.</kbd> (dot) key on your keyboard that'll open up the PR in VS Code Online. Navigate to Explorer menu, find the required folder and make the appropriate cut-paste change.

After that, stage and commit your changes from Source Control menu on the left. (The changes are automatically pushed)

stephengillie (approved):

PR #115303, opened 2023-08-04
Add Developer ISSUE_TEMPLATE

CoolPlayLin:
CC @denelon @stephengillie

-----

 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-pkgs/pull/115303&drop=dogfoodAlpha

### DIFF
--- a/.github/ISSUE_TEMPLATE/package_request.yml
+++ b/.github/ISSUE_TEMPLATE/package_request.yml
@@ -1,5 +1,5 @@
 name: "➕ New Package Request for User"
-description: Suggest a package for submission (this does not mean you have to implement it) This template is provide with user who love software.
+description: Suggest a package for submission (this does not mean you have to implement it), This template is provide for users who love software.
 labels: ["Help-Wanted", "Package-Request"]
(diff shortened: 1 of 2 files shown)


Trenly:
As discussed on Gitter / Element - I feel that having a template with such a similar name will only confuse users more, especially because it won't be clear who should use which template or when.

CoolPlayLin:
@Trenly
I've modified the description of these template to differentiate them.

Trenly (changes requested):
> @Trenly
> I've modified the description of these template to differentiate them.

I still don’t think the difference is clear. Also, if someone is taking the time to provide all the manifest information in an issue, it would be just as fast (and arguably easier) for them to make the manifest themselves.

I disagree with the concept at a fundamental level

stephengillie:
If this template gathers all the info needed to build a manifest, then it should also kick off a GitHub Action to generate the template PR. Basically this would be a web-based intake method. But should be optional.

Trenly:
Since this has been sitting for over a month, marking it as stale;

PR #115308, opened 2023-08-04
Remove: calibre.calibre version 6.23.0

pxwiz:
###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-pkgs/pull/115308&drop=dogfoodAlpha


Trenly:
Is there a reason for removing this package?

pxwiz:
> Is there a reason for removing this package?

Submitted manifests for new version of this package & the portable version.

Trenly:
> > Is there a reason for removing this package?
> 
> Submitted manifests for new version of this package & the portable version.

But the URLs for this version still work? Newer versions don't always mean the older versions need to be removed, as long as the download and install still works, I think it's best to keep the versions available for users to install with `--version`

pxwiz:
> > > Is there a reason for removing this package?
> > 
> > 
> > Submitted manifests for new version of this package & the portable version.
> 
> But the URLs for this version still work? Newer versions don't always mean the older versions need to be removed, as long as the download and install still works, I think it's best to keep the versions available for users to install with `--version`

As there is no cleanup for historical versions, I typically submit manifests for new version & delete version together, so my previous submission is removed.  I don't want to manage historical versions, if the developer decides to remove the binaries from his downloads sections.

stephengillie (approved):

PR #117762, opened 2023-08-23
New version: Google.Chrome version 116.0.5845.111

PckgrBot:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/117762)


Trenly (changes requested):
Because google chrome uses a vanity URL, the previous version should also be removed when the hash changes

Trenly (approved):

Issue #63220, opened 2022-06-09
[Package Issue]: Freeplane.Freeplane


oldium:
### Please confirm these before moving forward

- [X] I have searched for my issue and not found a work-in-progress/duplicate/resolved issue.
- [X] I have not been informed if the issue is resolved in a preview version of the winget client.

### Category of the issue

Installation issue.

### Brief description of your issue

The download URL of version 1.9.14 uses wrong URL, so the installation currently fails on 404 (File Not Found).

Wrong URL: https://sourceforge.net/projects/freeplane/files/freeplane%20stable/Freeplane-Setup-1.9.14.exe/download

Good URL: https://sourceforge.net/projects/freeplane/files/freeplane%20stable/archive/1.9.14/Freeplane-Setup-1.9.14.exe/download

Also a new version 1.10.2 is available, which uses the following URL: https://sourceforge.net/projects/freeplane/files/freeplane%20stable/archive/1.10.2/Freeplane-Setup-1.10.2.exe/download

### Steps to reproduce

`winget install --id=Freeplane.Freeplane  -e`

### Actual behavior

```
#> winget install --id=Freeplane.Freeplane  -e
Found Freeplane [Freeplane.Freeplane] Version 1.9.14
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Downloading https://sourceforge.net/projects/freeplane/files/freeplane%20stable/Freeplane-Setup-1.9.14.exe/download
An unexpected error occurred while executing the command:
Download request status is not success.
0x80190194 : Nenalezeno (404)```

### Expected behavior

Download and installation succeeds.

### Environment

```shell
Windows Package Manager v1.2.10271
Windows: Windows.Desktop v10.0.19044.1741
Package: Microsoft.DesktopAppInstaller v1.17.10271.0
```

oldium:
Wow. That was fast. Thanks. 👍

PR #118144, opened 2023-08-26
New version: Freeplane.Freeplane version 1.11.6

jeremytmsi:

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/118144)

### DIFF
--- a/manifests/f/Freeplane/Freeplane/1.11.6/Freeplane.Freeplane.installer.yaml
+++ b/manifests/f/Freeplane/Freeplane/1.11.6/Freeplane.Freeplane.installer.yaml
@@ -1,0 +1,16 @@
+# Created with Komac v1.10.1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.5.0.schema.json
+
+PackageIdentifier: Freeplane.Freeplane
+PackageVersion: 1.11.6
+Platform:
+- Windows.Desktop
+InstallerType: inno
+Scope: machine
+UpgradeBehavior: install
+Installers:
+- Architecture: neutral
+  InstallerUrl: https://sourceforge.net/projects/freeplane/files/freeplane%20stable/Freeplane-Setup-1.11.6.exe/download
+  InstallerSha256: F02FF69C0AA17925E3989E77EF614AE707DB4D7AD3D26DBA3024F22BC6868E89
+ManifestType: installer
+ManifestVersion: 1.5.0
--- a/manifests/f/Freeplane/Freeplane/1.11.6/Freeplane.Freeplane.installer.yaml
+++ b/manifests/f/Freeplane/Freeplane/1.11.6/Freeplane.Freeplane.installer.yaml
@@ -10,7 +10,7 @@ Scope: machine
 UpgradeBehavior: install
 Installers:
 - Architecture: neutral
-  InstallerUrl: https://sourceforge.net/projects/freeplane/files/freeplane%20stable/Freeplane-Setup-1.11.6.exe/download
-  InstallerSha256: F02FF69C0AA17925E3989E77EF614AE707DB4D7AD3D26DBA3024F22BC6868E89
+  InstallerUrl: https://sourceforge.net/projects/freeplane/files/freeplane%20stable/Freeplane-Setup-with-Java-1.11.6.exe/download
+  InstallerSha256: 9CD74C6A516BB7229EA83F14EA7A2C25D78EA4844B777E913F2732882F6FE20F
 ManifestType: installer
 ManifestVersion: 1.5.0
--- a/manifests/f/Freeplane/Freeplane/1.11.6/Freeplane.Freeplane.locale.en-US.yaml
+++ b/manifests/f/Freeplane/Freeplane/1.11.6/Freeplane.Freeplane.locale.en-US.yaml
@@ -1,0 +1,26 @@
+# Created with Komac v1.10.1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.5.0.schema.json
+
+PackageIdentifier: Freeplane.Freeplane
+PackageVersion: 1.11.6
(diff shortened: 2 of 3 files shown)


oldium:
Just faced the issue with invalid download on 1.11.5 - the download link gets invalid with every release, so it would make sense to use `https://sourceforge.net/projects/freeplane/files/freeplane%20stable/archive/<version>/Freeplane-Setup-with-Java-<version>.exe/download` instead of plain `https://sourceforge.net/projects/freeplane/files/freeplane%20stable/Freeplane-Setup-with-Java-<version>.exe/download`.

oldium:
I fixed download URIs of 1.11.2 and 1.11.5 in #118158 and #118159

stephengillie:
> Just faced the issue with invalid download on 1.11.5 - the download link gets invalid with every release, so it would make sense to use archive link `https://sourceforge.net/projects/freeplane/files/freeplane%20stable/archive/<version>/Freeplane-Setup-with-Java-<version>.exe/download` instead of link `https://sourceforge.net/projects/freeplane/files/freeplane%20stable/Freeplane-Setup-with-Java-<version>.exe/download`, which is valid only for the latest release.
> 
> See also #63220

Hi @oldium,

The issue is really in that the developer removes old versions. The suggestion is basically to use a "vanity" URL that is updated with every release instead of a static URL, which runs counter to one of our best practices. 

We have to create a new manifest version (usually by copying forward details from the previous version), to add the new file hash. That the developer removes old versions means we have a 2nd step of removing the old manifest also. 

Right now, users will see a download error. If we update from the static URL, users will see the download begin and complete, then the install fail due to a hash mismatch. I think the former user experience is better, what are your thoughts?

stephengillie (approved):

oldium:
> > Just faced the issue with invalid download on 1.11.5 - the download link gets invalid with every release, so it would make sense to use archive link `https://sourceforge.net/projects/freeplane/files/freeplane%20stable/archive/<version>/Freeplane-Setup-with-Java-<version>.exe/download` instead of link `https://sourceforge.net/projects/freeplane/files/freeplane%20stable/Freeplane-Setup-with-Java-<version>.exe/download`, which is valid only for the latest release.
> > See also #63220
> 
> Hi @oldium,
> 
> The issue is really in that the developer removes old versions. The suggestion is basically to use a "vanity" URL that is updated with every release instead of a static URL, which runs counter to one of our best practices.
> 
> We have to create a new manifest version (usually by copying forward details from the previous version), to add the new file hash. That the developer removes old versions means we have a 2nd step of removing the old manifest also.
> 
> Right now, users will see a download error. If we update from the static URL, users will see the download begin and complete, then the install fail due to a hash mismatch. I think the former user experience is better, what are your thoughts?

@stephengillie That is not completely correct. The developer maintains the latest version at `https://sourceforge.net/projects/freeplane/files/freeplane%20stable`, where it is updated every release, so it is expected that the old release is replaced by a new release there and any link to that place gets broken. But at the same time the _identical_ latest release files and also all previous releases can be found at fixed location at `https://sourceforge.net/projects/freeplane/files/freeplane%20stable/archive`. It should be sufficient to take the release simply always from the archive and I do not understand why it is not like that.

stephengillie:
> > > Just faced the issue with invalid download on 1.11.5 - the download link gets invalid with every release, so it would make sense to use archive link `https://sourceforge.net/projects/freeplane/files/freeplane%20stable/archive/<version>/Freeplane-Setup-with-Java-<version>.exe/download` instead of link `https://sourceforge.net/projects/freeplane/files/freeplane%20stable/Freeplane-Setup-with-Java-<version>.exe/download`, which is valid only for the latest release.
> > > See also #63220
> > 
> > 
> > Hi @oldium,
> > The issue is really in that the developer removes old versions. The suggestion is basically to use a "vanity" URL that is updated with every release instead of a static URL, which runs counter to one of our best practices.
> > We have to create a new manifest version (usually by copying forward details from the previous version), to add the new file hash. That the developer removes old versions means we have a 2nd step of removing the old manifest also.
> > Right now, users will see a download error. If we update from the static URL, users will see the download begin and complete, then the install fail due to a hash mismatch. I think the former user experience is better, what are your thoughts?
> 
> @stephengillie That is not completely correct. The developer maintains the latest version at `https://sourceforge.net/projects/freeplane/files/freeplane%20stable`, where it is updated every release, so it is expected that the old release is replaced by a new release there and any link to that place gets broken. But at the same time the _identical_ latest release files and also all previous releases can be found at fixed location at `https://sourceforge.net/projects/freeplane/files/freeplane%20stable/archive`. It should be sufficient to take the release simply always from the archive and I do not understand why it is not like that.
> 
> There is no hash change with new release at `https://sourceforge.net/projects/freeplane/files/freeplane%20stable`, the files are named differently (they have version in them), so when there is new release, the installation fails on 404 error, not on non-matching hash.

Thanks for agreeing with me, that the static URLs are preferred to the dynamic "vanity" URL. And that the 404 error is preferrable to a hash error. I'm confused why the agreement was stated in the form of a correction and challenge, as though confrontation is necessary to conversation, which is untrue.

oldium:
@stephengillie I actually do not understand what you mean with static and dynamic "vanity" URLs. I just described where the developer is storing release files, it is consistent and predictable with every release (`https://sourceforge.net/projects/freeplane/files/freeplane%20stable/archive/<version>/Freeplane-Setup-with-Java-<version>.exe/download`) and I just think we should use that URL in WinGet manifests. Do not take it as an insult, it just described the situation and stated my recommendation.

I also have not expressed any preference on 404 error or hash error, I think (and I repeat myself) we should use the URL with "archive" in it to prevent any download errors. This is why I do not understand (and do not take it as insult) why this Pull Request was merged, when we know it uses URL, which breaks. I do not see this as a Freeplane developer's problem, it is this Pull Request's problem.

PR #118695, opened 2023-09-01
Rename betterbird 102.15.0-bb40 to 102.15.0

aaronliu0130:

---
See also #117917. I have no idea why @borque uses these version names but they aren't good for updates as winget still detects the version without the bb40. 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/118695)


Trenly (approved):

borque:
@aaronliu0130 it has a rather simple reason: if you look at betterbird.eu my versions are the ones officially used by the project They append bb and an increasing number to indicate the Betterbird-version of the corresponding Thunderbird series. So instead of renaming it completely as you propose, it might be better to use an AppsandFeatures-entry here?!
Besides: giving me a bit more time than less than 24h to react to your question https://github.com/microsoft/winget-pkgs/pull/118483#issuecomment-1701063141 would have been nice!

borque:
@aaronliu0130
I am still a bit unsure on how to use AppsAndFeatures correctly, but as I understand it, it should be the other way around:
DisplayVersion should be 102.15.0, as that is what is set in the ARP-entries and the manifest version should be 102.15.0-bb40 as this is the official version used by betterbird.eu "for marketing".
Perhaps @mdanish-kh could comment on this aspect, as he already tried to ;-) teach me a bit on how to use the ARP-entries? Or @Trenly ?
And perhaps @Betterbird could comment on the versioning scheme, and their view on imporance of the appended -bb.. for their versioning?

Betterbird:
Well, our versions are based on Thunderbird versions. To distinguish, we add -bbNN at the end. NN is the NNth release in the series, where series refers to 91, 102 and 115. Internally, the version is without the -bbNN, it's only used for display in the About box. Does that answer the question?

Compare `console.log(AppConstants.MOZ_APP_VERSION)` to `console.log(AppConstants.MOZ_APP_VERSION_DISPLAY)` pasted into the JS console in the error console.

We'll switch the official release to 115 next when we have all the localisations complete, 102 will be EOL at the end of September.

Trenly:
> DisplayVersion should be 102.15.0, as that is what is set in the ARP-entries and the manifest version should be 102.15.0-bb40 as this is the official version used by betterbird.eu "for marketing".

This is correcf

aaronliu0130:
@borque Currently what we have is that the DisplayVersion is set to 102.15.0 by default. If we have DisplayVersion set to 102.15.0 and manifest version set to 102.15.0-bb40, which was what happened before I made this PR, Winget will constantly detect an update for Betterbird. What we need to do is make DisplayVersion ≥ the manifest version.

Trenly:
> @borque Currently what we have is that the DisplayVersion is set to 102.15.0 by default because that's what the Betterbird installer does. If we have DisplayVersion set to 102.15.0 and manifest version set to 102.15.0-bb40, which was what happened before I made this PR, Winget will constantly detect an update for Betterbird. What we need to do is make DisplayVersion ≥ the manifest version. Judging by Betterbird's clarification of the versioning we don't need the bb, especially if their installer doesn't give us it in ARP's display by default.

Before this PR (and yet still) there is no DisplayVersion key set. If it were set, it should prevent the upgrade loop

aaronliu0130:
That doesn't make sense as 102.15.0 is still smaller than 102.15.0-bb40 lexicographically. Why would that change anything if the displayed ARP version is the same?

Trenly:
> That doesn't make sense as 102.15.0 is still smaller than 102.15.0-bb40 lexicographically. Why would that change anything if the displayed ARP version is the same?

When the DisplayVersion key is present, winget creates a map between the PackageVersion and DisplayVersion.  By setting the PackageVersion to 102.15.0-bb40 - that is the version that will be used for comparison, you are correct. However, by setting the DisplayVersion key as 102.15.0, when the package that is installed has the version of 102.15.0 - winget will treat it as if it were version 102.15.0-bb40.

aaronliu0130:
Ah, that makes sense.

Trenly:
> Ah, that makes sense. So the parameter doesn't set the displayed version but creates a map? Is there a way to make this clearer in the schema?

The schema simply sets what parameters are available, so I don’t know that it would be possible to make it clearer in the schema itself. However, it may be possible to improve the documentation on the [schema fields](https://github.com/microsoft/winget-pkgs/blob/master/doc/manifest/schema/1.5.0/installer.md#fields). 

DisplayVersion falls under the AppsAndFeaturesEntries, which is described as
> AppsAndFeaturesEntries are used to match installed packages with manifests in configured sources.

And then specifically for DisplayVersion - 
> DisplayVersion is the version displayed in the Control Panel for installed packages.

So taking those two together, the DisplayVersion is the version displayed in Control panel and is used to match installed packages with manifests in configured sources

aaronliu0130:
Yeah, I was talking about the documentation. I thought it would change ARP. So we add something like this for all locales in the installer manifest?
```
AppsAndFeaturesEntries:
	- DisplayName: Betterbird (x64 en-US)
	  DisplayVersion: 102.15.0
	  Publisher: Betterbird Project
```
image

Trenly:
> Yeah, I was talking about the documentation. I thought it would change ARP. So we add something like this for all locales in the installer manifest?
> 
> ```
> AppsAndFeaturesEntries:
> 	- DisplayName: Betterbird (x64 en-US)
> 	  DisplayVersion: 102.15.0
> 	  Publisher: Betterbird Project
> ```
> 
> image

Well, yes, and also no. Its complicated.

Because of how winget manages DisplayName, it actually doesn’t really care about the items in parentheses and would still find the match using just PackageName. That being said, because it is locale specific, it would be good to include the DisplayName.

For Publisher, it would be best to change the value in the locale files to match what is written in control panel. But, if that isn’t possible or doesn’t accurately portray the publisher, then it's best to include it in the AppsAndFeaturesEntries.

For DisplayVersion, it should only be included when it is different than the PackageVersion.

-----

Basically the mappings are like this:
PackageVersion <--> DisplayVersion
Publisher (defaultLocale) <--> Publisher
PackageName (defaultLocale) <--> DisplayName

PackageVersion and Publisher are exact match fields, so if they don’t match what is written to Control Panel exactly, then AppsAndFeaturesEntries are needed.

PackageName is a normalized match - so a PackageName `Betterbird` would match DisplayNames of `Betterbird version 1.0.0`, `Betterbird (x64 en-US)`, `Betterbird (remove-only)`, et cetera; but would not match `Betterbird Beta`.  However, if the value of DisplayName is specified in AppsAndFeaturesEntries, then that value is used for an exact match.

Trenly:
But for clarification, yes, the snippet of code you posted seems accurate

borque:
I think DisplayVersion is enough and the other parts of the AppsAndFeatures section can can be omitted here.
I discussed this with @mdanish-kh in this PR: https://github.com/microsoft/winget-pkgs/pull/115322#issuecomment-1666086995 . 
The conclusion was that the ARP section should be used as rarely as possible because none of the current manifest-creators can update it automatically and manifests containing it need manual post processing.

To summarize our results here: future manifest versions will be something like 115.3.0-bb11 but we add an ARP entry 115.3.0?

aaronliu0130:
Yeah I think that's fine. So just add
```
AppsAndFeaturesEntries:
	- DisplayVersion: 115.3.0
```
and change the locales' publishers to "Betterbird Project" and the same thing in other locales?

Trenly:
Yes, but only if the PackageVersion is the full version (including the -bb##), otherwise it causes extra load on the backend

PR #119914, opened 2023-09-15
New package: Beeradmoore.dlss-swapper version 1.0.2

NicolasConstant:

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/119914)


stephengillie (approved):

SpecterShell:
May I fix up some bugs and rename the identifier for this package? @NicolasConstant

NicolasConstant:
Wouldn't changing the identifier break the update process? Why changing it? 

For the bugs, sure. What are they?

SpecterShell:
WinGet detects installed programs by reading ARP entries (which are written by the installers and can be seen in Programs and Features in Control Panel) and matching them to the packages here. Itself doesn't record any packages (except the portable ones) installed by it. So it is safe to rename the identifier without breaking the update process. (It may break install process though, but it should be fine since this package has just been here for one week).

And to match the ARP entry to the package, the `PackageName` field in the manifest usually should be the same as the `DisplayName` ARP entry written by the installer, otherwise the package won't be detected and updated, which is the case here (with 1.0.0.0 installed):
image

NicolasConstant:
I see.
Since I only bootstrapped the profile to help the project's maintainer getting started, maybe @beeradmoore should be aware of this, and if important changes have to be made, it would be his call.

beeradmoore:
More than happy for people who know more than me to improve product and distribution. So long as someone can explain it to me like I'm 5 before we hit a big red button 😂

beeradmoore:
That update is now merged in, thanks @SpecterShell. Was there any changes you wanted to make the to manifest here? I figured I'd push 1.0.3.2 as it contains Ubisoft fixes and has sat around in pre-release for some time, but if there are other things to tidy up we may as well do both at once.

SpecterShell:
I was to rename the identifier to `Beeradmoore.dlss-swapper` to `beeradmoore.DLSSSwapper`, but if you think the triple "S" inside is hard to recognise then I wouldn't do it. 🙂
There will also be some fixes for the manifest.

beeradmoore:
I prefer the lower b. As for the rest of it I prefer the `-` (so `beeradmoore.dlss-swapper`). We have used a mixture of `-` or `_` as a separator to avoid the triple S. Reddit `DLSS_Swapper` (ironically moved from DLSSSwapper), Twitter `dlss-swapper`, repo itself is `diss-swapper`.

Installer files themselves are `DLSS.Swapper-1.0.2.0-installer.exe` because felt odd to have a space, and having it be `DLSS-Swapper-1.0.2.0-installer.exe` also felt like too many dashes 😅

PR #120764, opened 2023-09-25
Add version: ChristianHohnstadt.xca version RELEASE.2.5.0

vedantmgoyal9:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/120764)

### DIFF
--- a/manifests/c/ChristianHohnstadt/xca/RELEASE.2.5.0/ChristianHohnstadt.xca.installer.yaml
+++ b/manifests/c/ChristianHohnstadt/xca/RELEASE.2.5.0/ChristianHohnstadt.xca.installer.yaml
@@ -1,0 +1,20 @@
+# Created with WinGet Automation using Komac v1.11.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.5.0.schema.json
+
+PackageIdentifier: ChristianHohnstadt.xca
+PackageVersion: RELEASE.2.5.0
+InstallerLocale: en-US
+InstallerType: wix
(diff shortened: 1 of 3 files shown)


stephengillie (approved):

tconex:
@stephengillie @vedantmgoyal2009
I think Versioning Scheme is not right for this PR and the last one on this product.
image
This is just a hint as I was slightly confused when updating, please ignore if not relevant.

stephengillie:
Hi @tconex,

It appears that you are correct:
image

br2soft:
Versioning Scheme is indeed not right for this PR. With a version sting "RELEASE.x.y.z" the CLI gets confused. Even when I try to install it by hand with the MSI release 2.5.0 published on the website winget thinks it is younger than 2.4.0 what is not the case. The display version on Windows Apps Page shows 2.5.0, but winget displays "RELEASE.2.5.0". Please use standard Versioning Scheme.

mrjcap:
Hopefully PR https://github.com/microsoft/winget-pkgs/pull/121902 fixes this issue.

Trenly:
It's worth noting that there is nothing inherently wrong with using `RELEASE.2.5.0` as the PackageVersion, since the `AppsAndFeaturesEntries` are included in the manifest.  However, because there are already existing versions that do not have the `RELEASE.`, it will cause errors with WinGet being able to detect the upgrade.  Either way is correct, as long as it is consistent across all versions of the package

PR #120992, opened 2023-09-27
Update version: Element.Element version 1.11.44

vedantmgoyal9:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/120992)


Trenly (changes requested):
Release notes URL should be added back

kbdharun (commented):
> Release notes URL should be added back

Agreed, also I have no idea why the action creates PRs for removing just the release links. All other things match with the existing file. @vedantmgoyal2009 If possible, can you fix this edge case as this isn't the first PR the action did this.

stephengillie (approved):

PR #121482, opened 2023-10-01
Delete manifests/s/ShiningLight/OpenSSL/Light/1.1.1w

JohnMcPMS:
A recent change by @szhu25 introduced `DisplayVersion` values for all 3 versions of this package that are all `1.0.0`.  This makes the update process unhappy; removing the older two versions for now.  If possible, I would suggest updating the installer to write the appropriate version.  If not, then the only choice is to remove the `DisplayVersion` again.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/121482)


denelon (approved):

stephengillie:
>  Sequence contains no elements

stephengillie:
[This package](https://github.com/microsoft/winget-pkgs/blob/master/manifests/s/ShiningLight/OpenSSL/Light/1.1.1w/ShiningLight.OpenSSL.Light.installer.yaml) had the DisplayVersion removed - could these errors be from trying to remove an already-removed line?

szhu25:
@stephengillie I noticed the issue after @JohnMcPMS posted this PR, and the fix got merged faster than this one... So I think this might be able to be closed? (If the conflict is already resolved)

PR #121770, opened 2023-10-03
CloudNet 1.36.2.4

dayunet2009:
CloudNet
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/121770)


stephengillie:
I believe the [Policy 2.5](https://learn.microsoft.com/en-us/windows/package-manager/package/windows-package-manager-policies#25-offensive-content) check is due to the city name in the Publisher, but speculating on why might invoke controversy.

PR #121895, opened 2023-10-04
Update version: OpenWhisperSystems.Signal version 6.32.0

vedantmgoyal9:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/121895)


stephengillie:
@vedantmgoyal2009, this change keeps getting tried. The only change is to downgrade the architecture to x86, but it requires x64 to run. So it appears to be an errant PR.

Trenly:
Close with reason: Bad PR;

PR #122259, opened 2023-10-10
Removed all old versions that are no longer supported for ORMCore

skdataas:
A removal of all the old versions of ORM that are no longer supported, leaving only the latest one. 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/122259)


Trenly:
Please submit separate PRs for each version.

skdataas:
The problem is, they are screwing things up for winget, our server sends you the latest version, but somehow you check version numbers in string and not in "last updated", so an old version is requested with the wrong hash key. 

I know you wish a PR request for each delete, but basically all versions but the latest added can be removed.  I really don't see why this is such a problem when clearly it's something on your side where you don't use the latest version added as your check.

PR #122574, opened 2023-10-13
Add `RequireExplicitUpgrade: true` to Chocolatey 2.2.0.0

mdanish-kh:
Since the package is meant for installation only - 

https://github.com/microsoft/winget-pkgs/blob/825581972c7c8ebd2cea7ea0f490dea6055e65d2/manifests/c/Chocolatey/Chocolatey/2.2.2.0/Chocolatey.Chocolatey.locale.en-US.yaml#L15 

We can remove this package from `winget upgrade --all` flow by adding `RequireExplicitUpgrade: true`

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/122574)

### DIFF
--- a/manifests/c/Chocolatey/Chocolatey/2.2.0.0/Chocolatey.Chocolatey.installer.yaml
+++ b/manifests/c/Chocolatey/Chocolatey/2.2.0.0/Chocolatey.Chocolatey.installer.yaml
@@ -5,6 +5,7 @@ PackageIdentifier: Chocolatey.Chocolatey
 PackageVersion: 2.2.0.0
 InstallerLocale: en-US
 Scope: machine
+RequireExplicitUpgrade: true
 AppsAndFeaturesEntries:
 - DisplayName: Chocolatey (Install Only)
   UpgradeCode: '{62FC4E65-47F0-488A-B841-ECA07E98A58B}'


stephengillie:
Hi @pauby,

Do you approve of this change?

Trenly:
@mdanish-kh - I would prefer we set `UpgradeBehavior: deny` once 1.6 is released to pipelines

mdanish-kh:
> @mdanish-kh - I would prefer we set `UpgradeBehavior: deny` once 1.6 is released to pipelines

Had that in mind but I imagine 1.6 schema is still some months away. Oh well... since no one has made an issue for this, guess we can wait for it and do it the correct way

pauby:
@stephengillie I don't have anything against this change. It makes sense. If, as @mdanish-kh says, that 1.6 is some months away, then I don't see any reason to wait for it and this be merged now.

mdanish-kh:
Well my choice of words wasn't all that great. I'm not a part of the Winget team, so I was more of guessing that it has to be sometime around the release of 1.7 WinGet client which I also guess is some months away 😆 

Regardless of my poor choice of words and since @pauby also approves of this change, I think we can merge this PR right now to at the very least block the package from `winget upgrade --all`. When `UpgradeBehavior: deny` is available in the validation pipelines, we can simply raise another PR to add that field in.

stephengillie (approved):

PR #122673, opened 2023-10-14
fix: align Eclipse Temurin definitions with each other

Okeanos:
- use consistent order of keys
- use consistent definition of keys (where applicable)
- use consistent general/generic keys
- add some missing keys (where applicable) such as the install switch to modify the installation location

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/122673)

### DIFF
--- a/manifests/e/EclipseAdoptium/Temurin/11/JRE/11.0.20.8/EclipseAdoptium.Temurin.11.JRE.installer.yaml
+++ b/manifests/e/EclipseAdoptium/Temurin/11/JRE/11.0.20.8/EclipseAdoptium.Temurin.11.JRE.installer.yaml
@@ -3,7 +3,9 @@
 
 PackageIdentifier: EclipseAdoptium.Temurin.11.JRE
 PackageVersion: 11.0.20.8
+ReleaseDate: 2023-07-21
 InstallerLocale: en-US
+MinimumOSVersion: 10.0.0.0
 InstallerType: wix
 Scope: machine
 InstallModes:
@@ -11,6 +13,7 @@ InstallModes:
 - silent
 - silentWithProgress
 InstallerSwitches:
+  Custom: Installlevel=3
   InstallLocation: INSTALLDIR="<INSTALLPATH>"
 UpgradeBehavior: install
 Commands:
@@ -21,7 +24,6 @@ FileExtensions:
 - jar
 - java
 - jsp
-ReleaseDate: 2023-07-21
 Installers:
 - Architecture: x86
   InstallerUrl: https://github.com/adoptium/temurin11-binaries/releases/download/jdk-11.0.20+8/OpenJDK11U-jre_x86-32_windows_hotspot_11.0.20_8.msi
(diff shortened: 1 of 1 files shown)


manifests/e/EclipseAdoptium/Temurin/11/JRE/11.0.20.8/EclipseAdoptium.Temurin.11.JRE.installer.yaml
```diff
@@ -3,14 +3,17 @@
 
 PackageIdentifier: EclipseAdoptium.Temurin.11.JRE
 PackageVersion: 11.0.20.8
+ReleaseDate: 2023-07-21
 InstallerLocale: en-US
+MinimumOSVersion: 10.0.0.0
```
russellbanks, line 8:
```suggestion
```

The `MinimumOSVersion` is not needed.

russellbanks (changes requested):

manifests/e/EclipseAdoptium/Temurin/11/JRE/11.0.20.8/EclipseAdoptium.Temurin.11.JRE.installer.yaml
Okeanos: That's a fair point that I did consider very briefly. I am wondering whether removing this (or keeping it) actually does anything? Is there any client side software using the winget repository NOT on >= Win 10? I can absolutely update the definitions, no problem.

Why did I add it in the first place? This is an amalgamation of all the other definitions, some of which already had it, and ultimately I think I should have validated some of these values a little more I guess.

Okeanos: So, the 8.1 version string would be `6.3.9600` or is omitting it the better idea altogether?

russellbanks: > Is there any client side software using the winget repository NOT on >= Win 10?

I don't think so but even then the entry is redundant.

> This is an amalgamation of all the other definitions, some of which already had it

Manifests created with yamlcreate add that entry but I don't know why because it shouldn't be needed. If it is needed, the pipelines will give a https://github.com/microsoft/winget-pkgs/labels/Manifest-Installer-Validation-Error - I've only had this occur for MSIXs that have that hard minimum requirement on a specific Windows 10 version.

> So, the 8.1 version string would be `6.3.9600` or is omitting it the better idea altogether?

Omitting it would be best as there shouldn't be anything running winget below Windows 10 and that minimum Temurin has on their website is probably for support - it could well work on Windows 7, for example.

russellbanks (approved):

PR #123692, opened 2023-10-25
Add AppsAndFeaturesEntries in PowerShell Preview 7.4.0.4

mdanish-kh:
Related to https://github.com/microsoft/winget-pkgs/pull/123657#pullrequestreview-1696907836

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/123692)

PR #123693, opened 2023-10-25
Add AppsAndFeaturesEntries in PowerShell Preview 7.4.0.5

mdanish-kh:
Related to https://github.com/microsoft/winget-pkgs/pull/123657#pullrequestreview-1696907836

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/123693)

### DIFF
--- a/manifests/m/Microsoft/PowerShell/Preview/7.4.0.5/Microsoft.PowerShell.Preview.installer.yaml
+++ b/manifests/m/Microsoft/PowerShell/Preview/7.4.0.5/Microsoft.PowerShell.Preview.installer.yaml
@@ -19,11 +19,19 @@ Installers:
   InstallerUrl: https://github.com/PowerShell/PowerShell/releases/download/v7.4.0-preview.5/PowerShell-7.4.0-preview.5-win-x64.msi
   InstallerSha256: 1C09C4765C635B36CB17FAB4B14914B37D3B893ACD0A22B69A87AF7A1337E11C
   ProductCode: '{5172C645-89F1-476D-A58C-F86584D3A784}'
+  AppsAndFeaturesEntries:
+  - Publisher: Microsoft Corporation
+    DisplayName: PowerShell 7-preview-x64
+  ElevationRequirement: elevatesSelf
 - Architecture: x86
   InstallerType: wix
   InstallerUrl: https://github.com/PowerShell/PowerShell/releases/download/v7.4.0-preview.5/PowerShell-7.4.0-preview.5-win-x86.msi
   InstallerSha256: 501FF76E9FAF08A2AD038BF72699A3945DB938CC42776FF857941E6CF092EEC7
   ProductCode: '{39EB56F7-8C79-4A6B-8929-E7E211DE40C5}'
+  AppsAndFeaturesEntries:
+  - Publisher: Microsoft Corporation
+    DisplayName: PowerShell 7-preview-x86
+  ElevationRequirement: elevatesSelf
 - Architecture: arm
   MinimumOSVersion: 10.0.17763.0
   Platform:


mdanish-kh:
[Policy] Needs-Manual-Merge

Because of https://github.com/microsoft/winget-pkgs/pull/111849#issuecomment-1633345701

stephengillie:
Sorry for the delay on these.

stephengillie:
Hi @anamnavi,

Do you approve of this change?

mdanish-kh:
> Sorry for the delay on these.

No worries at all :)

anamnavi:
I'm a bit confused with regards to this and the original issue. My understanding was that winget upgrades by looking at the `PackageIdentifier` property not `DisplayName` in `AppsAndFeaturesEntries`. And `PackageIdentifier` is `Microsoft.PowerShell` for stable releases and `Microsoft.PowerShell.Preview` for preview releases. Do we need separate DisplayNames by architecture for preview and stable?

mdanish-kh:
> I'm a bit confused with regards to this and the original issue. My understanding was that winget upgrades by looking at the `PackageIdentifier` property not `DisplayName` in `AppsAndFeaturesEntries`. And `PackageIdentifier` is `Microsoft.PowerShell` for stable releases and `Microsoft.PowerShell.Preview` for preview releases. Do we need separate DisplayNames by architecture for preview and stable?

@anamnavi Actually WinGet gets its package matching metadata from the Registry. Properties such as DisplayVersion, PackageName, ProductCode etc. are taken locally from the registry and then the metadata is mapped to an existing manifest here in winget-pkgs (these are some of the fields you can also see in Control Panel / Apps&Features). In this case the PackageName added here in `Microsoft.PowerShell.Preview` manifests is `PackageName: PowerShell Preview` and since winget uses fuzzy matching for PackageName field, this has a chance to collide with `PackageName: PowerShell` that's used in `Microsoft.PowerShell` manifests. We can mitigate this by providing `AppsAndFeaturesEntries` and exactly match the name written to Registry. By doing this, `PackageName` can serve as the marketing name that users get to see in `winget search / show / install / list` but under the hood winget will use `AppsAndFeatures: DisplayName` for package matching.

PackageIdentifier is nothing but a unique ID given to every package that's generally in the form `<Publisher>.<PackageName>` so that its easier to remember/target a given package. It doesn't play a role in package matching.

anamnavi:
I see, that explanation makes a lot of sense. I think our team's understanding of this has been incorrect so far. We thought that providing a DisplayName property would break users because it differs from the PackageName. So that's why we have most of our manifests without the `DisplayName` and only the msixbundle installers for arm and arm64 have AppsAndFeaturesEntries.

If going forward we include the DisplayName and it varies between PowerShell and PowerShell Preview and also by architecture, could it break users in any way? Just trying to ascertain if there's any risk for that. If not we can start including that info going forward, thanks!

mdanish-kh:
It's a good idea to include `AppsAndFeatureEntries` in the manifest where possible. Some of the common reasons why we don't include it for every manifest are:

- PackageName, PackageVersion already provide a good match with the actual Registry metadata and we don't need to separate marketing versions/names from actual version/names
- Makes the manifest more complex for new contributors
- Sometimes the complete name written to registry is version specific i.e. something like `MyApp 2.3`. The DisplayName field then needs to be manually updated for every new manifest (current manifest creators don't have the ability to automate these fields). So generally we avoid putting DisplayName in those cases when the PackageName already provides a good enough match.


> If going forward we include the DisplayName and it varies between PowerShell and PowerShell Preview and also by architecture, could it break users in any way?

Not at all, it would actually help winget target the correct package for future updates. Also we don't _particularly need_ to add this for the `Microsoft.PowerShell` manifests although there's no harm if we do.

anamnavi:
thanks for the detailed info @mdanish-kh. One question, what is the purpose of the `ElevationRequirement: elevatesSelf` line? Is it required?

It seems like to distinguish between Microsoft.PowerShell.Preview and Microsoft.PowerShell we need to at least include `DisplayName` in Microsoft.PowerShell.Preview entries. 

We will also work to include it in Microsoft.PowerShell manifests to reduce ambiguity where possible.

anamnavi (approved):

mdanish-kh:
> thanks for the detailed info @mdanish-kh. One question, what is the purpose of the `ElevationRequirement: elevatesSelf` line? Is it required?
> 

@anamnavi It's an optional field that tells WinGet how the installer behaves when it comes to [UAC elevation](https://learn.microsoft.com/en-us/windows/security/application-security/application-control/user-account-control/how-it-works#uac-elevation-prompts).

`ElevationRequirement: elevationProhibited`: Installer prohibits elevation / running from an admin shell (only installs in a user shell)
`ElevationRequirement: elevatesSelf`: If installed is invoked in a user shell, the installer will prompt to elevate.
`ElevationRequirement: elevationRequired`: Installer requires that the installation is from an admin shell (won't work in user shell  and won't prompt)

For now, winget will this information to print out a message like "The installer will request to run as administrator, expect a prompt." in case of `elevatesSelf` but further down the line winget may be able to do more with the information so having the manifest as complete as possible is preferred :

image
 

> It seems like to distinguish between Microsoft.PowerShell.Preview and Microsoft.PowerShell we need to at least include `DisplayName` in Microsoft.PowerShell.Preview entries.
> 

That's correct!

mdanish-kh:
@stephengillie Looks like this PR, #123692, #123694 can be merged now 👀

stephengillie (approved):

PR #124602, opened 2023-11-02
Republish Microsoft.Azd 1.4.400 (package 1.4.3)

### LINKED ISSUE #124600: Can't install Microsoft.Azd latest version 1.4.3
despite the 1.4.400 manifest version being [merged last week](https://github.com/microsoft/winget-pkgs/pull/123748), 1.4.300 is still what is found. Also, winget even says a newer version is available, but it doesn't show up when running winget upgrade

image

heaths:
Reverts #123987, resolves #124600

Because Windows Installer `ProductVersion` only [supports 3 fields that affect upgrade behavior](https://learn.microsoft.com/windows/win32/msi/productversion), we - like many other products - pack a semver with optional prerelease metadata into a dotted triple using the algorithm: `major.minor.(patch * 100 + preview version)`.

So the manifest version 1.4.400 is `azd` version 1.4.3 which was published by our pipeline prior to this manifest being inserted.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/124602)

### DIFF
--- a/manifests/m/Microsoft/Azd/1.4.400/Microsoft.Azd.installer.yaml
+++ b/manifests/m/Microsoft/Azd/1.4.400/Microsoft.Azd.installer.yaml
@@ -1,0 +1,16 @@
+# Created using wingetcreate 1.5.5.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.5.0.schema.json
+
+PackageIdentifier: Microsoft.Azd
+PackageVersion: 1.4.400
+InstallerLocale: en-US
+InstallerType: wix
+InstallerSwitches:
+  Custom: INSTALLEDBY="winget"
+ProductCode: '{487AFABD-E830-460B-944A-5B3470AEA2AA}'
+Installers:
+- Architecture: x64
+  InstallerUrl: https://github.com/Azure/azure-dev/releases/download/azure-dev-cli_1.4.3/azd-windows-amd64.msi
+  InstallerSha256: 84484B3F394BA7DAAA2D51CFAA92F314376237878958C4ADA198A1C9C26A3DF9
+ManifestType: installer
+ManifestVersion: 1.5.0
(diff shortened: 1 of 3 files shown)


denelon (approved):

denelon:
@kkamegawa you removed this manifest previously. I believe there was some confusion about the version of AZD.

denelon:
@stephengillie this manifest is usually published by azure-sdk we should add this to our watch list, so we don't have another similar regression. @heaths is on the team supporting the publishing of the AZD package.

rajeshkamal5050:
@stephengillie @kkamegawa anything pending on this? can we get this merged?

BrandonWanHuanSheng:
Could be that one is a bug of approval process.

BrandonWanHuanSheng:
Could be that one is a bug of approval process.

stephengillie:
> @ stephengillie this manifest is usually published by azure-sdk we should add this to our watch list, so we don't have another similar regression. @heaths is on the team supporting the publishing of the AZD package.

@denelon and @heaths, 

Currently [Auth.csv](https://github.com/microsoft/winget-pkgs/blob/master/Tools/Auth.csv) has :
```
PackageIdentifier | Account   | strictness
...
Microsoft.Azd     | azure-sdk | should
...
```
I'll change the `strictness` to `must`. Should `azure-sdk` remain an uploader, or should only @heaths be allowed to upload these?

heaths:
@danieljurek and I work on the package build and pipelines and I only reverted the revert because our pipelines only publish anew. If making this `must` would've prevented some other person from deleting it (reverting the add) in the first place, that seems reasonable; though, is there any recourse to being able to update it? Talking with @denelon offline, there's a possibility we may want to go and update existing manifest - either delete some older ones, or change data - and having to do that through a one-off pipeline feels burdensome.

stephengillie:
Hi @heaths,

Yes, we can add multiple approved uploaders. This is for a scripted manifest checking tool that I have written as a stopgap until the Verified Publishers system is complete. When a manifest with a matching PackageIdentifier is checked, it creates a warning message that only the listed submitters are allowed. 

I'm not aware of the specific implementation of the Verified Publishers system, but I'm hopeful that it will also allow multiple submitters in some way.

PR #125221, opened 2023-11-08
Update version: Microsoft.PowerShell version 7.3.9.0

vedantmgoyal9:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/125221)


stephengillie:
Hi @anamnavi,

Do you approve of these changes? The main change seems to be adding the artifact hashes as release notes.

anamnavi:
@stephengillie sorry for the delayed response. This change is fine. It adds the `DisplayName` field under `AppsAndFeaturesEntries`...do we need to go back and add this to prior (stable release) manifests? 

@mdanish-kh I know we said we need to add this for preview release manifests (I've attached the PR with the discussion for that below), but I'm wondering about stable releases in the past, will it upgrade correctly if past versions don't have this change? We'd like to add DisplayName going forward for stable releases anyways, I just want to make sure if we need to update older releases to not break upgrading I do so,

RE the discussion here: https://github.com/microsoft/winget-pkgs/pull/123693

mdanish-kh:
@anamnavi Past versions do not need to have this change. As long as the DisplayName is the correct value (identical to what's written to Control Panel/Registry), it won't break any upgrade scenarios.

mdanish-kh:
Close with reason: See above;

PR #129167, opened 2023-12-04
Improve how YamlCreate writes to files

Trenly:

Previously, YamlCreate would write the header to a file, then append the manifest content, read the content back in, perform some transforms on the raw content, re-write it all back out to file, read the content back in, and re-write it all back out *again* with the proper encoding. This is highly inefficient due to the extra time spent writing to file and reading from file and only gets worse the larger manifest files get.

This PR changes the flow to write all the content in one shot by performing the final content transformations in memory. Since the conversion to raw content would be happening anyways, there is no increased resource utilization of the script. It also abstracts the common code to a function and makes it a little bit easier to follow exactly what the code is doing.

cc @mdanish-kh @denelon @stephengillie
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/129167)

### DIFF
--- a/Tools/YamlCreate.ps1
+++ b/Tools/YamlCreate.ps1
@@ -2084,7 +2084,7 @@ Function Write-ManifestContent {
     [string] $FilePath,
     [Parameter(Mandatory = $true, Position = 1)]
     [PSCustomObject] $YamlContent,
-    [Parameter(Mandatory = $true, Position = 1)]
+    [Parameter(Mandatory = $true, Position = 2)]
     [string] $Schema
   )
   [System.IO.File]::WriteAllLines($FilePath, @(
--- a/Tools/YamlCreate.ps1
+++ b/Tools/YamlCreate.ps1
@@ -163,7 +163,7 @@ if ($Settings) {
   exit
 }
 
-$ScriptHeader = '# Created with YamlCreate.ps1 v2.2.10'
+$ScriptHeader = '# Created with YamlCreate.ps1 v2.2.11'
 $ManifestVersion = '1.5.0'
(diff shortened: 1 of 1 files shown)


denelon (approved):

Tools/YamlCreate.ps1
```diff
@@ -2077,6 +2077,26 @@ Function Get-DebugString {
...
+    [Parameter(Mandatory = $true, Position = 0)]
+    [string] $FilePath,
+    [Parameter(Mandatory = $true, Position = 1)]
+    [PSCustomObject] $YamlContent,
+    [Parameter(Mandatory = $true, Position = 2)]
+    [string] $Schema
+  )
+  [System.IO.File]::WriteAllLines($FilePath, @(
+      $ScriptHeader + $(Get-DebugString);
+      "# yaml-language-server: `$schema=$Schema)";
```
mdanish-kh, line 2092:
```suggestion
      "# yaml-language-server: `$schema=$Schema";
```
This looks to be a typo. `)`

Tools/YamlCreate.ps1
```diff
@@ -2077,6 +2077,26 @@ Function Get-DebugString {
...
+      '';
+      $(ConvertTo-Yaml $YamlContent).TrimEnd() -replace "(.*)$([char]0x2370)", "# `$1"
```
mdanish-kh, line 2094: I would suggest adding a comment explaining how the regex is working (commenting out special characters?)

mdanish-kh (changes requested):

mdanish-kh:
Ah, the timing 😄

PR #130038, opened 2023-12-11
Create DSC for bootstrapping YamlCreate.ps1

Trenly:

cc @denelon / @stephengillie
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/130038)


denelon:
I love taking the approach of building a WinGet configuration here. I've been looking and haven't found anything that gives the SKU of Windows to make sure Windows SandBox is available.

It would be nice if we could somehow deal with the scenario where a Windows SKU doesn't have the ability to run the Windows Sandbox rather than just having the configuration fail in that case.

@ryfu-msft have you found anything that can do this?

Trenly:
> I love taking the approach of building a WinGet configuration here. I've been looking and haven't found anything that gives the SKU of Windows to make sure Windows SandBox is available.
> 
> It would be nice if we could somehow deal with the scenario where a Windows SKU doesn't have the ability to run the Windows Sandbox rather than just having the configuration fail in that case.
> 
> @ryfu-msft have you found anything that can do this?

It may be possible to use the `Script` DSC resource to write a custom Get, Test, and Set function but that's a fair bit of work. There are also ways to enable sandbox in Home Edition, but they involve manually registering all the containers packages with DISM and then using DISM again to enable the feature.

denelon:
It might be worth just documenting that via comments and pulling it out of this configuration since it would likely cause the configuration to fail. Maybe a different configuration could be used to bootstrap Sandbox appropriately and launch it.

denelon (approved):

denelon:
Nice fix for the "project file" thing too!

PR #130481, opened 2023-12-14
Update MicaForEveryone.MicaForEveryone.installer.yaml URL

rakleed:

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/130481)

### DIFF
--- a/manifests/m/MicaForEveryone/MicaForEveryone/1.3.1.2/MicaForEveryone.MicaForEveryone.installer.yaml
+++ b/manifests/m/MicaForEveryone/MicaForEveryone/1.3.1.2/MicaForEveryone.MicaForEveryone.installer.yaml
@@ -9,13 +9,13 @@ ReleaseDate: 2023-07-15
 Installers:
 - Architecture: x64
   Scope: user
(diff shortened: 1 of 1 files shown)


Trenly (changes requested):
The URLs should be pointed at the specific version and not at latest. If the Latest version changes, it will cause the package to become broken.

The URLS existing in the manifest should not be changed

Trenly:
> > The URLs should be pointed at the specific version and not at latest. If the Latest version changes, it will cause the package to become broken.
> 
> @Trenly but why? If I specify the URL to the latest version, then your system could automatically detect a new version when the installer hash will change.

The detection system is not perfect, and it will also remove old versions. It is intended only for Vanity URLs when there is not a versioned URL available.  There are other tools available for detecting new GitHub releases such as @vedantmgoyal2009's automation or his WinGetReleaser action

stephengillie:
I'll add this as a buggy solution. It should be removed and replaced with a more robust solution in the future.

stephengillie (approved):

PR #130758, opened 2023-12-17
New package: Fort.Firewall version 3.10.3

ghost:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/130758)


Emi-Emi-Emi:

Emi-Emi-Emi:
I used Komac, which for some reason doesn't recognize the installer as _inno_, but as **exe** with `/SILENT` works fine.

I tested both, normal and with `-i` and it worked as expected.

Now, two possible issues that someone has to tell me if I have to tell the Dev to fix.

When the program is installed, and running, the installer doesn't close the program and actually gives the notice how the Program is running but once the program is closed, it installs silently with no issues.

The other issues as mentioned by the manifest in the Description 
>Important: Fort has its own kernel driver based on the Windows Filtering Platform (WFP) and it is not Compatible with HVCI (Core Isolation).

That problem has to do with the EV certificate, because for a single dev of an open source project is hard to buy a new EV certificate to sign the driver. That's why I added the notice, but the installer will throw an error of being incompatible if people have it enabled.

While people might argue that this is 'more insecure', Microsoft has this documentation [Options to Optimize Gaming Performance in Windows 11](https://support.microsoft.com/en-us/windows/options-to-optimize-gaming-performance-in-windows-11-a255f612-2949-4373-a566-ff6f3f474613) which advices people to turn it off. Because people have the choice to do it.

Without Core isolation, the program installs fine, and then you can't enable core isolation if the driver is detected as 'not compatible'.

Of course it is someone's decision to allow or not Fort Firewall, but these are the only two issues I can think of, I tested the manifest in modern computer with disabled Core isolation support, and an older computer without it and it install fine.

If there is a problem about an unsigned driver because lack of resources of an open source project to get the EV certificate, it will be understood. Developer is a good person, and he understand the incompatibility with HVCI can cause issues.

Thank you.

Emi-Emi-Emi:
I used Komac, which for some reason doesn't recognize the installer as _inno_, but as **exe** with `/SILENT` works fine on my side.

I tested both, normal = silent and with `-i` and it worked as expected.

Now, possible issues for this unattended notice.

`Visual C++ x86 redistributable` is needed, if not the installer will throw an error and give the URL to install it. But since most people already should have that installed, because many programs use it... is that an issue? Do the installer have to include it?

The other issue, which I think it is the problematic one is the one I mentioned in the manifest, in the Description 
>Important: Fort has its own kernel driver based on the Windows Filtering Platform (WFP) and it is not Compatible with HVCI (Core Isolation).

If this is the issue, well, that problem has to do with the EV certificate, and that's because for a single dev of an open source project is hard to buy / pay for a EV certificate to sign the driver yearly, I based this information on [this Microsoft documentation for the code signing](https://learn.microsoft.com/en-us/windows-hardware/drivers/dashboard/code-signing-reqs), where it gives links to sites where you can get a certificate and like I said, open source projects can't easily afford that.

That's why I added the notice in the manifest, but the dev made it impossible to install if Core Isolation is enabled, so it won't cause issues if people try to install it, then they can decide if they disable Core Isolation or not.

But with Core Isolation disabled and VCRedist installed,  it installs _silently_ as expected.

Of course it will be someone to tell me if Fort Firewall can be in Winget, or what changes can be done for it to be allowed for future reference. Because these are the only issues I can think of, that could cause this problem.

Thank you.

Emi-Emi-Emi:
I unassigned myself because there is nothing I can do, so I will wait for feedback from any admin to see Fort Firewall situation and Winget requirements.

using cURL and Terminal would be less troubled than what Winget is been, but I would like to make it easier for Fort Firewall users to install and update Fort since Winget is now included in every Win 11 installation.

Emi-Emi-Emi:
@denelon 

Sorry to bother you, but can you tell me if this type of program is allowed to be added in Winget?

I just closed this PR because Fort got a new update today, so there is no much point of adding an already old version, when the new version added an amazing feature and all that, but I don't know if I should make a 3.10.4 PR, because the system might complain about it and tag it with `Validation-Unattended-Failed` again and go back in a cycle of not even knowing if this would be allowed it based on Winget guidelines.

I don't know if this firewall with a custom driver that is not properly signed because signing drivers doesn't seem cheap, so core isolation has to be turned off for it to wok can be allowed or if `Visual C++ x86 redistributable` has to be embedded in the installer.

I tried to find a policy or other programs with the same issue, and didn't find any, didn't even find other programs with similar issues.

If I knew this, I would have never tried to upload it to Winget. Now I got this 'closed' PR with an awful red button haunting me forever unless I delete/re-create my account.

denelon:
You can add the Visual C++ redistributable to the manifest as a dependency for the package. 

Validation-Unattended-Failed is usually because some part of the install flow is expecting user input so we're not getting the expected return code of 0 after the install should have completed. The [Windows Sandbox Test Script](https://github.com/microsoft/winget-pkgs/blob/master/Tools/SandboxTest.ps1) might also be a way for you to test the PR in the Windows Sandbox using WinGet.

denelon:
Once you get a working manifest, you can follow the example in [PowerToys](https://github.com/microsoft/PowerToys/blob/main/.github/workflows/package-submissions.yml) or [Oh-My-Posh](https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/.github/workflows/winget.yml) to automate publishing the package when new releases are available.

Emi-Emi-Emi:
@denelon

I will add the dependency in the manifest then and test it.

But what about the driver? is there no problem with the fact that people have to go and turn Core Isolation's - Memory Integrity off? Since it can be seen as a security issue? what are the policies about non properly EV signed drivers for Winget?
You know the one in Win11 `Settings` -> `Privacy & Security` -> `Device Security` -> `Core Isolation` -> `Memory Integrity` and then computer has to be restarted.

Without turning it off the installer will not proceed, this was added because if not the Fort driver would throw a bluescreen and all that, but due to economical reasons no proper signed driver.

Can that throw the error `Validation-Unattended-Failed`? 
I will test adding the Visual C++ redistributable dependency, but since that computer is a little old and doesn't support core isolation, I am not even sure if Windows Sandbox will throw an error for it or not or if Winget policies will not want a program without a proper signed driver on it that requires some security feature to be turned off to work.

denelon:
As long as the package will install and "launch" without crashing, you can add Installation notes to tell users what they need to do in order to get it to work properly after installation. If the install crashes without those settings in advance, then it's not going to be able to pass validation.

You can dig into the build output to see logs and potentially a screenshot of what the validation system is capturing. If the unsigned drivers require user interaction, that's likely where the failure is happening.

PR #130892, opened 2023-12-18
NI.ni-daqmx version 23.8.0

lars-nvhgroup:

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/130892)

### DIFF
--- a/manifests/n/NI/ni-daqmx/23.8.0/NI.ni-daqmx.installer.yaml
+++ b/manifests/n/NI/ni-daqmx/23.8.0/NI.ni-daqmx.installer.yaml
@@ -1,0 +1,27 @@
+# Created using wingetcreate 1.5.7.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.5.0.schema.json
+
+PackageIdentifier: NI.ni-daqmx
+PackageVersion: 23.8.0
+MinimumOSVersion: 10.0.0.0
+InstallerType: exe
+Scope: machine
+InstallModes:
+- interactive
+- silent
+InstallerSwitches:
+  Silent: --quiet --accept-eulas
+  SilentWithProgress: --quiet --accept-eulas
+UpgradeBehavior: install
+Dependencies:
+  PackageDependencies:
+  - PackageIdentifier: NI.ni-packagemanager
+    MinimumVersion: 23.8.0
+InstallLocationRequired: false
+ElevationRequirement: elevationRequired
+Installers:
+- Architecture: x64
+  InstallerUrl: https://download.ni.com/support/nipkg/products/ni-d/ni-daqmx/23.8/online/ni-daqmx_23.8_online.exe
+  InstallerSha256: 3D59496FAF29096236A3F517CE38CE4F5FD8916946794B2E63B58C5FC9D1E9B1
+ManifestType: installer
+ManifestVersion: 1.5.0
--- a/manifests/n/NI/ni-daqmx/23.8.0/NI.ni-daqmx.locale.en-US.yaml
+++ b/manifests/n/NI/ni-daqmx/23.8.0/NI.ni-daqmx.locale.en-US.yaml
@@ -1,0 +1,13 @@
+# Created using wingetcreate 1.5.7.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.5.0.schema.json
+
+PackageIdentifier: NI.ni-daqmx
+PackageVersion: 23.8.0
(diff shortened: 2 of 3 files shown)


manifests/n/NI/ni-daqmx/23.8.0/NI.ni-daqmx.installer.yaml
```diff
@@ -0,0 +1,27 @@
...
+PackageIdentifier: NI.ni-daqmx
+PackageVersion: 23.8.0
+MinimumOSVersion: 10.0.0.0
+InstallerType: exe
+Scope: machine
+InstallModes:
+- interactive
+- silent
+InstallerSwitches:
+  Silent: --quiet --accept-eulas
```
Trenly, line 13: The 	`--accept-eulas` switch makes me a bit nervious

cc @denelon

lars-whta: Otherwise it will not be silent/unattended. EULA is available in interactive mode, as usual.

lars-whta: It is mentioned in the license field:
"NI General Purpose Software License Agreement"
Available here:
https://www.ni.com/en/about-ni/legal/software-license-agreement.html
https://www.ni.com/content/dam/web/pdfs/legal/software_license_agreement_en.pdf

denelon: I don't think we're going to be able to let this package in if it's accepting EULAs without a user consent. The community repository requires a silent or silent with progress (not interactive in either case). Agreements require the publisher to be verified which isn't available for the community repository yet.

lars-whta: I don't want to fight this windmill too much, but a strict enforcement of that policy would remove quite a few of the packages, I believe, e.g.: Google.AndroidStudio enforces its license upon installation/use: https://developer.android.com/studio/terms
But it is just listed in the LicenseUrl‎ field.

Plus some additional ones here:
https://github.com/search?q=repo%3Amicrosoft%2Fwinget-pkgs%20AcceptEULA&type=code

denelon: @lars-whta yes, unfortunately, we're going to need to remove several packages. I believe @Trenly has already started to look for manifests with automatic switches to accept agreements and begin to remove them. The license is something that has been required in the manifest, but the manifest shouldn't be automatically passing arguments to agree to agreements without a user's consent.

denelon: I certainly don't want to see the number of available packages reduced, but we have to maintain our policies. We are working on a solution for verified developers so we can have WinGet present the agreements from the publisher to the users from the community repository. In those cases, the "--accept-package-agreements" would be allowed to provide consent from a user, but we will still need to work on the proper mechanism for that to pass the agreement to the installer, and our policy of requiring an unattended installation to occur would still prevent some of those packages from being added.

lars-whta: Is it possible to provide a mechanism for this, similar to how winget always ask for this:
"The `msstore` source requires that you view the following agreements before using. ... Do you agree to all the source agreements terms?" in your sandbox script.

denelon: The mechanism exists, but the part that doesn't for the community repository is the "verified developer" process to ensure the publisher is the one requiring agreements, and not just a community member submitting the package.

The Microsoft Store establishes a relationship with the publisher which is why those agreements are required for packages coming from the "msstore" source.

denelon: @lars-whta we're going to look at a partial implementation of agreements with the exclusion of the "Agreement Text" so packages with agreements in switches can be added to the community repository. We're in the process of identifying the changes necessary. Huge thanks to @Trenly for doing the right thing here and helping thread the needle on a solution!

I truly appreciate the community support here. I'm working on getting our policies modified to allow Agreement Title, and URL for the cases where manifests need switches to achieve a silent install for WinGet packages.

PR #131091, opened 2023-12-18
Remove: Google.Flutter version 3.16.3

Trenly:
﻿## Package is linked to a bat file which violates the policy of no scripts as installers

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/131091)


mdanish-kh (approved):

denelon (approved):

mwisnicki:
@Trenly out of curiosity - is this policy published somewhere officially?

Trenly:
> @Trenly out of curiosity - is this policy published somewhere officially?

Yes. It's referenced in the [original devblog](https://devblogs.microsoft.com/commandline/winget-install-learning/) and more explicitly stated by @denelon in https://github.com/microsoft/winget-cli/discussions/2940

mwisnicki:
Thanks, I didn't really consider blogs and discussions to be official policy documents.
It would be nice to have a better, more visible space for such things.

Trenly:
> Thanks, I didn't really consider blogs and discussions to be official policy documents. It would be nice to have a better, more visible space for such things.

I mean, it could also fall under the [product policies](https://learn.microsoft.com/en-us/windows/package-manager/package/windows-package-manager-policies#product-policies) section 1.2

mwisnicki:
> > Thanks, I didn't really consider blogs and discussions to be official policy documents. It would be nice to have a better, more visible space for such things.
> 
> I mean, it could also fall under the [product policies](https://learn.microsoft.com/en-us/windows/package-manager/package/windows-package-manager-policies#product-policies) section 1.2

Kinda vague. It only mentions remote scripts but clearly even embedded scripts are banned.

w568w:
IMHO it does not seem a persuasive reason to remove Flutter.

Three articles were cited as sources for the policy:

1. [winget install learning](https://devblogs.microsoft.com/commandline/winget-install-learning/):

> No scripts during install – something that we completely agreed with and don’t allow with MSIX

This is a loose statement: which process is defined as "install"? The Flutter SDK is actually a portable zip archive, with only the process of unpacking and linking it. 

`flutter.bat` is the main executable (or launcher) within the SDK, like what we had in Gradle (`gradlew.bat`) and some old Eclipse IDE variants (`launch-eclipse.bat`). 

`flutter.bat` does download some files on the first launch, but this depends on the command line arguments used to execute it (e.g. `flutter run android`), so it is more like an OTA than an installation.

I'm sorry I haven't found documentation clarifying these definitions. If the definition of "install" is what I think it is, then **this rule does not apply to Flutter**.

2. https://github.com/microsoft/winget-cli/discussions/2940:

> We validate installers (portable packages, and .zip compressed packages) and use SHA256 hashing to identify installers so we can easily validate they haven't been modified before installing them. 

Yes, Flutter SDK is a .zip compressed package too, so you can verify its integrity. Nothing to do with scripts here?

> The manifests are declarative and do not directly contain scripts that would also need to be  scrutinized and handled via other more complex means.

This seems to be talking about using **scripts (written by the maintainers or fetched remotely) in the manifest file**. 

**This also does not seem to apply in the case of Flutter package**: the package manifest don't have scripts attached to it, or fetch some from certain remote locations.

3. [product policies](https://learn.microsoft.com/en-us/windows/package-manager/package/windows-package-manager-policies#product-policies) section 1.2:

> The Product should not, for example, download a remote script and subsequently execute that script in a manner that is not consistent with the described functionality.

Again, as @mwisnicki said, Flutter SDK carries no remote script. **So this rule is quite vague in this situation.**

---

All of them are somehow not directly related to this removal.

Can anyone refer to some other sources clarifying that even embedded scripts are banned? 

It is strange, because with respect to security, open-source `.bat` files are obviously easier to be reviewed compared to closed-source executables; and with respect to maintainability, I do not think a embedded executable is any more superior than a embedded script.

denelon:
The [policies](https://github.com/microsoft/winget-pkgs/blob/master/doc/Policies.md#installer-types) were updated in this repository to reflect our current stance on .bat and .ps1 files. 

I'll bring this topic up again with the security team.

w568w:
> The [policies](https://github.com/microsoft/winget-pkgs/blob/master/doc/Policies.md#installer-types) were updated in this repository to reflect our current stance on .bat and .ps1 files.

Thanks for your information! I have not checked that yet.

I wonder whether it means that even a package with only embedded scripts from upstream is forbidden too? That feels slightly strange to me:

1. Flutter team publishes a zip package with a simple `flutter.bat` launcher: **Not okay**
2. Flutter team publishes a zip package with a `flutter.exe` launcher, compiled from `flutter.bat` by [Bat2Exe](https://github.com/dehoisted/Bat2Exe): **Okay?**
3. Flutter team publishes a zip package with a `flutter.exe` binary, fully closed-source: **Okay?**

> I'll bring this topic up again with the security team.

That would be great! Looking forward to further discussions and decisions from the team.

denelon:
The policies here are specific to the "community" source. If you have a private REST source, the restrictions will not apply the same way. We've had plenty of the same debates internally. I'm certainly no security expert though, so I don't fully understand all the subtle nuances. For example, I would have personally though sudo wouldn't be approved in Windows, and it recently was (yay!). I think it just takes time for some of these concerns to be considered both in isolation as well as how they apply in an ecosystem.

mwisnicki:
So I found this discussion https://github.com/microsoft/winget-cli/discussions/2940 about banning scripts in installers with somewhat reasonable arguments but it looks like it has since been misapplied to also ban scripts inside applications which does not make sense. Hopefully security team will clear this up as it seems like some miscommunication.

PR #131095, opened 2023-12-18
Removal: Google.Chrome accepts EULA on user's behalf

lars-nvhgroup:
"By downloading Chrome, you agree to the [Google Terms of Service](https://policies.google.com/terms) and [Chrome and ChromeOS Additional Terms of Service](https://www.google.com/chrome/terms/)"

Accepts the EULA terms on user's behalf, silently.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/131095)

### DIFF
--- a/manifests/s/Segger/EmbeddedStudioARM/6.32b/Segger.EmbeddedStudioARM.installer.yaml
+++ b/manifests/s/Segger/EmbeddedStudioARM/6.32b/Segger.EmbeddedStudioARM.installer.yaml
@@ -1,0 +1,31 @@
+# Created with YamlCreate.ps1 v2.1.2 $debug=QUSU.5-1-19041-1682
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+
+PackageIdentifier: Segger.EmbeddedStudioARM
+PackageVersion: 6.32b
+InstallerLocale: en-US
+MinimumOSVersion: 10.0.0.0
+InstallerType: exe
+Scope: machine
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+InstallerSwitches:
+  Silent: --silent
+  SilentWithProgress: --silent
+  Custom: --accept-license
+UpgradeBehavior: install
+FileExtensions:
+- emarchive
+- empackage
+- emproject
+Installers:
+- Architecture: x64
+  InstallerUrl: https://dl2.segger.com/files/embedded-studio/Setup_EmbeddedStudio_ARM_v632b_win_x64.exe
+  InstallerSha256: 04E159DA540526E3F3B84B9A92F8E252F03428F6D989AF70366A8F87650E90F6
+- Architecture: x86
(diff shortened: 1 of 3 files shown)


denelon:
This is not the same as a manifest passing explicit agreements. Users aren't required to accept agreements to download via https://www.google.com/chrome/index.html.

denelon:
The license URL does not have a restriction for download based on: https://www.google.com/intl/en_us/chrome/terms/

lars-whta:
> "By downloading Chrome, you agree to the [Google Terms of Service](https://policies.google.com/terms) and [Chrome and ChromeOS Additional Terms of Service](https://www.google.com/chrome/terms/)"



> This is not the same as a manifest passing explicit agreements. Users aren't required to accept agreements to download via https://www.google.com/chrome/index.html.

It contradicts what is stated here:
https://www.google.com/chrome/

"By downloading Chrome, you agree to the [Google Terms of Service](https://policies.google.com/terms) and [Chrome and ChromeOS Additional Terms of Service](https://www.google.com/chrome/terms/)"

lars-whta:
> This is not the same as a manifest passing explicit agreements. Users aren't required to accept agreements to download via https://www.google.com/chrome/index.html.

How is it not the same? By downloading Chrome the EULA is implicitly accepted, per the language on that page.

lars-whta:
> This is not the same as a manifest passing explicit agreements. Users aren't required to accept agreements to download via https://www.google.com/chrome/index.html.

They definitely are. How can you as a user avoiding the EULA by downloading, if it is accepted by mere downloading?

image

Trenly:
The difference is in how the installer handles the terms and licenses.

If you download the google chrome installer and use the command line switch `/SILENT`, the application will install and run perfectly fine. There is no blocking dialogue which requires accepting the terms of a EULA.  Because the publisher allows for the agreement to be implicit through the use of the silent switch, having the license URL is no different than what Google does by publishing the license URL on their downloads page. Users aren't required to read the EULA to download and install Chrome, so they aren't required to read the EULA to download and install it using WinGet.

With installers that require explicit agreements, users must read the EULA or agree to it before the install can continue. These agreements are the ones that cannot be suppressed with just a `/SILENT` switch. Even though there may be a `--accept-licenses` switch, that is still a form of the user explicitly saying `I have read and agree to the license terms and conditions`. Because the publisher of the installer has required that the user explicitly agree, having the license URL is *not* enough to satisfy the requirement.  Users must be required to either check a box or click accept on the agreement. So, if WinGet doesn't show the agreement and get explicit consent then WinGet cannot pass any form of `--accept-licenses` to the installer because the user, *legally speaking*, has not agreed to the EULA in the manner prescribed by the publisher of the installer.

denelon:
We are actively working on the "verified publisher" feature so we can have the same mechanisms provided by WinGet in the "msstore" source available to the community repository.

PR #131120, opened 2023-12-19
New version: MusicBrainz.Picard 2.10.0

phw:

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/131120)

### DIFF
--- a/manifests/m/MusicBrainz/Picard/2.10.0/MusicBrainz.Picard.installer.yaml
+++ b/manifests/m/MusicBrainz/Picard/2.10.0/MusicBrainz.Picard.installer.yaml
@@ -1,0 +1,19 @@
+# Created using wingetcreate 1.5.7.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.5.0.schema.json
+
+PackageIdentifier: MusicBrainz.Picard
+PackageVersion: 2.10.0
+Platform:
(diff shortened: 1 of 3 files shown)


phw:

stephengillie:
Hi @phw,

> Current manifest version 2.9.1000.0 has 4 parameters, which is greater than this PR's version number 2.10.0's 3 parameters.

Is this intentional?

(Automated response - build 486)

phw:
@stephengillie  We usually use only major.minor.patch. The previous version number seems to have been generated from the MS store distribution of Picard (which was never intended to be used outside of the MS store distribution and is not supported for any other use case), which also includes the Microsoft typical build number as a fourth element.

Also more specifically the MS store forbids submitted applications to use the build number as it is reserved for internal use, hence store builds use 2.9.1000 instead of 2.9.1 so the patch number provides some digits to bump build number if necessary. The actual version number of the application is 2.9.1, though.

If possible I'd like to stick with the official versioning, which would be 2.10.0 for the current release and 2.10.1 for the upcoming patch release. If winget will expect the four element version we can also go with 2.10.0.0 and 2.10.1.0 respectively.

stephengillie (approved):

PR #131477, opened 2023-12-21
New package: Fort.Firewall version 3.10.4

ghost:

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/131477)

### DIFF
--- a/manifests/f/Fort/Firewall/3.10.4/Fort.Firewall.installer.yaml
+++ b/manifests/f/Fort/Firewall/3.10.4/Fort.Firewall.installer.yaml
@@ -11,12 +11,6 @@ Installers:
   Dependencies:
     PackageDependencies:
     - PackageIdentifier: Microsoft.VCRedist.2015+.x64
-- Architecture: x86
-  InstallerUrl: https://github.com/tnodir/fort/releases/download/v3.10.4/FortFirewall-3.10.4-windows-x86.exe
-  InstallerSha256: E0B7A77C1EA1F1D69C6805957C927E0F9C499D676CEF78DF9E8BD6722F9361B9
-  Dependencies:
-    PackageDependencies:
-    - PackageIdentifier: Microsoft.VCRedist.2015+.x86
 - Architecture: arm64
   InstallerUrl: https://github.com/tnodir/fort/releases/download/v3.10.4/FortFirewall-3.10.4-windows10-arm64.exe
   InstallerSha256: 8572DE16827F954E907D08CAE58673806898530C25B557C48D3FC5BF924ADBF5
--- a/manifests/f/Fort/Firewall/3.10.4/Fort.Firewall.installer.yaml
+++ b/manifests/f/Fort/Firewall/3.10.4/Fort.Firewall.installer.yaml
@@ -1,0 +1,21 @@
+# Created using wingetcreate 1.5.7.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.5.0.schema.json
+
+PackageIdentifier: Fort.Firewall
+PackageVersion: 3.10.4
+InstallerType: inno
+Installers:
+- Architecture: x64
+  InstallerUrl: https://github.com/tnodir/fort/releases/download/v3.10.4/FortFirewall-3.10.4-windows10-x86_64.exe
+  InstallerSha256: B90413715AA804ABD56D4C20599798806E7276DE84F3B000AE66934EED023CB7
+- Architecture: x86
+  InstallerUrl: https://github.com/tnodir/fort/releases/download/v3.10.4/FortFirewall-3.10.4-windows-x86.exe
+  InstallerSha256: E0B7A77C1EA1F1D69C6805957C927E0F9C499D676CEF78DF9E8BD6722F9361B9
+- Architecture: arm64
+  InstallerUrl: https://github.com/tnodir/fort/releases/download/v3.10.4/FortFirewall-3.10.4-windows10-arm64.exe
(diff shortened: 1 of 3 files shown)


Emi-Emi-Emi:
@denelon 

I did what you said and I still get the `Validation-Unattended-Failed` message/tag on this.

Tested it on my side and it worked fine. I even added the `Microsoft.VCRedist.2015+.x86` dependency to the manifest as you made me learn how it could be added in the manifest.

I also used Winget Creator this time, which detected the installation as Inno correctly unlike Komac, and I even tested on another computer installing it uninstalling the Microsoft.VCRedist.2015+.x86 to make sure everything was okay, and it worked fine in the live environment as well.

The only thing that might be affecting this is the Memory Integrity / Core isolation requirement, there is no other explanation. So how can Fort deal with that?

This is what the installer checks and the message it throws when Memory Integrity is still on.

```
function HVCIEnabled(): Boolean;
var
  EnabledValue: Cardinal;
begin
  Result := RegQueryDWordValue(HKEY_LOCAL_MACHINE,
      'SYSTEM\CurrentControlSet\Control\DeviceGuard\Scenarios\HypervisorEnforcedCodeIntegrity',
      'Enabled', EnabledValue)
    and (EnabledValue = 1);
end;

function InitializeSetup: Boolean;
begin
  if HVCIEnabled() then
  begin
    MsgBox('This program is not compatible with HVCI (Core Isolation).', mbCriticalError, MB_OK);

    Result := False;
    Exit;
  end;

  Result := True;
end;
```

That was added to avoid bluescreens when driver loads on installation, so it was better to throw the message than letting people install something that would cause issues, of course this message might be incompatible with Azure pipeline, and therefore throwing the message that _**failed to install without user input.**_

Should I give up until driver is properly signed (if) someday or is there a way to do this without having the same `Validation-Unattended-Failed` on every update.

Emi-Emi-Emi:
I also read winget doesn't like scripts in the installers. I forgot how fort has scripts for the driver,to make it easy to uninstall and reinstall the driver.

I don't know why it labels as Unattended in that case, but could be another issue?

https://github.com/tnodir/fort/blob/master/deploy/FortFirewall.iss

That's the installer. So it can be quickly checked for anything winget won't like to get the answer if I should close this or there is a way for Fort to pass validation.

denelon:
It looks like the dependency might not be correct. The screenshot from validation shows the error.

fort

Emi-Emi-Emi:
Ok I know the issue. Let's ignore everything I said, it was my fault but on my defense, it's Microsoft.VCRedist fault, because when you uninstall it, it **doesn't** remove the files from the system32, I don't know why, but even when I removed every redist, that dll still exists in win32.

I tried to use Sandbox but it was bluescreening that old computer, and I don't have any more computers to 'clean' test it (my fault), so I started to do the live test in that computer without supposedly having vcredist installed, but the dll file is there, so the installer wasn't throwing the error.

I mean, there is no excuse to uninstall something and don't remove all the files it copied to the system, especially when I can manually go to system32 and remove the files so they are not even in use or anything, but I will fix this, because it is pretty much obvious I have to add a dependency for every different build, especially ARM. 

Thank you for the feedback, this will be the last time you hear from me! 👍

denelon:
@Emi-Emi-Emi no worries at all 😊

Feel free to ask questions, we all have to start somewhere!

If you haven't played with Hyper-v, that might be a reasonable way to build a "clean" Windows VM to test with. It's a huge download initially, but once you have a good image to start with, it's reasonably easy to create a VM and have a "clean" Windows base image.

denelon:
You can click through the build at GitHub to get to the validation build in Azure Dev Ops, and for each run, there are artifacts you can open up to see all the logs and screenshots if any were taken. That's where I got the screenshot above.

Emi-Emi-Emi:
I tried to get the screenshots you told me, clicked everywhere but then the only thing I didn't do was to 'sign in', when it asked me.
But I found another old computer laying around to get it to test in Sandbox, it didn't bluescreen so so I tested it and it worked fine, but I can see there is an issue.

The installer is checking for VCRedist x86, which is the reason why I thought was the dependency, so even if you have the x64 correctly installed, it still opens Edge and the website to download and install the latest Microsoft.VCRedist.2015+.

So I will have to talk to dev to fix that. Because might be another issue. I can't test x86 or ARM though. but x64 version has this issue of Edge. But dependencies should work fine now, at least in x64.

Emi-Emi-Emi:
If there an issue with the tags or something? The Validation Pipeline was under maintenance, so it didn't work for couple days, but now it is 'healthy' and other packages are being _checked_, but not this one which only has the `Retry-1` tag on it.

Does it has to be forced/manually to run for it to work again?

Emi-Emi-Emi:
The screenshot says it was the x86 version having issues with the VCRedist issue, I added the dependency to the manifest so I don't know why the .dll error or what happened.

I decided to remove x86 installer, because Windows 11 is 64bits only, and the probability of someone installing Fort in Win10 32bits is minimal, most people using 32 bits are Win7 users which doesn't have winget support.

I still don't get why so many issues with this, but the less installers the better, especially when I can't test x86 without a virtual machine or ARM64.... I am keeping ARM64, but If it fails I will remove it too, and keep it simple.

But since new version 3.10.5 is almost out, I might close this and try again in some days when the stable final version is out.

Emi-Emi-Emi:
Well, apparently there was also an error with dependency in x64, but it installed with no issues in Sandbox, so I better close this and try when next version drops.

PR #131630, opened 2023-12-22
Update: EverEdit.EverEdit version 4.5.0.4500

SpecterShell:
- [X] Does your manifest conform to the [1.5 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.5.0)?

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/131630)

### DIFF
--- a/manifests/e/EverEdit/EverEdit/4.5.0.4500/EverEdit.EverEdit.installer.yaml
+++ b/manifests/e/EverEdit/EverEdit/4.5.0.4500/EverEdit.EverEdit.installer.yaml
@@ -1,5 +1,5 @@
-# Created with YamlCreate.ps1 v2.2.4 $debug=QUSU.CRLF.7-3-4.Win32NT
-# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.4.0.schema.json
+# Created with YamlCreate.ps1 v2.2.12 $debug=QUSU.CRLF.7-4-0.Win32NT
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.5.0.schema.json
 
 PackageIdentifier: EverEdit.EverEdit
 PackageVersion: 4.5.0.4500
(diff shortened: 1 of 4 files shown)


SpecterShell:
The URLs come from the homepage as is.

stephengillie:
Hi @SpecterShell,

From the PackageUrl, I found the Download link which has links to:

- 32-bit http://www.everedit.net/latest.php?cpu=x86
- 64-bit http://www.everedit.net/latest.php?cpu=x64

Should these be used instead? 

I couldn't find the InstallerUrl - could you link me to the page it's on?

SpecterShell:
> Hi @SpecterShell,
> 
> From the PackageUrl, I found the Download link which has links to:
> 
>     * 32-bit http://www.everedit.net/latest.php?cpu=x86
> 
>     * 64-bit http://www.everedit.net/latest.php?cpu=x64
> 
> 
> Should these be used instead?
> 
> I couldn't find the InstallerUrl - could you link me to the page it's on?

Better not. The manifests will get expired once the publisher updates the installer even if the old version is still available on the server.

And the InstallerUrl is what these two URLs redirect to.

image

stephengillie:
I didn't try the URLs before. Trying now, they don't appear to work in Edge:
image

Does this package have other URLs?

SpecterShell:
> I didn't try the URLs before. Trying now, they don't appear to work in Edge: image
> 
> Does this package have other URLs?

It works in Edge. Just click "Keep" in the three dots menu.

manifests/e/EverEdit/EverEdit/4.5.0.4500/EverEdit.EverEdit.installer.yaml
```diff
@@ -95,16 +95,16 @@ FileExtensions:
 - xslt
 - yaml
 - yml
-ReleaseDate: 2023-04-02
+ReleaseDate: 2023-10-16
 ElevationRequirement: elevationRequired
 Installers:
 - Architecture: x86
-  InstallerUrl: http://116.198.40.241/everedit_win32_4500_setup.exe
+  InstallerUrl: http://8.140.57.126/everedit_win32_4500_setup.exe
```
stephengillie, line 102:
```suggestion
  InstallerUrl: http://www.everedit.net/latest.php?cpu=x86
```

manifests/e/EverEdit/EverEdit/4.5.0.4500/EverEdit.EverEdit.installer.yaml
```diff
@@ -95,16 +95,16 @@ FileExtensions:
...
   InstallerSha256: 4BE41D193387FA1A32BB9268D0359C4108EFE39371D670656174AAF61AB7D5D2
   ProductCode: EverEdit
 - Architecture: x64
-  InstallerUrl: http://116.198.40.241/everedit_win64_4500_setup.exe
+  InstallerUrl: http://8.140.57.126/everedit_win64_4500_setup.exe
```
stephengillie, line 106:
```suggestion
  InstallerUrl: http://www.everedit.net/latest.php?cpu=x64
```

stephengillie:
I'm still having difficulty navigating from the PackageUrl to the InstallerUrl. 

Falling back to IP addresses is both a security gap and an automation failure. It's much more difficult to get numerous SSL certs for every server and address, than just get 1 for the whole domain. This is why these can't be downloaded securely - because the host didn't bother to get an SSL cert foreach endpoint. 

> Better not. The manifests will get expired once the publisher updates the installer even if the old version is still available on the server.

This is a normal vanity URL, and this problem is the same for a large quantity of our packages. So we should use the same solution of depending on automation and community members to update these when they become out of date. 

Alternatively, if the developer offers a versioned link, feel free to use this here. 

Dropping DNS would exacerbate the problem instead of ameliorate, as mentioned above. It prevents the automation of repointing domains to different addresses during normal and emergency situations.

SpecterShell:
Well the publisher updated the URLs to a named domain today.

stephengillie (approved):

PR #131760, opened 2023-12-24
Version Update: TeamRadarr.Radarr version 5.2.6

babblebauble:

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/131760)

### DIFF
--- a/manifests/t/TeamRadarr/Radarr/5.2.6/TeamRadarr.Radarr.installer.yaml
+++ b/manifests/t/TeamRadarr/Radarr/5.2.6/TeamRadarr.Radarr.installer.yaml
@@ -1,0 +1,16 @@
+# Created using wingetcreate 1.5.7.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.5.0.schema.json
+
+PackageIdentifier: TeamRadarr.Radarr
+PackageVersion: 5.2.6
+InstallerType: inno
+Installers:
+- Architecture: x86
+  InstallerUrl: https://github.com/Radarr/Radarr/releases/download/v5.2.6.8376/Radarr.master.5.2.6.8376.windows-core-x86-installer.exe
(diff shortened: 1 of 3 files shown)


stephengillie:
Hi @babblebauble,

This manifest removes Apps and Features entries that are present in previous PR versions. Should these entries be added to this version also?

(Automated response - build 489.)

babblebauble:
> Hi @babblebauble,
> 
> This manifest removes Apps and Features entries that are present in previous PR versions. Should these entries be added to this version also?
> 
> (Automated response - build 489.)

Hi. I tested installation without the Apps and Features entries in the manifest and the installer writes the correct ARP entries to the registry so having the following entries in the manifest is unnecessary. 

`AppsAndFeaturesEntries:`
`- DisplayName: Radarr version x.x.x`

Capture

stephengillie:
Hi @Trenly and @mdanish-kh,

Thoughts on Apps and Features entries for this package?

Trenly:
Since it is *only* the display name that is in the manifest, I *think* it's okay to be left off. The biggest concern is if PackageVersion is missing.

I'd say let's merge this, I can do some testing after it passes publishing and add the entry if needed

mdanish-kh:
It is generally a good idea to add as much metadata as you can, and Apps&Features helps with package correlation. In this case however, winget can already get a good match using the `PackageName` and `PackageVersion`. Having `AppsAndFeatures > DisplayName` for this package means that it would need to be manually updated for every new version (since current manifest creators don't have the ability to automate updating these fields).

Since it's only the `DisplayName` field in question here, and there is no need for explicit matching, I would gravitate towards excluding it

Trenly (approved):

Trenly:
Ah, probably won't merge with the Needs-Attention label

[Policy] Reset Feedback

Trenly (approved):

Issue #32785, opened 2021-10-27
[Package Request]: 


Hellements:
### Package Requested

- [X] I would like someone else to build the manifest.
- [ ] I would like help so I can submit the manifest.
- [X] I have performed a search and couldn't find this package.
- [ ] I have checked that the installer for this package is not a standalone or portable executable or a .zip file
- [ ] I think there is a new version available and I have provided the URL.

### Please provide the following information

Publisher: Adobe Systems Incorporated
Publisher Url: https://www.adobe.com/acrobat/pdf-reader.html
Package Name: Adobe.AdobeAcrobatReaderDCx64
Description: The free global standard for reliably viewing, printing, signing, and commenting on PDF documents.
Package Version: 2021.007.20099
Installer URL: https://get.adobe.com/reader/otherversions/

Current Package Identifier (for new versions):

Hellements:
> 
> 
> > > #30135 https://github.com/microsoft/winget-pkgs/tree/master/manifests/a/Adobe/AdobeAcrobatReaderDC
> > > @denelon
> > 
> > 
> > Hello there.
> > That link is for 32bits version, not for x64 as I requested.
> > Thanks.
> 
> You can simply add this link and architecture: x64 to the current manifest. Edit: Cannot find direct link to x64 offline installer, could anyone help me?

Hello,
Unfortunately I don't have a direct link for x64 offline installer. What I do is use a PowerShell script from @farag2 that downloads and install the latest version available. Please check it, maybe you guys discover how to get the link.
https://github.com/farag2/Utilities/blob/master/Download/Acrobat_Reader_DC_x64.ps1
Thanks.

Hellements:
> 
> 
> > > #30135 https://github.com/microsoft/winget-pkgs/tree/master/manifests/a/Adobe/AdobeAcrobatReaderDC
> > > @denelon
> > 
> > 
> > Hello there.
> > That link is for 32bits version, not for x64 as I requested.
> > Thanks.
> 
> You can simply add this link and architecture: x64 to the current manifest. Edit: Cannot find direct link to x64 offline installer, could anyone help me?

Hellements:
> 
> 
> > > #30135 https://github.com/microsoft/winget-pkgs/tree/master/manifests/a/Adobe/AdobeAcrobatReaderDC
> > > @denelon
> > 
> > 
> > Hello there.
> > That link is for 32bits version, not for x64 as I requested.
> > Thanks.
> 
[...]

PR #132904, opened 2024-01-05
Merge x64 and x86 Adobe Acrobat Reader installers

rakleed:

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/132904)

### DIFF
--- a/manifests/a/Adobe/Acrobat/Reader/23.008.20421/Adobe.Acrobat.Reader.installer.yaml
+++ b/manifests/a/Adobe/Acrobat/Reader/23.008.20421/Adobe.Acrobat.Reader.installer.yaml
@@ -1,7 +1,7 @@
 # Created using wingetcreate 1.5.7.0
 # yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.5.0.schema.json
 
-PackageIdentifier: Adobe.Acrobat.Reader.32-bit
+PackageIdentifier: Adobe.Acrobat.Reader
 PackageVersion: 23.008.20421
 MinimumOSVersion: 10.0.0.0
 InstallerType: exe
@@ -19,15 +19,12 @@ FileExtensions:
 - pdfa
 - pdfx
 - xfx
-ProductCode: '{AC76BA86-7AD7-FFFF-7B44-AC0F074E4100}'
-AppsAndFeaturesEntries:
-- DisplayName: Adobe Acrobat Reader DC MUI
-  Publisher: Adobe Systems Incorporated
-  ProductCode: '{AC76BA86-7AD7-FFFF-7B44-AC0F074E4100}'
-  InstallerType: msi
 Installers:
 - Architecture: x86
   InstallerUrl: https://ardownload2.adobe.com/pub/adobe/reader/win/AcrobatDC/2300820421/AcroRdrDC2300820421_MUI.exe
   InstallerSha256: FAB45F55F9EE6D96F123654D28C1A0EF88140BE3D2D51B2CBFB79A446967D99D
+- Architecture: x64
+  InstallerUrl: https://ardownload2.adobe.com/pub/adobe/acrobat/win/AcrobatDC/2300820421/AcroRdrDCx642300820421_MUI.exe
+  InstallerSha256: 5C79E5ABF6CD39A94893F4FDD73BAA9B8CBD7B501DE90CF45E04A05BB6669332
 ManifestType: installer
 ManifestVersion: 1.5.0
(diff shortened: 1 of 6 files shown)


rakleed:
If that's ok, let me know, and I will create PRs for older version and separate x64 packages.

rakleed:
If that's ok, let me know, and I will create PRs for older version and separate x64 packages.

mdanish-kh (changes requested):
I am not a user of this package, but I know that this package was broken into different architectures because the two architectures install side-by-side. I believe a user reported that the two applications work very differently and that there could be value in having winget handle updates separately for the two variants. 

Also, changing the PackageId would be a breaking change for scripts/applications in the wild that are using winget to install this package. If we want to do this, there should be a convincing enough reason for it. I'm curious to know why would you like to merge the two IDs into one?

rakleed:
@mda since https://github.com/microsoft/winget-pkgs/issues/32785#issuecomment-956495573 and https://helpx.adobe.com/acrobat/kb/about-acrobat-reader-dc-migration-to-64-bit.html.

But in https://github.com/microsoft/winget-pkgs/pull/40297#issuecomment-1008152647 @OfficialEsco mentioned that two separate PackageIDs are needed. @OfficialEsco, can you tell if there is a reason to leave two separate PackageIDs or can they be combined by simply removing AppsAndFeaturesEntries from the manifest?

OfficialEsco:
Ohboi there have been so many changes with the Adobe Acrobat Reader package...
Technically; If a Package can be installed Side-By-Side they should be separate. (Atleast when i was actively validating packages (2 years ago))
IIRC there was a difference (DisplayName?) between the x64 and x86 AND the MUI installation. However yes this can be fixed with the ARP field now days.

Just to note; You're not allowed to move a package, it has to be delete/create PRs

rakleed:
@mdanish-kh @OfficialEsco just checked, they cannot be installed at the same time, only one version. If you install the x86 version first and then try to install it on top of x64, it will automatically update to x64 in silent mode, and in interactive mode it will prompt you to do this:
image
If canceled, only the x86 version will remain.

If the x64 version was installed, then x86 simply will not be installed:
image

I googled and most of the answers were that there is no difference between applications with different architectures.

But yes, x86 and x64 have different DisplayName, Publisher, ProductCode. I was able to add ProductCode to Installers, but can I do the same for DisplayName and Publisher?

mdanish-kh:
If that is indeed the case, merging the PackageIds seems to be a viable option. (_Perhaps the installers behaved differently in the past and have now got updated_?)

I'm out for a couple of days... so I can't test this right now in my VM. I would also like to add some other folks into the loop

- @PckgrBot who reported the two arches behave differently on element
- @Trenly (originally broke the package into different IDs)

Trenly:
There is also the aspect of currently installed architecture. Because of what @PckgrBot mentioned, it would be best to leave users on the architecture they currently have installed. I'd be worried that if we merge the packages, then there would be more issues with matching the currently installed architecture and ensuring users receive the correct upgrade package. There might be ways to get around this with AppsAndFeaturesEntries, but given the history here with Adobe, it may be best to keep them separate

rakleed:
@PckgrBot @Trenly as far as I understand, two manifestos were originally made, since [Manifest 1.0.0 did not support multiple architectures at the same time](https://github.com/microsoft/winget-pkgs/issues/32785#issuecomment-956495573). Then they were apparently separated again, since [it was possible to install x86 and x64 versions at the same time](https://github.com/microsoft/winget-pkgs/pull/132904#issuecomment-1880039371).

But for a long time the manifest has supported several architectures simultaneously, and [Adobe since February 2021 (version 21.xxx.xxxxx and later) is gradually migrating to x64 versions](https://helpx.adobe.com/acrobat/kb/about-acrobat-reader-dc-migration-to-64-bit.html), and does not allow applications from x86 and x64 to be installed simultaneously (I checked this with the earliest available 21.xxx.xxxxx in winget x86 and x64 versions). Apparently there were problems with the 2020 version (20.xxx.xxxxx), since even [on the Adobe website they are highlighted as separate products](https://www.adobe.com/products/catalog.html#category=pdf&types=desktop).

As for the fact that x64 Acrobat Reader is both Acrobat Standard/Pro, I don’t see a problem here, since most users use the free Reader version anyway.

In addition to `ProductCode`, I also added `DisplayName` and `Publisher` to the manifest, validation passed. `InstallerType: msi` was also previously specified, apparently because `ProductCode` only works for them. If necessary, I can also add `InstallerType`.

rakleed:
@stephengillie

stephengillie:
> @ stephengillie

Hi @rakleed,

There's more discussion than information in this PR comment thread. The question is: "Should these installers be merged into a single manifest?" 

- Can these be installed side-by-side? If so they should be separate, if no they should not.
- x86 seems to be losing support going forward, so will need to be removed at some point.

It sounds like - instead of merging them, we should depreciate the `x86` version. 

This situation seems similar to when `Git.Git` depreciated their `x86` version, starting with putting a message in their `x86` package telling users to upgrade - it couldn't be skipped or made silent by installer switch. There, we had no option but to drop the `x86` version from the manifest. Here, Adobe have been working on this for 3 years. 

Do we have many users of this software who are currently using a 32-bit OS and CPU? Or are most users running the 32-bit version on a 64-bit OS?

PR #132906, opened 2024-01-05
Intel.WiFiDriversx86 version 22.160.0

Eddga:

---

I wasn't able to test the installation as I don't have an x86 system and this package is x86 only.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/132906)

### DIFF
--- a/manifests/i/Intel/WiFiDriversx86/22.160.0/Intel.WiFiDriversx86.installer.yaml
+++ b/manifests/i/Intel/WiFiDriversx86/22.160.0/Intel.WiFiDriversx86.installer.yaml
@@ -1,0 +1,20 @@
+# Created using wingetcreate 1.5.7.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.5.0.schema.json
+
+PackageIdentifier: Intel.WiFiDriversx86
+PackageVersion: 22.160.0
+InstallerLocale: en-US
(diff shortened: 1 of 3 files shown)


Eddga:
Before approving: Please I need some guidance here on the naming scheme.
This is the latest and last 32-bit release for the Intel Wi-Fi driver. 64-bit is actively developed and I've submitted a separate PR for that with a separate package name: #132908
But I reckon adding it as a version in the same tree as the 64-bit driver will end up with this version not being found by users/clients. Or is winget as clever as checking all versions and searching for the latest 32-bit release if run on a 32-bit system?
In any case I'm not sure if "Intel.WiFiDriversx86" is a proper package name. But I couldn't do better. 😅
And as this will only ever be a legacy package and x64 is the default I thought that it would be a good idea to drop the architecture in the package name for x64.

stephengillie:
Hi @Eddga,

These versions should be in the same manifest, so users can use the ` -a` or `--architecture` WinGet switch to select between them. https://github.com/microsoft/winget-pkgs/pull/132908#discussion_r1443419319

Eddga:
Thanks for your feedback @stephengillie but I think I'll just drop the 32-bit version completely for now. Whenever somebody really needed it he/she could create a request and we will see if somebody could figure out a better way to resolve this.

PR #132936, opened 2024-01-05
Microsoft.VSTOR version 10.0.60912.00

Eddga:

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/132936)

### DIFF
--- a/manifests/m/Microsoft/VSTOR/10.0.60912.00/Microsoft.VSTOR.locale.en-US.yaml
+++ b/manifests/m/Microsoft/VSTOR/10.0.60912.00/Microsoft.VSTOR.locale.en-US.yaml
@@ -5,7 +5,7 @@ PackageIdentifier: Microsoft.VSTOR
 PackageVersion: 10.0.60912.00
 PackageLocale: en-US
 Publisher: Microsoft Corporation
-PackageName: Microsoft Visual Studio Tools for Office Runtime 2010 Redistributable
+PackageName: Microsoft Visual Studio 2010 Tools for Office Runtime
 License: Proprietary
 Copyright: © Microsoft Corporation. All rights reserved.
 ShortDescription: This download installs the Visual Studio 2010 Tools for Office Runtime, which is required to run Microsoft Office based solutions built using Microsoft Visual Studio 2010, 2012, 2013, 2015, 2017 and 2019.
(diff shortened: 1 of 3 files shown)


manifests/m/Microsoft/VSTOR/10.0.60912.00/Microsoft.VSTOR.locale.en-US.yaml
```diff
@@ -0,0 +1,13 @@
+# Created using wingetcreate 1.5.7.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.5.0.schema.json
+
+PackageIdentifier: Microsoft.VSTOR
+PackageVersion: 10.0.60912.00
+PackageLocale: en-US
+Publisher: Microsoft Corporation
+PackageName: Microsoft Visual Studio Tools for Office Runtime 2010 Redistributable
```
OfficialEsco, line 8:
```suggestion
PackageName: Microsoft Visual Studio 2010 Tools for Office Runtime
```

OfficialEsco (commented):
I'm seeing this DisplayName?

manifests/m/Microsoft/VSTOR/10.0.60912.00/Microsoft.VSTOR.locale.en-US.yaml
Eddga: @OfficialEsco thanks. You're right. Does this package name refer to the app name as seen in the Windows programs list? I haven't checked that and also haven't worked with that attribute before. Don't even know when it would be of use/need.
But when I install it on my x64 system I see it as 'Microsoft Visual Studio 2010 Tools for Office Runtime (x64)' - does the architecture mention at the end matter?

OfficialEsco: WinGet matches against what you see in "Add or remove programs"/"Programs and Features" which is for example located in `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall` (64bit machine wide) registry

Publisher, DisplayName and Version is the most important.
There is a more advanced field called "AppsAndFeaturesEntries" where you can add this if the `PackageName` cannot match.
WinGet strips the (x64) and the version away from the matching, also it uses wildcard to detect anything after whats defined in the `PackageName`

Eddga: I see. Thanks for the explanation. And is this matching used in winget update process or to determine successfully completed installation?

OfficialEsco: It is used for "winget list" which then uses "winget upgrade" yes.
The installation just depends on the exit code, so something completely could be installed with success.

OfficialEsco (approved):

PR #133024, opened 2024-01-07
New package ViennaSymphonicLibrary.ViennaAssistant version 1.1.340

marco-eckstein:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/133024)

### DIFF
--- a/manifests/v/ViennaSymphonicLibrary/ViennaAssistant/1.1.340/ViennaSymphonicLibrary.ViennaAssistant.installer.yaml
+++ b/manifests/v/ViennaSymphonicLibrary/ViennaAssistant/1.1.340/ViennaSymphonicLibrary.ViennaAssistant.installer.yaml
@@ -1,0 +1,16 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.5.0.schema.json
+
+# On x64, this package will silently also install Microsoft.VCRedist.2015+.x64.
+# On x86, it might install Microsoft.VCRedist.2015+.x86 (unconfirmed).
+# Since the installer is neutral, we cannot depend on either package.
+
+PackageIdentifier: ViennaSymphonicLibrary.ViennaAssistant
+PackageVersion: 1.1.340
+InstallerType: inno
+ProductCode: Vienna Assistant_is1
+Installers:
+  - InstallerUrl: https://eu.vsl.co.at/?FileID=353266
+    Architecture: neutral
+    InstallerSha256: 81B63CDAB6E42EFF70BD0ED4A2C33659E4A2D779D42A94CD5353F356118DA050
+ManifestType: installer
+ManifestVersion: 1.5.0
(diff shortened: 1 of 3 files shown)


marco-eckstein:
> One or more of the installer URLs doesn't appear valid.
> 
> This may happen for sites with policies prohibiting distribution or use by third parties.
> 
> This may happen for URLs pointing to domains that do not align with the publisher domain or package domain. If you could provide supporting evidence from the publisher that the URLs for the installer are correct, that would help us to validate and approve this PR.

The installer URL is `https://eu.vsl.co.at/?FileID=353266`. The publisher URL is `https://www.vsl.co.at`. So it is quite clear that the installer URL belongs to the publisher. If it helps, you could also change the publicher URL to `https://vsl.co.at`, so that the installer URL's domain is a proper subdomain.

stephengillie:
> > One or more of the installer URLs doesn't appear valid.
> > This may happen for sites with policies prohibiting distribution or use by third parties.
> > This may happen for URLs pointing to domains that do not align with the publisher domain or package domain. If you could provide supporting evidence from the publisher that the URLs for the installer are correct, that would help us to validate and approve this PR.
> 
> The installer URL is `https://eu.vsl.co.at/?FileID=353266`. The publisher URL is `https://www.vsl.co.at`. So it is quite clear that the installer URL belongs to the publisher. If it helps, you could also change the publicher URL to `https://vsl.co.at`, so that the installer URL's domain is a proper subdomain.

Hi @marco-eckstein,

I just have to be able to navigate from the PackageUrl to the InstallerUrl.

marco-eckstein:
> I just have to be able to navigate from the PackageUrl to the InstallerUrl.

The package URL https://www.vsl.info/en/manuals/vienna-assistant/introduction links to https://www.vsl.co.at/service/vawin/, which allows the download then, but the download URL that Firefox shows for the download is https://eu.vsl.co.at/?FileID=353266. I don't know whether https://www.vsl.co.at/service/vawin/ could be used as `InstallerUrl`.

marco-eckstein:
> This package didn't run when launched on a Windows 10 VM. I see the Installation Notes of: `In case you start the application and see neither GUI nor an error message, restart the computer.` But I'm not sure if this violates our "doesn't require reboot" policy.

What exactly is that policy? I cannot find anything via Google. The message was added by me. It is apparently a bug in the package - one can start the application directly after installing, and it will show up task manager, but the GUI remains invisible. Only after a restart does it function properly.

stephengillie:
We have a policy that packages can't force or cause the system to reboot. I'm less clear about packages that don't function until after a reboot, but I believe we might not accept those also.

marco-eckstein:
> We have a policy that packages can't force or cause the system to reboot. I'm less clear about packages that don't function until after a reboot, but I believe we might not accept those also.

That would be bad, because is is not uncommon that programs require a reboot to properly work, not just in this case which is apparently not intended by the publisher. I do not see harm in allowing that.

marco-eckstein:
> This pull request has been automatically marked as stale because it has been marked as requiring author feedback but has not had any recent activity. It will be closed if no further activity occurs **within 3 days of this comment**.

It is unclear what kind of feedback is required.

stephengillie:
Hi @marco-eckstein,

Unfortunately, this package's nature of requiring a reboot before use might make it a poor fit for installation through the WinGet package manager, as it violates our policy of not requiring a reboot before use. I'll close this PR now, to avoid future confusion.

PR #133087, opened 2024-01-08
New package SaferNetworking.RegAlyzer.1 version 1.6.2.16

marco-eckstein:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/133087)


marco-eckstein:
> One or more of the installer URLs doesn't appear valid.
> 
> This may happen for sites with policies prohibiting distribution or use by third parties.
> 
> This may happen for URLs pointing to domains that do not align with the publisher domain or package domain. If you could provide supporting evidence from the publisher that the URLs for the installer are correct, that would help us to validate and approve this PR.

The URL is legit. The installer URL is `https://download.spybot.info/RegAlyzer/regalyz-1.6.2.16.exe`, and the parent domain https://spybot.info redirects to the publisher URL https://www.safer-networking.org/. If not possible to approve this, it would also be possible to change the package URL to https://download.spybot.info/RegAlyzer/. That URL is still dedicated to the package even though not the homepage of the package.

stephengillie (approved):

PR #136754, opened 2024-01-27
New version: CloudFoundry.CLI.v8 version 8.7.1

AshhubAli:

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/136754)

### DIFF
--- a/manifests/c/CloudFoundry/CLI/v8/8.7.1/CloudFoundry.CLI.v8.installer.yaml
+++ b/manifests/c/CloudFoundry/CLI/v8/8.7.1/CloudFoundry.CLI.v8.installer.yaml
@@ -31,7 +31,7 @@ Installers:
   InstallerUrl: https://s3-us-west-1.amazonaws.com/v8-cf-cli-releases/releases/v8.7.1/cf8-cli-installer_8.7.1_winx64.zip
   InstallerSha256: 2B15E48A8B56E63B911ADEF36B5A54D401901B23B6636539BF1544B11E465DE2
   AppsAndFeaturesEntries:
-  - DisplayVersion: v8.7.2
+  - DisplayVersion: v8.7.1
     ProductCode: Cloud Foundry CLI_is1
   ElevationRequirement: elevatesSelf
 - Architecture: x86
@@ -42,7 +42,7 @@ Installers:
   InstallerUrl: https://s3-us-west-1.amazonaws.com/v8-cf-cli-releases/releases/v8.7.1/cf8-cli-installer_8.7.1_win32.zip
   InstallerSha256: D408F21753E2CB09D558DA832BC9D84F23DD1D081B4437D21BB42B3E8A782092
   AppsAndFeaturesEntries:
-  - DisplayVersion: v8.7.2
+  - DisplayVersion: v8.7.1
     ProductCode: Cloud Foundry CLI_is1
   ElevationRequirement: elevatesSelf
 ManifestType: installer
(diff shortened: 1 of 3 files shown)


stephengillie:
> 2024-01-27T06:33:23.1540889Z ##[error] Manifest Error: DisplayVersion declared in the manifest has overlap with existing DisplayVersion range in the index. Existing DisplayVersion range in index:  [ [VERSION, VERSION]]

(Automated response - build 669.)

manifests/c/CloudFoundry/CLI/v8/8.7.1/CloudFoundry.CLI.v8.installer.yaml
```diff
@@ -0,0 +1,49 @@
...
+  NestedInstallerType: inno
+  NestedInstallerFiles:
+  - RelativeFilePath: cf8_installer.exe
+  Scope: machine
+  InstallerUrl: https://s3-us-west-1.amazonaws.com/v8-cf-cli-releases/releases/v8.7.1/cf8-cli-installer_8.7.1_win32.zip
+  InstallerSha256: D408F21753E2CB09D558DA832BC9D84F23DD1D081B4437D21BB42B3E8A782092
+  AppsAndFeaturesEntries:
+  - DisplayVersion: v8.7.1
+    ProductCode: Cloud Foundry CLI_is1
+  ElevationRequirement: elevatesSelf
```
mdanish-kh, line 47:
```suggestion
```

DisplayVersion is not `v8.7.1` in this case. DisplayVersion must match whatever is written to Control Panel / Apps&Features after installing the package. In this case, the `inno` installers don't write a proper version. See -

image

Now we can modify the installer nodes and set DisplayVersion to `VERSION`, but there's another issue with this approach. All versions of this package seem to write the same `VERSION` to Control Panel. WinGet requires the DisplayVersion to be unique for every version of a package. This unfortunately means that we cannot offer the `inno` installers for this package. At the very least, we could only offer it in the latest version and go back and remove in the rest so that it becomes the 'unique' version. But that also has problems of its own (users can face an "infinite upgrade" issue and puts more burden on the contributor to know the context around the package). We should probably just remove these installer nodes as I've suggested above, and only offer the "portable" versions.

mdanish-kh (changes requested):

manifests/c/CloudFoundry/CLI/v8/8.7.1/CloudFoundry.CLI.v8.locale.en-US.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+PackageIdentifier: CloudFoundry.CLI.v8
+PackageVersion: 8.7.1
+PackageLocale: en-US
+Publisher: Cloud Foundry Foundation
+PublisherUrl: https://cli.cloudfoundry.org/
+PublisherSupportUrl: https://github.com/cloudfoundry/cli/issues
+# PrivacyUrl: 
+# Author: 
+PackageName: Cloud Foundry CLI
```
mdanish-kh, line 12:
```suggestion
PackageName: Cloud Foundry CLI v8
```
We can offer a better package name now since we control what is written to Control Panel in case of portable installers.

manifests/c/CloudFoundry/CLI/v8/8.7.1/CloudFoundry.CLI.v8.installer.yaml
NightWalker7558: I initially tried writing `VERSION` to it, but just like you have mentioned, it resulted in violating the unique property. So, just as suggested, I will remove the `inno` installation for this package.

stephengillie (approved):

PR #136894, opened 2024-01-28
Update VCLibs Hash for SandboxTest.ps1

Trenly:
@denelon @stephengillie

* Resolves #136891
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/136894)

### DIFF
--- a/Tools/SandboxTest.ps1
+++ b/Tools/SandboxTest.ps1
@@ -122,7 +122,7 @@ $ProgressPreference = $oldProgressPreference
 
 $vcLibsUwp = @{
   url    = 'https://aka.ms/Microsoft.VCLibs.x64.14.00.Desktop.appx'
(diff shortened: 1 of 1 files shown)


stephengillie (approved):

stephengillie:
Should this be made dynamic in some way? Possibly through checking the latest manifest?

Trenly:
> Should this be made dynamic in some way? Possibly through checking the latest manifest?

Possibly; We could omit the hash entirely, but then we would be relying on the assumption that there hasn’t been a compromise of the CDN, aka.ms link, or any of the domain controllers along the way.

One other option I can look into is installing the winget PowerShell cmdlets and then calling the function to repair winget, which, in theory, should boostrap everything needed

PR #137066, opened 2024-01-29
Use Repair-WinGetPackageManager as a fallback install method

Trenly:

From @stephengillie's point in https://github.com/microsoft/winget-pkgs/pull/136894, this PR ~entirely removes~ reduces the dependency of the SandboxTest script to check the hash of VCLibs, UI.Xaml, or WinGet. If the hash does not match or the file can't be downloaded, the SandboxTest script will now install the `Microsoft.Winget.Client` PowerShell module from the PSGallery and will use the `Repair-WinGetPackageManager` function to bootstrap and install WinGet instead of failing outright.

This ensures that if, or when, https://github.com/microsoft/winget-cli/issues/3380 is completed, the SandboxTest script will not break. ~However, it does mean that it takes longer to bootstrap as the cmdlets will need to download VCLibs, UIXaml, and the correct version of WinGet every time the sandbox is opened - whereas the current script copies them from the disk of the user's device~

cc @mdanish-kh / @stephengillie / @denelon 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/137066)


Tools/SandboxTest.ps1
```diff
@@ -219,26 +153,23 @@ function Get-ARPTable {
...
-$bootstrapPs1Content += @"
 Write-Host @'
 --> Installing WinGet
 '@
-`$ProgressPreference = 'SilentlyContinue'
-Add-AppxPackage -Path '$($desktopAppInstaller.pathInSandbox)' -DependencyPath '$($vcLibsUwp.pathInSandbox)','$($uiLibsUwp.pathInSandbox)'
 
+Install-PackageProvider -Name NuGet -Force | Out-Null
+Install-Module -Name Microsoft.WinGet.Client -Force -Repository PSGallery | Out-Null
+# Need to use Start-Job to prevent progress bars from staying on screen. Could do this by setting the progress preference too, but this seemed cleaner
```
mdanish-kh, line 163: Are progress bars still displayed though? Wording makes it sound that they still are, but Start-Job prevents it from staying indefinitely. I recall Windows PowerShell downloads get incredibly slow with Progress bars, so you may consider going back to the $ProgressPreference = 'SilentlyContinue' approach?

Related:
- https://github.com/PowerShell/PowerShell/issues/2138
- https://github.com/PowerShell/PowerShell/issues/13414

mdanish-kh (commented):
For some reason, I could not get the script to work on my VM. Didn't have much time to debug, but I suspect it's cause I messed something up in my VM. I'll see if I can find the time to debug, but I assume the script is working at your end, so I'm fine with the PR;

```
 ~/winget-pkgs .\Tools\SandboxTest.ps1 .\manifests\j\jqlang\jq\1.6\
--> Validating Manifest

Get-Content: C:\Users\User\winget-pkgs\Tools\SandboxTest.ps1:143
Line |
 143 |            $_.Value = ($((Get-Content "Env:$($_.Name)") + ";$($_.Value …
     |                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot process argument because the value of argument "path" is not valid. Change the value of the "path" argument and run the operation again.
```

Tools/SandboxTest.ps1
Trenly: It shows progress bars for installing the PowershellCmdlets, but not for the rest of it

Trenly:
> For some reason, I could not get the script to work on my VM.

Seems it was an issue with variable parsing. Not sure why it wasn't showing up for me earlier, but this should be fixed now.

mdanish-kh (commented):
Got the script running with the latest commits;

Just realized having these changes does mean we got rid of the local caching we used to do with the binaries; Each run is significantly slower now😥For validating manifests locally, I don't know if I value the script never breaking over script breaking once a while but allowing me to validate much faster every time. I'll leave this to you though

Trenly:
> Got the script running with the latest commits;
> 
> Just realized having these changes does mean we got rid of the local caching we used to do with the binaries; Each run is significantly slower now😥For validating manifests locally, I don't know if I value the script never breaking over script breaking once a while but allowing me to validate much faster every time. I'll leave this to you though

That is a good point and is something I noticed as well.

I wonder if both could be implemented; Try using the local cache, and if theres a hash error, instead of throwing and exiting the script,  use the Repair-WinGetPackageManager method

mdanish-kh:
I like the sound of that, but given my understanding of its implementation, that may mean users won't notice a hash mismatch as the script won't ever fail. If a user isn't aware of this context, they may never realize that the script is executing sub-optimally and don't report an issue to update the hashes.

Maybe we can have better messaging in the script to inform user of the hash mismatch and encourage the user to open an issue at https://github.com/microsoft/winget-pkgs?

Trenly:
> I like the sound of that, but given my understanding of its implementation, that may mean users won't notice a hash mismatch as the script won't ever fail. If a user isn't aware of this context, they may never realize that the script is executing sub-optimally and don't report an issue to update the hashes.
> 
> Maybe we can have better messaging in the script to inform user of the hash mismatch and encourage the user to open an issue at https://github.com/microsoft/winget-pkgs?

I think that I can add a message either way

Trenly:
@mdanish-kh - That was easier than I thought it would be, whenever you want to take a look at it. Easiest way to trigger the fallback is to manually set the hash to an incorrect value

Tools/SandboxTest.ps1
```diff
@@ -165,11 +166,15 @@ foreach ($dependency in $dependencies) {
...
     } catch {
-      #Pass the exception as an inner exception
-      throw [System.Net.WebException]::new("Error downloading $($dependency.url).", $_.Exception)
+      # If the download failed, remove the item so the sandbox can fall-back to using the PowerShell module
+      Remove-Item $dependency.SaveTo -Force | Out-Null
     }
     if (-not ($dependency.hash -eq $(Get-FileHash $dependency.SaveTo).Hash)) {
-      throw [System.Activities.VersionMismatchException]::new('Dependency hash does not match the downloaded file')
+      # If the hash didn't match, remove the item so the sandbox can fall-back to using the PowerShell module
+      Write-Host -ForegroundColor Red '      Dependency hash does not match the downloaded file'
```
mdanish-kh, line 174:
```suggestion
      Write-Host -ForegroundColor Red "      Dependency hash does not match the downloaded file $(Split-Path -Path $dependency.SaveTo -Leaf)"
```

nit: we can add the file name, since it can theoretically print twice (not likely tho with the UI.Xaml Url)

mdanish-kh (approved):
Love this, just a few minor comments

Tools/SandboxTest.ps1
Trenly: We could, but we shouldn't need to -
image

mdanish-kh: Ah - I wasn't paying attention to the full output. Looks correct

Tools/SandboxTest.ps1
```diff
@@ -165,11 +166,15 @@ foreach ($dependency in $dependencies) {
...
+      Write-Host -ForegroundColor Red '      Please open an issue referencing this error at https://bit.ly/WinGet-SandboxTest-Needs-Update'
```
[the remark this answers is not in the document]
denelon: Sadly, I cannot. The URL is too long and has special characters not supported in aka.ms links.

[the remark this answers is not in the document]
mdanish-kh: Ah, that's unfortunate.

@stephengillie / @denelon This PR is ready for merge in that case

denelon (approved):

PR #138952, opened 2024-02-13
Remove version: Fort.Firewall version 3.11.4

ghost:
Replaced by https://github.com/microsoft/winget-pkgs/pull/138957
---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/138952)


Emi-Emi-Emi:
This should be made easier, instead of throwing an error. Just look at Fort Firewall, two updates in the same day, so it is hard to keep a single version because old versions are pretty much useless, but I hate to see these errors instead of just waiting for a mod to approve it or not based on the reasoning for the removal of the version, which is in this case is just to keep a single version as much as possible since the latest one will always be the only one needed.

I am the only one trying to bring Fort Firewall to Winget, so I don't want to see 8000 useless versions in a single year and then no way to clean them up.

Hope it can be improved. I will try to clean new versions eventually, but since Fort Firewall is some updates from time to time to too many in few hours it's harder to maintain on Winget just a single latest version.

stephengillie (approved):

Issue #90642, opened 2022-12-05
[Package Issue]: Discord.Discord  - Autostart with OS forcefully enabled on upgrade even if user does not want this.


R-Adrian:
### Please confirm these before moving forward

- [X] I have searched for my issue and not found a work-in-progress/duplicate/resolved issue.
- [X] I have not been informed if the issue is resolved in a preview version of the winget client.

### Category of the issue

Installation issue.

### Brief description of your issue

i keep my Discord app settings to have the Startup settings to always be OFF by default, but yesterday when i upgraded Discord via winget, it forcefully set the app to start with the OS, even if the in-app setting is to not start with OS.

### Steps to reproduce

1) install an older version of Discord  (e.g. 1.0.9006)
2) configure it to always start manually, not at OS/user logon
3) run "winget upgrade Discord.Discord"  (currently this will upgrade it to 1.0.9007)

### Actual behavior

either winget or the Discord installer has set the app startup mode to always start at OS/User logon and completely bypassed the user setting to NOT do that (from the Discord settings).

### Expected behavior

Respect the user setting/preference and do not automatically start Discord at logon.

### Environment

```shell
Windows Package Manager v1.3.2691
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.19045.2311
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.18.2691.0

Logs: %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir

Links
---------------------------------------------------------------------------
Privacy Statement   https://aka.ms/winget-privacy
License Agreement   https://aka.ms/winget-license
Third Party Notices https://aka.ms/winget-3rdPartyNotice
Homepage            https://aka.ms/winget
Windows Store Terms https://www.microsoft.com/en-us/storedocs/terms-of-sale



winget show discord.discord
Found Discord [Discord.Discord]
Version: 1.0.9007
Publisher: Discord Inc.
Publisher Url: https://discord.com
Publisher Support Url: https://support.discord.com
Author: Discord Inc.
Moniker: discord
Description: Whether you’re part of a school club, gaming group, worldwide art community, or just a handful of friends that want to spend time together, Discord makes it easy to talk every day and hang out more often.
Homepage: https://discord.com
License: Proprietary
License Url: https://discord.com/licenses
Privacy Url: https://discord.com/privacy
[...]

PR #139072, opened 2024-02-14
Discord.Discord.x64 version 1.0.9031

### LINKED ISSUE #137934: [Update Request]: Splitting Discord.Discord into two packages for each architecture
### What type of update are you requesting?

A new locale, architecture, or installer type of an existing version

### Current Package Identifier

Discord.Discord

### Package Version

1.0.9030

### Please describe the changes you would like to see

Would it be possible to split Discord into two packages: `Discord.Discord.x86` & `Discord.Discord.x64`

The current system keeps showing me updates for x86 Discord when i have x64 Discord installed.

drshtmstry:

---

- Resolves #137934

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/139072)

### DIFF
--- a/manifests/d/Discord/Discord/x64/1.0.9031/Discord.Discord.x64.installer.yaml
+++ b/manifests/d/Discord/Discord/x64/1.0.9031/Discord.Discord.x64.installer.yaml
@@ -1,0 +1,18 @@
+# Created using wingetcreate 1.5.7.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.5.0.schema.json
+
+PackageIdentifier: Discord.Discord.x64
+PackageVersion: 1.0.9031
+InstallerLocale: en-US
+InstallerType: exe
+InstallModes:
+- interactive
+- silent
+Protocols:
+- discord
+Installers:
+- Architecture: x64
+  InstallerUrl: https://discord.com/api/downloads/distributions/app/installers/latest?channel=stable&platform=win&arch=x64
+  InstallerSha256: A201708C74975F82BFDF900331893C9A52455ED8F8C38E68CAEFC652C8677313
+ManifestType: installer
+ManifestVersion: 1.5.0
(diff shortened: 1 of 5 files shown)


Trenly:
As I mentioned in the issue, Discord does not have a way to identify which architecture is installed, and this may cause more issues. 

I don't believe splitting the packages will help, and it may cause more issues than it solves

stephengillie:
Review file blocks this package: 

> Self-updates and does not update the version number in the registry when doing so - applying an update via winget will break/override the user preferences and forcefully enable auto-start on system boot.

- #90642

drshtmstry:
> As I mentioned in the issue, Discord does not have a way to identify which architecture is installed, and this may cause more issues.
> 
> I don't believe splitting the packages will help, and it may cause more issues than it solves

If you are right then when x64 is installed, it should update to x86 newer version automatically.
I tested this by installing x64 from manifest and it did not upgrade.
Previously it only showed winget upgrade because both architecture were using same ID & x86 had higher version number.

Screenshot 2024-02-15 085334
Screenshot (19)
Screenshot 2024-02-15 085818

Trenly:
> If you are right then when x64 is installed, it should update to x86 newer version automatically. I tested this by installing x64 from manifest and it did not upgrade. Previously it only showed winget upgrade because both architecture were using same ID & x86 had higher version number.

Thats not quite what I meant. Discord itself is not going to detect updates incorrectly - internally it knows which architecture it is running.

The issue is that winget will have no way to know which version a user has installed. If you look at the information Discord writes to the registry, the DisplayName, Publisher, and ProductCode are exactly the same between the x86 and x64 versions. This is why they can be in the same manifest without requiring `AppsAndFeaturesEntries`. If the two manifests are split, winget may not be able to determine which package is actually installed since both `Discord.Discord.x86` and `Discord.Discord.x64` will be a strong match to the data in the registry.

xAphex:
> > If you are right then when x64 is installed, it should update to x86 newer version automatically. I tested this by installing x64 from manifest and it did not upgrade. Previously it only showed winget upgrade because both architecture were using same ID & x86 had higher version number.
> 
> Thats not quite what I meant. Discord itself is not going to detect updates incorrectly - internally it knows which architecture it is running.
> 
> The issue is that winget will have no way to know which version a user has installed. If you look at the information Discord writes to the registry, the DisplayName, Publisher, and ProductCode are exactly the same between the x86 and x64 versions. This is why they can be in the same manifest without requiring `AppsAndFeaturesEntries`. If the two manifests are split, winget may not be able to determine which package is actually installed since both `Discord.Discord.x86` and `Discord.Discord.x64` will be a strong match to the data in the registry.

How about deprecating x86 version all together?

PR #139462, opened 2024-02-16
[Agreements] Add back XmacsLabs.Mogan

mdanish-kh:
cc @denelon @stephengillie 

Installer prompt:
image

### DIFF
--- a/manifests/x/XmacsLabs/Mogan/1.2.0/XmacsLabs.Mogan.locale.zh-CN.yaml
+++ b/manifests/x/XmacsLabs/Mogan/1.2.0/XmacsLabs.Mogan.locale.zh-CN.yaml
@@ -20,7 +20,8 @@ Description: |-
   墨干编辑器是墨者实验室旗下的结构化编辑器。
   墨干编辑器基于 GNU TeXmacs 研发定制的自由软件。
 Agreements:
-  - AgreementUrl: https://github.com/XmacsLabs/mogan/blob/branch-1.2/LICENSE
+  - AgreementLabel: GNU 通用公共许可证
+    AgreementUrl: https://github.com/XmacsLabs/mogan/blob/branch-1.2/LICENSE
 # Moniker:
 Tags:
 - latex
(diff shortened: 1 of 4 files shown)


manifests/x/XmacsLabs/Mogan/1.2.0/XmacsLabs.Mogan.locale.zh-CN.yaml
```diff
@@ -0,0 +1,42 @@
...
+License: GPL-3.0
+LicenseUrl: https://gitee.com/XmacsLabs/mogan/blob/branch-1.2/LICENSE
+Copyright: © 2020-2023 by Xmacs Labs
+CopyrightUrl: https://gitee.com/XmacsLabs/mogan/blob/branch-1.2/COPYING
+ShortDescription: 基于 GNU TeXmacs 的结构化编辑器
+Description: |-
+  墨干编辑器是墨者实验室旗下的结构化编辑器。
+  墨干编辑器基于 GNU TeXmacs 研发定制的自由软件。
+Agreements:
+  - AgreementUrl: https://github.com/XmacsLabs/mogan/blob/branch-1.2/LICENSE
```
stephengillie, line 23:
```suggestion
  - AgreementLabel: GNU GENERAL PUBLIC LICENSE
    AgreementUrl: https://github.com/XmacsLabs/mogan/blob/branch-1.2/LICENSE
```

stephengillie (changes requested):
Is the `AgreementLabel` necessary? I think it is.

manifests/x/XmacsLabs/Mogan/1.2.0/XmacsLabs.Mogan.locale.zh-CN.yaml
mdanish-kh: It's not necessary, see https://github.com/microsoft/winget-pkgs/pull/139478. But it's better to add one. Though I think `License agreement` is more suitable of a label here? (since it conveys what type of an agreement it is)

mdanish-kh: Got the translation for `GNU GENERAL PUBLIC LICENSE` from https://www.gnu.org/licenses/old-licenses/gpl-2.0.zh-cn.html

Trenly:
> Is the `AgreementLabel` necessary? I think it is.

Strictly speaking, the CLI will work without the Label, but it is probably a best practice to include it

image

ItzLevvie (approved):

stephengillie (approved):

PR #140843, opened 2024-02-24
New version: dail8859.NotepadNext version 0.7

dail8859:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/140843)


stephengillie (approved):

dail8859:
@stephengillie 

It does. For future reference, is it preferred to keep the 3 digit version numbering? Meaning it is preferable to use `0.7.0` rather than just `0.7`?

image

stephengillie:
Hi @dail8859,

I searched semantic versioning guidance and couldn't find advice on dropping a trailing zero. So this is left to developer discretion. And so is using semantic versioning. 

The version number in the manifest has to match the Apps and features page, as in your screenshot. This is part of how installed versions are matched up with manifest versions, so our package manager knows if an upgrade is available.

PR #141342, opened 2024-02-28
Update SandboxTest Again now that the packages are published

Trenly:
@denelon @mdanish-kh

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/141342)

### DIFF
--- a/Tools/SandboxTest.ps1
+++ b/Tools/SandboxTest.ps1
@@ -20,7 +20,7 @@ Param(
 
 $ErrorActionPreference = 'Stop'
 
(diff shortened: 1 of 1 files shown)


mdanish-kh (approved):
We probably don't need the nuget flow anymore, but doesn't hurt keeping it in there

Trenly:
> We probably don't need the nuget flow anymore, but doesn't hurt keeping it in there

Exactly why I put in the logic to have it controlled by the variable

denelon (approved):

PR #142814, opened 2024-03-07
Update AUTHORING_MANIFESTS.md (take 2)

stephengillie:

---
Previous PR https://github.com/microsoft/winget-pkgs/pull/141399 became closed when trying to fix a merge conflict. 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/142814)


AUTHORING_MANIFESTS.md
```diff
@@ -77,3 +86,135 @@ It is important to test your manifest before submission to ensure it meets the r
 
 After enabling the setting for local manifests (`winget settings --enable LocalManifestFiles`), manifests can be tested locally with `winget install --manifest <path>`.
 If your system supports Windows Sandbox, you can also use the [SandboxTest.ps1 Script](https://github.com/microsoft/winget-pkgs/blob/master/doc/tools/SandboxTest.md) to test the manifest in the Windows Sandbox. This is the preferred method, as it ensures the package doesn't require any dependencies to install.
+
+## Version Matching & Correllation
+
+The goal is to accurately determine the currently installed version, and all available versions of this application, so upgrades can be correctly suggested when available, and not suggested when not available. Most installers write accurate version data to the Windows Registry. But some do not, and `Portable` type applications generally cannot as they usually don't have installers. The data available to meet this goal consists of the files on disk, data in Windows Registry, and manifests in the winget-pkgs repo.
```
Trenly, line 92: Winget writes the ARP entries for portable installer type, so they are guaranteed to match

AUTHORING_MANIFESTS.md
```diff
@@ -77,3 +86,135 @@ It is important to test your manifest before submission to ensure it meets the r
...
+
+### Metadata 
+
+- DisplayName
+- DisplayVersion
+  - Displayed in the Control Panel's Add & Remove Programs page (ARP) and/or Apps and features Settings page (ANF). 
+  - Should not be used if it's always equal to PackageVersion.
+  - Must be unique per package. if multiple versions are sharing the same DisplayVersion, we may only keep the 'latest' one in the repo.
```
Trenly, line 100:
```suggestion
  - Must be unique per version. if multiple versions are sharing the same DisplayVersion, we may only keep the 'latest' one in the repo.
```

AUTHORING_MANIFESTS.md
```diff
@@ -77,3 +86,135 @@ It is important to test your manifest before submission to ensure it meets the r
...
+- Package Familyname (MSIX)
```
Trenly, line 101: This isn’t a valid AppsAndFeaturesEntry field and can only be used outside of the AppsAndFeaturesEntries

AUTHORING_MANIFESTS.md
```diff
@@ -77,3 +86,135 @@ It is important to test your manifest before submission to ensure it meets the r
...
+- ProductCode & Upgradecode (MSI)
```
Trenly, line 102:
```suggestion
- ProductCode & UpgradeCode (MSI)
```

Trenly, line 102: Exe's should also have product codes. Even though they aren’t contained in the file the same way an MSI has it defined in the database, the registry key the EXE writes into defines the product code

AUTHORING_MANIFESTS.md
```diff
@@ -77,3 +86,135 @@ It is important to test your manifest before submission to ensure it meets the r
...
+
+### 4 main use cases: 
+
+- Package does not write DisplayVersion to registry.
+  - DisplayName contains version. DisplayName is required in every manifest, to match installed packages to available versions.
+  - ProductCode contains version. ProductCode is required in every manifest, to match installed packages to available versions.
+- PackageVersion differed from DisplayVersion at any point in the manifest history. PackageVersion is required in every manifest, to prevent version range mapping errors.
+- ProductCode of the installed product differs from the ProductCode of the installer. ProductCode is required in at least one manifest to match installed packages to available packages.
+
+Note: Having incorrect values may have adverse effects, so these shouldn't be used unless there are reports of matching issues.
```
Trenly, line 113: This isn’t univerally true, it just depends on the field. ProductCode and UpgradeCode is added for most MSI packages automatically even if there aren’t matching issues, and the same for PackageFamilyName with MSIX packages. I'll even add the product code for an exe if I have time. Of course having a wrong value will cause issues, but not all fields need to be avoided

AUTHORING_MANIFESTS.md
```diff
@@ -77,3 +86,135 @@ It is important to test your manifest before submission to ensure it meets the r
...
+
+## Sorting and sort order
+
+As long as a PackageVersion sticks to either semantic or string, it's pretty easy to sort. But switching usually causes the order to become confused. Insofar as being YAML compliant, these are all supposed to be strings, leading to a few manifests having their PackageVersion numbers nested in quotes.
```
Trenly, line 117: This is very confusing the way it is worded. Inherently all the versions are strings - a sematic version is just a string with a certain format. YAML will interpret any alpha-numerics as strings automatically, and if there are multiple `.` it is smart enough to interpret that value as a string also. The quotes are only needed when the value is also a valid decimal number as `1.23` can be interpreted as string or a number whereas `"1.23"` can only be interpreted as a string

AUTHORING_MANIFESTS.md
```diff
@@ -77,3 +86,135 @@ It is important to test your manifest before submission to ensure it meets the r
...
+
+- A "version parameter" is a set of numbers between dots in a version number - major, minor, patch, build, etc. 
+- Digits are sorted within a version parameter, with further-left parameters having magnitude over further-right parameters. 
+  - This might disrupt some temporal versioning schemae, such as `DD.MM.YYYY`, as a version from 31 Mar 2023 (`31.03.2023`) would be sorted as newer than one from 1 Feb 2024 (`01.02.2024`). 
+
+### String
+
+If any letters, spaces, or other non-numeric characters (other than dots) are added, the PackageVersion will fallback to string interpretation. 
+
+- Digits are sorted independently, and version `3.9.E` would be sorted higher than `3.10.B`. because 9 is larger than 1. This system doesn't have the concept of version parameters.
```
Trenly, line 131: This is not how winget parses versions. Winget breaks at each `.` and compares each part individually. Where a part contains an numeral-alpha, the numeral is considered before the alpha.

For your example, the first comparison results in `3 == 3`, so then the second part is evaluated, `10 > 9`, so `3.10.B` is greater than `3.9.E`. The comparison returns and the third part is never evaluated.

I suggest reading through the commented code over at winget-cli for more detail on the nuances of how versions are broken into Parts, and Parts are broken into "integer" and "other"
https://github.com/microsoft/winget-cli/blob/8a006549c4aa0dd13cc17f6185ce30a1f4c2e71b/src/AppInstallerSharedLib/Public/AppInstallerVersions.h#L19

AUTHORING_MANIFESTS.md
```diff
@@ -77,3 +86,135 @@ It is important to test your manifest before submission to ensure it meets the r
...
+
+Notes, ideas, and discussion on how to implement and support release channels [is ongoing](https://github.com/microsoft/winget-cli/issues/147).
+
+## Troubleshooting
+
+###  Display Version Overlap: 
+
+- Can be caused by the PackageVersion differring from the DisplayVersion at any point in the manifest history, and the fix is to add the PackageVersion to every manifest. Another option is to remove the manifests where they differ.
+- Actual log example: `2024-01-31T22:16:48.6603831Z ##[error] Manifest Error: DisplayVersion declared in the manifest has overlap with existing DisplayVersion range in the index. Existing DisplayVersion range in index: [ [5.17.5 (31030), 5.17.5 (31030)]]`
+- DisplayVersion is treated as a range, from the lowest manifest version in the repo to the highest.
```
Trenly, line 189: This is incorrect. It is from the lowest DisplayVersion in a single manifest to the highest DisplayVersion in the *same* manifest. Each manifest creates its own version range. Two manifests cannot have overlapping ranges, but any given package can have multiple ranges if there are multiple manifests

AUTHORING_MANIFESTS.md
```diff
@@ -77,3 +86,135 @@ It is important to test your manifest before submission to ensure it meets the r
...
+
+## Troubleshooting
+
+###  Display Version Overlap: 
+
+- Can be caused by the PackageVersion differring from the DisplayVersion at any point in the manifest history, and the fix is to add the PackageVersion to every manifest. Another option is to remove the manifests where they differ.
+- Actual log example: `2024-01-31T22:16:48.6603831Z ##[error] Manifest Error: DisplayVersion declared in the manifest has overlap with existing DisplayVersion range in the index. Existing DisplayVersion range in index: [ [5.17.5 (31030), 5.17.5 (31030)]]`
+- DisplayVersion is treated as a range, from the lowest manifest version in the repo to the highest. 
+- This might lead to unexpected matching situations. For example, if the version ranges are 5.17.2 to 5.17.29988, and a PR's version is 5.17.5, then it falls inside that range.
+- In the future, it might be changed from a range, to an array of known manifest DisplayVersions, and require exact matching.
```
Trenly, line 191: Talking about future can be confusing. I would rather see "There is ongoing discussion around this behavior in <issue_link>" or no mention at all

AUTHORING_MANIFESTS.md
```diff
@@ -77,3 +86,135 @@ It is important to test your manifest before submission to ensure it meets the r
...
+
+### Scope swapping, accidental side-by-side installs, and one of the packages refuses to upgrade: 
+
+- This issue appears to be common to MSI installers.
```
Trenly, line 195: It is specific to WIX, not all MSI

AUTHORING_MANIFESTS.md
```diff
@@ -77,3 +86,135 @@ It is important to test your manifest before submission to ensure it meets the r
...
+
+- Two manifests have the same data. If both the variants use the same DisplayName, ProductCode, Publisher and other package matching related information, then an existing install of a stable package may be mapped to a higher available version of a pre-release version by WinGet, even if the PackageIdentifiers are different. The fix is to remove the pre-release manifest, as the mainline has priority.
+- Version schema changes significantly - or changes between string and semantic. If a letter is added to one version number, then it can cause this issue with every version of the package. The fix is to remove manifests with the previous version schema, and only offer those with the latest schema.
+- Installer not writing accurate version data to the Windows Registry - either every installer writes the same version number, do not write a version number, do not write a consistent or accurate version number or have a similar issue with the version data in the registry. The fix here is to add the Apps and features metadata described above.
```
Trenly, line 215: Or it could be that the manifest has an incorrect PackageVersion specified too

AUTHORING_MANIFESTS.md
```diff
@@ -77,3 +86,135 @@ It is important to test your manifest before submission to ensure it meets the r
...
+
+### For a package that always writes the same version to registry: 
+
+Our package manager could update the Windows Registry, modify the installer nodes, and set DisplayVersion for the installer. But if the user or application performed any activity (reinstall, repair, update, uninstall, etc) outside of the package manager, then it would overwrite, or not write, this data.
```
Trenly, line 219: Winget can't update the registry or set DisplayVersion, except for portable applications.

Trenly (changes requested):

AUTHORING_MANIFESTS.md
stephengillie: If 2 packages have the same DisplayVersion, won't this cause an update loop?

Trenly: > If 2 packages have the same DisplayVersion, won't this cause an update loop?

What are you considering a package? My understanding is that a `Package` is a unique `PackageIdentifier`. For example, `Google.Chrome` and `Microsoft.Office` can have the same DisplayVersion.

If referring to individual manifests within the same PackageIdentifier, I would consider that as a single `manifest` or `version`. I've also seen `Package-version` or `Package version`

AUTHORING_MANIFESTS.md
stephengillie: This is a direct quote from @mdanish-kh. (Most of these lines are from him, you or denelon.)

AUTHORING_MANIFESTS.md
```diff
@@ -128,8 +128,10 @@ If composed solely of numbers and dots, the PackageVersion will generally be int
 
 If any letters, spaces, or other non-numeric characters (other than dots) are added, the PackageVersion will fallback to string interpretation. 
 
-- Digits are sorted independently, and version `3.9.E` would be sorted higher than `3.10.B`. because 9 is larger than 1. This system doesn't have the concept of version parameters. 
-- This includes all versions. So if a package has manifests with PackageVersions `0.1.1`, `v0.1.3`, and `v0.1.4` in the winget-pkgs repo, then the string sorting determines `0.1.1` as the highest version number. 
+- Winget breaks at each dot and compares each part individually. Where a part contains an numeral-alpha, the numeral is considered before the alpha.
```
Trenly, line 131: 😄 I didn't mean that this should be included literally, I was hoping you might be able to take a look at how the versions are actually sorted and update the markdown to match

AUTHORING_MANIFESTS.md
```diff
@@ -213,8 +216,8 @@ This issue is slightly different from scope swapping above, in that there's only
 - Two manifests have the same data. If both the variants use the same DisplayName, ProductCode, Publisher and other package matching related information, then an existing install of a stable package may be mapped to a higher available version of a pre-release version by WinGet, even if the PackageIdentifiers are different. The fix is to remove the pre-release manifest, as the mainline has priority.
 - Version schema changes significantly - or changes between string and semantic. If a letter is added to one version number, then it can cause this issue with every version of the package. The fix is to remove manifests with the previous version schema, and only offer those with the latest schema.
 - Installer not writing accurate version data to the Windows Registry - either every installer writes the same version number, do not write a version number, do not write a consistent or accurate version number or have a similar issue with the version data in the registry. The fix here is to add the Apps and features metadata described above.
+- Or it could be that the manifest has an incorrect PackageVersion specified too
```
Trenly, line 219: May want to reformat this to match the writing style of the above points

Trenly (commented):
My apologies for not being clearer in my initial review - All of my comments are more for your own guidance and information, so you can make informed updates to the content. I don't believe my comments can just be dropped into the document, they are just identifying gaps I see or areas where confusion could occur

AUTHORING_MANIFESTS.md
stephengillie: > If referring to individual manifests within the same PackageIdentifier, I would consider that as a single `manifest` or `version`. I've also seen `PackageVersion`, `Package-version`, or `Package version`

Yes, these. If two of them have the same DisplayVersion, won't this cause an update loop?

AUTHORING_MANIFESTS.md
stephengillie: This implies that metadata might always be unnecessary for Portable packages. Is this true?

AUTHORING_MANIFESTS.md
stephengillie: > Each manifest creates its own version range.

That's unexpected, and might contribute to the confusion around how these entries drive matching and sort order.

AUTHORING_MANIFESTS.md
stephengillie: I'll have to revise my canned messages.

AUTHORING_MANIFESTS.md
Trenly: Yes

AUTHORING_MANIFESTS.md
stephengillie: Some parts aren't the most clear - I'm not sure about the `other Other`. (Converted to Polyglot)

- If Part is less than the other Part
  - if Integer is less than other Integer, return true.
    - else if Integer is greater than other Integer, return false.
    - else if Other is empty, return false. (If this Other is empty, it is at least is greater than or equal to.)
    - else if Other is not empty and other Other is empty, return true. (If the other Other is empty and this is not, this is less.)
    - else if m_foldedOther is less than other m_foldedOther, return true. (Compare the folded versions.)
	(else Other is greater than or equal to other Other, return false.)
- If Part is equal to other Part, return Integer is equal to the other Integer and m_foldedOther is equal to other m_foldedOther.
- If Part is not equal to other Part, return pointer-this is not equal to other 
- If Part is greater than other Part, return m_channel is less than other m_channel.

Trenly: `other Other` is a bit confusing for sure!

Consider the version `1.2.34ab5`. Breaking it into parts looks like this, where each line is a `Part`. To compare a version, you start at the leftmost `Part`, if they are equal then you continue to the next `Part`. If they are not equal, the `Part` which is greater belongs to the `Version` which is greater

| Integer | Other |
| --- | --- |
| 1 | null |
| 2 | null |
| 34 | ab5 |

So when comparing `1.2.34ab5` and `1.2.34`
| this.Integer | this.Other | other.Integer | other.Other |
| --- | --- | -- | -- |
| 1 | null | 1 | null |
| 2 | null | 2 | null |
| 34 | ab5 | 34 | null |

1. Is `Integer` less than `other Integer`? 
  a. No, they are equal, continue
2. Is `other` present ? 
  a. No, it is not. Since `this other` is empty, the version is at least greater than or equal to, return false

\- Next Part -
1. Is `Integer` less than `other Integer`? 
  a. No, they are equal, continue
2. Is `other` present ? 
  a. No, it is not. Since `this other` is empty, the version is at least greater than or equal to, return false

\ - Next Part -
1. Is `Integer` less than `other Integer`? 
  a. No, they are equal, continue
2. Is `other` present ? 
  a. Yes, it is, continue
3. Is `other Other` present?
  b. No, it is not. `this` Part is less than `other`, return true.
  (Note: this seems counter-intuitive that it would be less, but is what causes 1.2.34 to be considered greater than 1.2.34-rc or `1.2.34b`)

stephengillie:
Close with reason: Superseded by #144543;

PR #143386, opened 2024-03-09
Remove version: OpenJS.NodeJS.LTS version 20.11.1

CoolPlayLin:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/143386)


mdanish-kh (changes requested):
I don't quite get the reason, @CoolPlayLin can you please elaborate why would you like these manifests removed?

mdanish-kh:
> I don't quite get the reason, @CoolPlayLin can you please elaborate why would you like these manifests removed?

I see, the other PRs you're referring are these ones:

- https://github.com/microsoft/winget-pkgs/pulls?q=is%3Aopen+is%3Apr+author%3Acoolplaylinbot+OpenJS

I don't agree with these changes, why would you want to prepend `v` in the version and use `DisplayVersion` field? The version written by the NodeJS installer exactly matches the current PackageVersions in the repo. I don't see a need for this change

Also, mass modifications should really be discussed in an issue first, so that it's easier to request changes and get to a consensus

mdanish-kh:
https://www.github.com/microsoft/winget-pkgs/pull/143386#issuecomment-1986949152 [Policy] Needs-Author-Feedback

CoolPlayLin:
@mdanish-kh This change can better distinguish this from OpenJS.NodeJS

mdanish-kh:
@CoolPlayLin Unfortunately, these changes won't help in distinguishing the packages better. WinGet would still use DisplayVersion internally for package matching, and the matching would remain the same as before. Having DisplayVersion will introduce an additional version layer that is unnecessary, and it'll bring more churn in updating the manifests. In addition, if someone misses adding DisplayVersion in the newer manifest, this can cause issues like upgrade loop scenarios. I think the current state of the manifests is good, and doesn't need changing

mdanish-kh:
Close with reason: Changes not needed, see discussion above;

PR #146328, opened 2024-03-25
Remove: TechPowerUp.GPU-Z version 2.57.0

coolplaylinbot:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/146328)


stephengillie:
URL: https://us2-dl.techpowerup.com/files/GPU-Z.2.57.0.exe 
Response status code does not indicate success: 404 (Not Found).

(Automated message - build 872)

stephengillie:
Hi @coolplaylinbot,

(Automated response - build 872.)

stephengillie:
URL: https://us2-dl.techpowerup.com/files/GPU-Z.2.57.0.exe 
Response status code does not indicate success: 404 (Not Found).

(Automated message - build 873)

mdanish-kh:
This is because TechPowerUp takes advantage of the user agent of package managers to avoid hotlinking. This static URL will not work for you in the browser, but will work when installing the package using WinGet.

cc @stephengillie - We may want to include TechPowerUp packages in the exclusion list if a PR comes around removing those packages


image

mdanish-kh:
Close with reason: See above comments;

stephengillie:
> This is because TechPowerUp takes advantage of the user agent of package managers to avoid hotlinking. This static URL will not work for you in the browser, but will work when installing the package using WinGet.
> 
> cc @stephengillie - We may want to include TechPowerUp packages in the exclusion list if a PR comes around removing those packages

This sounds like the opposite of the hosts which block package manager user agents but allow browser downloads. I'll add this to the Review list.

PR #150629, opened 2024-04-23
New version: Google.AndroidStudio 2023.2.1.25

aisgbnok:
Checklist for Pull Requests

Manifests
- [X] Have you [validated](https://github.com/microsoft/winget-pkgs/blob/master/doc/Authoring.md#validation) your manifest locally with `winget validate --manifest <path>`?
- [X] Does your manifest conform to the [1.6 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.6.0)?

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/150629)

### DIFF
--- a/manifests/g/Google/AndroidStudio/2023.2.1.25/Google.AndroidStudio.installer.yaml
+++ b/manifests/g/Google/AndroidStudio/2023.2.1.25/Google.AndroidStudio.installer.yaml
@@ -1,0 +1,23 @@
+# Created using wingetcreate 1.6.1.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+
+PackageIdentifier: Google.AndroidStudio
+PackageVersion: 2023.2.1.25
+InstallerType: nullsoft
+Scope: machine
+InstallModes:
+- interactive
+- silent
+InstallerSuccessCodes:
+- 1223
+UpgradeBehavior: install
+AppsAndFeaturesEntries:
+- DisplayVersion: "2023.2"
+ElevationRequirement: elevatesSelf
+ReleaseDate: 2024-04-09
+Installers:
+- Architecture: x64
+  InstallerUrl: https://redirector.gvt1.com/edgedl/android/studio/install/2023.2.1.25/android-studio-2023.2.1.25-windows.exe
+  InstallerSha256: EA891F977305DFAF58DFAFED7908A6B14D7FDD9D2B024A80A13BD80B3A82C346
+ManifestType: installer
+ManifestVersion: 1.6.0
--- a/manifests/g/Google/AndroidStudio/2023.2.1.25/Google.AndroidStudio.locale.en-US.yaml
+++ b/manifests/g/Google/AndroidStudio/2023.2.1.25/Google.AndroidStudio.locale.en-US.yaml
@@ -1,0 +1,44 @@
+# Created using wingetcreate 1.6.1.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.6.0.schema.json
+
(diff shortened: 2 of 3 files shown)


aisgbnok:
@stephengillie,

>  Manifest Error: DisplayVersion declared in the manifest has overlap with existing DisplayVersion range in the index. Existing DisplayVersion range in index:  [ [2023.2, 2023.2]]

Google.AndroidStudio `2023.2.1.25` and `2023.2.1.23` use the same `DisplayVersion`: `"2023.2"`. Unfortunately, Android Studio only writes `MAJOR.MINOR` to registry/arp entries, no `.PATCH`, etc.

I attempted to skirt around this by adding `InstallationMetadata`:

https://github.com/microsoft/winget-pkgs/blob/f9a90b9ab8dece8b7f3c3f096963b8a1d8b43c25/manifests/g/Google/AndroidStudio/2023.2.1.25/Google.AndroidStudio.installer.yaml#L18-L25

But it didn't work.

Is there any other way to resolve this? or can there only be one package with the same `DisplayVersion`? 

If that is the case, then would it be acceptable to remove `2023.2.1.23` and add/replace with `2023.2.1.25`? Thanks.

stephengillie:
> can there only be one package with the same DisplayVersion?

I believe so, yes. The `DisplayVersion` creates a Range with itself as the only member, and doesn't allow overlaps. Since both have to have that same range, there can be only one in the repo at a time.

aisgbnok:
> I believe so, yes. The `DisplayVersion` creates a Range with itself as the only member, and doesn't allow overlaps. Since both have to have that same range, there can be only one in the repo at a time.

Thanks, and your right only a single `DisplayVersion` at a time.

So I have removed `2023.2.1.23` and added `2023.2.1.25` in this single PR to ensure only one package with `DisplayVersion` `"2023.2"`. However, the validation fails because the `DisplayVersion` is still present in the index.

Is it possible to do an update like this without causing a situation where the `"2023.2"` version is momentarily removed from the index/unpublished? Should this be done in two PRs or one like this? What is preferred for a situation like this? Thank you for your insight!

stephengillie:
> 2024-04-26T19:56:19.0351217Z ##[error]Manifest Validation Failed
 2024-04-26T19:56:24.1005379Z ##[section]Finishing: Validate Manifest

(Automated response - build 877.)

stephengillie:
I'm not sure of the best way to proceed here. 

@mdanish-kh and @Trenly, could either of you advise on the best course here?

mdanish-kh:
@aisgbnok The validation works by scanning the current index for a duplicate DisplayVersion. In order to resolve this issue, the previous version should be removed in a separate PR. After that gets published and the index gets updated, we can re-run pipelines here and add the new version in

Trenly:
I've run into this issue in the past myself and the guidance I received was exactly as @mdanish-kh said, as the pipelines do not create a new index and check it as part of the pipeline run (creating the index is conputationally expensive, the publish pipeline is separate for this reason)

Perhaps opening a [feature request](https://github.com/microsoft/winget-pkgs/issues/new/choose) with more details might give @hackean-msft some ideas on how a commit processor could potentially apply the diffs from the commits to the existing index without rebuilding it, but for now all we can do is remove and then add the new version

Trenly:
> @wingetbot run

You need to wait for the removal to pass the publish pipeline

aisgbnok:
Index published so I've rebased on `origin/master`.

- https://github.com/microsoft/winget-pkgs/pull/151501#issuecomment-2083621293

It should work this time. Thank you everyone for your help.

Trenly (approved):

PR #153871, opened 2024-05-16
add git-tfs 0.34.0

pmiossec:
Checklist for Pull Requests
- [x] Is there a linked Issue? No.

Manifests

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/153871)

### DIFF
--- a/manifests/g/git-tfs/git-tfs/0.34.0/git-tfs.git-tfs.installer.yaml
+++ b/manifests/g/git-tfs/git-tfs/0.34.0/git-tfs.git-tfs.installer.yaml
@@ -1,0 +1,18 @@
+PackageIdentifier: git-tfs.git-tfs
+PackageVersion: '0.34.0'
+Installers:
+- Architecture: x64
+  InstallerLocale: en-US
+  InstallerType: zip
+  InstallerUrl: https://github.com/git-tfs/git-tfs/releases/download/v0.34.0/GitTfs-0.34.0.zip
+  InstallerSha256: a1524279ee37a9e49f9e08eeaec668fd5a7c5874838bc3aa4bf2c42065acc8d5
+  NestedInstallerType: portable
+  NestedInstallerFiles:
+  - RelativeFilePath: git-tfs.exe
+    PortableCommandAlias: git-tfs
+  UpgradeBehavior: uninstallPrevious
+  Commands:
+  - git-tfs
+  ReleaseDate: 2024-04-04
+ManifestType: installer
+ManifestVersion: 1.6.0
--- a/manifests/g/git-tfs/git-tfs/0.34.0/git-tfs.git-tfs.locale.en-US.yaml
+++ b/manifests/g/git-tfs/git-tfs/0.34.0/git-tfs.git-tfs.locale.en-US.yaml
@@ -1,0 +1,20 @@
+PackageIdentifier: git-tfs.git-tfs
+PackageVersion: '0.34.0'
+PackageLocale: en-US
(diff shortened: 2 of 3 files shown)


stephengillie:
Hi @pmiossec,

This package installed normally, but had a few issues running:
image

- First is a known issue where the package manager creates a symlink to the application, and while the application runs, the symlink does not. This is usually due to the "start in" folder being the symlink location instead of the application location, preventing the application from reaching files it needs.
- Second, this application gave an error when launched from its folder. Is this expected?

pmiossec:
@stephengillie Thanks for the feedback

>     * First is a known issue where the package manager creates a symlink to the application, and while the application runs, the symlink does not. This is usually due to the "start in" folder being the symlink location instead of the application location, preventing the application from reaching files it needs.

What do you recommend to solve the issue?
Removing the symlink is enough? Or there is a better solution?
I have not found a clear solution...

>     * Second, this application gave an error when launched from its folder. Is this expected? Edit: Should `Microsoft.VisualStudioCode` or similar be a dependency?

There is a dependency on "Microsoft Visual Studio" (not Code 😉 ).
But we can't enforce a dependency for a specific version because normally, the developer already has a version installed and it could be any version in 2015, 2017, 2019 or 2022. git-tfs successfully load the dependency dll depending on the version installed.
It could be handled a little more gracefully but I don't think we should block the merge for this reason (the behavior is like that for years and we never had a complain because it's a known dependency by the user)

stephengillie:
A bot is supposed to link this issue when the https://github.com/microsoft/winget-pkgs/labels/zip-binary label is applied: 
- https://github.com/microsoft/winget-cli/issues/2711

pmiossec:
> A bot is supposed to link this issue when the [ zip-binary](https://github.com/microsoft/winget-pkgs/labels/zip-binary) label is applied:

It did 😉 
That's why I said that "**I** have not found a clear solution" 😟
A lot of messages and talking about different subject (auto-unpack exe, bat file) bat without a message providing a clear answer of a solution to follow...

As I said, I think I will remove the symlink because I don't see an added value. Is it ok?

stephengillie:
> As I said, I think I will remove the symlink because I don't see an added value. Is it ok?

If it's possible to remove, I would be very interested to learn how.

pmiossec:
> > As I said, I think I will remove the symlink because I don't see an added value. Is it ok?
> 
> If it's possible to remove, I would be very interested in learning how.

Indeed. A previous attempt made me think that not adding `PortableCommandAlias:` line will not create the symlink but a quick try confirm your thought, the symlink is created. A dirty (working) workaround is to name it something like `PortableCommandAlias: _git-tfs_DONT_USE_THIS_ALIAS`...

pmiossec:
@stephengillie Do you think it's an acceptable (dirty) workaround?
If yes, I push it. Otherwise, I don't know how to go forward...

stephengillie:
Hi @pmiossec,

Sorry to miss this. Without the symlink alias, how will users find the application? Some of our users might be people who install random applications for a hobby, and some might install this.

Trenly:
@pmiossec - Support for zipped binaries has now been added and is available in the pipelines. This will require you to update your submission to ManifestVersion: 1.9.0 and set ArchiveBinariesDependOnPath: true in the installer node. To keep things clean, this PR is being closed. Please submit a new PR for the latest version of this package. Thank you for your contributions!

PR #154151, opened 2024-05-18
New package: Inkdrop.Inkdrop.Portable version 5.8.1

CoolPlayLin:
Checklist for Pull Requests

Manifests

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/154151)

### DIFF
--- a/manifests/i/Inkdrop/Inkdrop/Portable/5.8.1/Inkdrop.Inkdrop.Portable.installer.yaml
+++ b/manifests/i/Inkdrop/Inkdrop/Portable/5.8.1/Inkdrop.Inkdrop.Portable.installer.yaml
@@ -1,0 +1,16 @@
+# Created using wingetcreate 1.6.1.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+
+PackageIdentifier: Inkdrop.Inkdrop.Portable
+PackageVersion: 5.8.1
+InstallerType: zip
+NestedInstallerType: portable
+NestedInstallerFiles:
+- RelativeFilePath: Inkdrop.exe
+  PortableCommandAlias: inkdrop
+Installers:
+- Architecture: x64
+  InstallerUrl: https://d3ip0rje8grhnl.cloudfront.net/v5.8.1/Inkdrop-5.8.1-Windows.zip
+  InstallerSha256: E072BC48979B030721F3FE1C0154A45AD8BE96826E7DA7F56125F349E9B7D96B
+ManifestType: installer
+ManifestVersion: 1.6.0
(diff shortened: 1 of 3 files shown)


Trenly:
[Policy] Zip Binary

CoolPlayLin:
Why https://github.com/microsoft/winget-pkgs/labels/zip-binary ?

Trenly:
> Why [ zip-binary](https://github.com/microsoft/winget-pkgs/labels/zip-binary) Depends on library files that aren't available via symlink ?

image

CoolPlayLin:
But it can actually run normally

Trenly:
> But it can actually run normally

Not in all cases. If the user has Developer mode enabled, or runs the install as admin, then attempting to run inkdrop using the command line alias that gets added causes the paths to the DLLs to be unavailable.

og-mrk:
> > But it can actually run normally
> 
> Not in all cases. If the user has Developer mode enabled, or runs the install as admin, then attempting to run inkdrop using the command line alias that gets added causes the paths to the DLLs to be unavailable.

Kind of curious on this particular situation, if the CommandAlias won't work (that is, if the user installs the software as admin to make the Symbolic Link in the first place).. then can't we just remove the PortableCommandAlias field and partially resolve the zip-binary issue?

My apologies if the above suggestion/question might seem like a stupid idea, as I'm kind of new with this particular topic. Thanks in advance @Trenly 😄

Trenly:
> > > But it can actually run normally
> > 
> > 
> > Not in all cases. If the user has Developer mode enabled, or runs the install as admin, then attempting to run inkdrop using the command line alias that gets added causes the paths to the DLLs to be unavailable.
> 
> Kind of curious on this particular situation, if the CommandAlias won't work (that is, if the user installs the software as admin to make the Symbolic Link in the first place).. then can't we just remove the PortableCommandAlias field and partially resolve the zip-binary issue?
> 
> My apologies if the above suggestion/question might seem like a stupid idea, as I'm kind of new with this particular topic. Thanks in advance @Trenly 😄

No, removing the command alias doesn’t change the underlying behavior. WinGet will always attempt to add the program to PATH - if PortableCommandAlias isn’t set, then WinGet will just use the file name as the alias. And, since WinGet doesn’t add start menu entries or desktop icons for portable apps, the only ways to start the application are to use the command line, or navigate to the executable directly. So an admin install always results in a symlink, and a user install results in a symlink if developer mode is enabled, and if there are dll dependencies then the application won’t launch properly from the command line if a symlink is used.

Trenly:
@CoolPlayLin - Support for zipped binaries has now been added and is available in the pipelines. This will require you to update your submission to ManifestVersion: 1.9.0 and set ArchiveBinariesDependOnPath: true in the installer node. To keep things clean, this PR is being closed. Please submit a new PR for the latest version of this package. Thank you for your contributions!

PR #154319, opened 2024-05-20
Remove version: Fort.Firewall version 3.12.7

ghost:
Replaced by https://github.com/microsoft/winget-pkgs/pull/154152
---
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/154319)


Trenly (commented):
Is there a reason for removing this other than the fact that a new version is available? The releases are statically linked by version, so the download of the version still works

Trenly:
Thanks for the explanation!

Trenly:
As a note - as long as the package keeps the same identifier, you can usually add the new version and remove the old version in a single PR since it's considered a move and edit of a single manifest by the pipelines

Emi-Emi-Emi:
@Trely Ok Thanks for the note! I will do that from now. I wish I knew it sooner that I wouldn't break any check by doing that and then mess things up, I just didn't want to take the risk, so took the longer way. So it should be easier and better than having to create more PRs than necessary, especially when I see some versions being released after few hours of one version being already merged and all that.

Thank you.

Trenly (approved):

Issue #142854, opened 2024-03-07
Policy Service sometimes doesn’t enable automerge


Trenly:
Sometimes Auto-Merge isn’t being enabled and it isn’t clear why. The triggers are working for most PRs, but a few are getting into a state which require a manual merge. I expect this is an issue with the policy service itself either running into rate limits or internal errors

Trenly:
[Policy] Issue-Bug
[Policy] Area-External
[Policy] Needs-Attention
[Policy] Area-Bots

Trenly:
This is still in ongoing discussion with the GitOps team

R-Adrian:
Pull https://github.com/microsoft/winget-pkgs/pull/149544

The package for Dropbox version version 197.4.7571 seems also affected by this - the commit is apparently merged but not published.

![image](https://github.com/microsoft/winget-pkgs/assets/16319504/2554c57f-84cd-42e9-9deb-5c429e74a2ab)

Trenly:
> Pull #149544
> 
> The package for Dropbox version 197.4.7571 seems also affected by this - the commit is apparently merged but not published.

This has nothing to do with the policy service not enabling automerge

R-Adrian:
> 
> This has nothing to do with the policy service not enabling automerge

it's been 19+ hours already since the merge happened - shouldn't it have been picked up by winget cli already?

Trenly:
> > This has nothing to do with the policy service not enabling automerge
> 
> it's been 19+ hours already since the merge happened - shouldn't it have been picked up by winget cli already?

The publishing pipeline was paused for a deployment. The publish pipeline is completely separate from the policy service. 

You can see in the PR you linked that the Policy Service *did* enable the automerge
![image](https://github.com/microsoft/winget-pkgs/assets/12611259/a81997c5-fa83-4a24-aeee-f5f24ee3e7be)

Trenly:
.

Trenly:
.

Trenly:
.

Trenly:
For anyone watching this, it continues to be an issue

stephengillie:
> For anyone watching this, it continues to be an issue

It might look odd, with every linked issue being merged, but this is because it occurs on the merge step of the PR lifecycle, with the highly-utilized workaround of a manual merge being performed.

DuckDuckStudio:
Is that the same issue?
[...]

PR #157957, opened 2024-06-15
Added ALLUSERS=1 switch for Nushell machine-wide installation manifest

misaka4a92:
Checklist for Pull Requests

Manifests

---
closes https://github.com/microsoft/winget-pkgs/issues/157954
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/157957)

### DIFF
--- a/manifests/n/Nushell/Nushell/0.94.2/Nushell.Nushell.installer.yaml
+++ b/manifests/n/Nushell/Nushell/0.94.2/Nushell.Nushell.installer.yaml
@@ -5,6 +5,8 @@ PackageIdentifier: Nushell.Nushell
 PackageVersion: 0.94.2
 InstallerLocale: en-US
 InstallerType: wix
+InstallerSwitches:
+  Custom: ALLUSERS=1
 Scope: machine
 InstallModes:
 - interactive


stephengillie (approved):

stephengillie:
- https://github.com/microsoft/winget-pkgs/issues/142854

justinbarclay:
@stephengillie this commit seems to have [broken](https://github.com/nushell/nushell/issues/13281) the install process for nuget users. 

Before, we could just install nushell with `winget install nushell`, but now we need to do it from an elevated prompt.

stephengillie:
Hi @justinbarclay,

Before, would nushell install to `Program Files` or to `AppData`? I'm guessing that without any switches, the installer determines install location based on elevation and permissions - installilng into the `user` scope usually, and only installing into the `machine` scope if elevated. Is this accurate? 

The `ALLUSERS=1` switch should force the installer to install in `Program Files` or another machine-scoped location. This would match the `Scope: machine` entry on the subsequent line. 

One fix would be to remove both the switch and the `Scope`, which would let elevation again determine the install location. Another would be moving these to the installer level, and duplicating the current installers with `user` scope and switches - which would allow these to be selected at insatll time, and I believe default to `user` scope for install. 

Which fix would be preferred?

dander:
The current installer seems to support both per-user and per-machine, so the first solution seems preferable.

fdncred:
As one of the maintainers of nushell, seems to me like the second option is better. Thanks for the input. To follow along further, please see https://github.com/nushell/nushell/issues/13281

PR #161285, opened 2024-07-03
New version: SomePythonThings.WingetUIStore version 3.1.0

marticliment:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/161285)


marticliment:
Hi, it is the same issue as always. I thought I had fix it but what I changed was not the cause

marticliment:
It fails

marticliment:
> @marticliment, check the review and add the line

Excuse me, which line?

vikingnope:
> > @marticliment, check the review and add the line
> 
> Excuse me, which line?

Oh ye sometimes it does not show, here it is

```suggestion
InstallerSuccessCodes:
- 1
```

Add this just above `Upgrade Behaviour`

stephengillie:
The package manager is still using the dependency's exit code instead of the package's, it would appar.

stephengillie (approved):

PR #162317, opened 2024-07-10
Update version: GitButler.GitButler version 0.12.9

damn-good-b0t:
Checklist for Pull Requests

Manifests

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/162317)

### DIFF
--- a/manifests/g/GitButler/GitButler/0.12.9/GitButler.GitButler.locale.en-US.yaml
+++ b/manifests/g/GitButler/GitButler/0.12.9/GitButler.GitButler.locale.en-US.yaml
@@ -10,9 +10,9 @@ PrivacyUrl: https://app.termly.io/policy-viewer/policy.html?policyUUID=a001c8b7-
 PackageName: GitButler
 PackageUrl: https://gitbutler.com
 License: Functional Source License, Version 1.1, MIT Future License
-LicenseUrl: https://github.com/gitbutlerapp/gitbutler/blob/7d01a537d49e0219773dacb1655f6a7a58f5f355/LICENSE.md
+LicenseUrl: https://github.com/gitbutlerapp/gitbutler/blob/master/LICENSE.md
 Copyright: Copyright 2023-2024 GitButler Inc
-CopyrightUrl: https://github.com/gitbutlerapp/gitbutler/blob/7d01a537d49e0219773dacb1655f6a7a58f5f355/LICENSE.md
+CopyrightUrl: https://github.com/gitbutlerapp/gitbutler/blob/master/LICENSE.md
 ShortDescription: A Git client for simultaneous branches on top of your existing workflow.
 Tags:
 - git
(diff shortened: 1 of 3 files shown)


manifests/g/GitButler/GitButler/0.12.9/GitButler.GitButler.locale.en-US.yaml
```diff
@@ -0,0 +1,23 @@
...
+PackageIdentifier: GitButler.GitButler
+PackageVersion: 0.12.9
+PackageLocale: en-US
+Publisher: GitButler
+PublisherUrl: https://gitbutler.com
+PrivacyUrl: https://app.termly.io/policy-viewer/policy.html?policyUUID=a001c8b7-505b-4eab-81e3-fcd1c72bdd79
+PackageName: GitButler
+PackageUrl: https://gitbutler.com
+License: Functional Source License, Version 1.1, MIT Future License
+LicenseUrl: https://github.com/gitbutlerapp/gitbutler/blob/7d01a537d49e0219773dacb1655f6a7a58f5f355/LICENSE.md
```
mdanish-kh, line 13:
```suggestion
LicenseUrl: https://github.com/gitbutlerapp/gitbutler/blob/master/LICENSE.md
```

manifests/g/GitButler/GitButler/0.12.9/GitButler.GitButler.locale.en-US.yaml
```diff
@@ -0,0 +1,23 @@
...
+Copyright: Copyright 2023-2024 GitButler Inc
+CopyrightUrl: https://github.com/gitbutlerapp/gitbutler/blob/7d01a537d49e0219773dacb1655f6a7a58f5f355/LICENSE.md
```
mdanish-kh, line 15:
```suggestion
CopyrightUrl: https://github.com/gitbutlerapp/gitbutler/blob/master/LICENSE.md
```

mdanish-kh (changes requested):
It's better to use master branch URLs instead of specific commits, so users see the latest file contents when they visit this URL

a-mnich:
> It's better to use master branch URLs instead of specific commits, so users see the latest file contents when they visit this URL

@mdanish-kh  Yes, but that's only true for files where you actually want them to be up-to-date. For a specific version of a package, usually the license that was in the repository at that point in time applies. When the package is re-licensed in a later version, the old license still applies and therefore I think the manifest should point to a specific commit.

mdanish-kh:
@a-mnich Generally package updates here are made through tools that automatically generate the manifest for you (wingetcreate as an example). These tools generally don't have the ability to dynamically fetch the latest upstream commit's LicenseUrl for every new update. What you mention is desirable, and in that case, the `LicenseUrl` here should actually be `https://github.com/gitbutlerapp/gitbutler/blob/release/0.12.9/LICENSE.md` instead. The URL in the PR is the same carried over from previous manifests, which is exactly my point that the tools available don't allow you to easily change this URL every update. It is better to just go with a static latest URL, to add less friction in the contribution / PR generation process.

If the license changes, one can easily write a script to go through the older manifests and update `LicenseUrl` to some commit / release branch before the license change was made. However, license changes are generally very rare.

mdanish-kh (approved):

PR #163593, opened 2024-07-18
Opam: Use opam binaries directly from github

ManasJayanth:
Checklist for Pull Requests

Manifests

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/163593)

### DIFF
--- a/manifests/o/OCaml/opam/2.2.0/OCaml.opam.installer.yaml
+++ b/manifests/o/OCaml/opam/2.2.0/OCaml.opam.installer.yaml
@@ -16,7 +16,5 @@ Installers:
 - Architecture: x64
   InstallerUrl: https://github.com/ocaml/opam/releases/download/2.2.0/opam-2.2.0-x86_64-windows.exe
   InstallerSha256: DB8401B2190DF94920494972D9336AB6ECC906E4720A527EB0E97F814B0EA813
-  Commands:
-  - opam
 ManifestType: installer
 ManifestVersion: 1.6.0
--- a/manifests/o/OCaml/opam/2.2.0/OCaml.opam.installer.yaml
+++ b/manifests/o/OCaml/opam/2.2.0/OCaml.opam.installer.yaml
@@ -1,0 +1,23 @@
+# Created using wingetcreate 1.6.1.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+
+PackageIdentifier: OCaml.opam
+PackageVersion: 2.2.0
+Platform:
+- Windows.Desktop
+MinimumOSVersion: 10.0.18362.0
+InstallModes:
+- silent
+- interactive
+UpgradeBehavior: install
+Commands:
+- opam
+InstallerType: portable
+Installers:
+- Architecture: x64
+  InstallerUrl: https://github.com/ocaml/opam/releases/download/2.2.0/opam-2.2.0-x86_64-windows.exe
+  InstallerSha256: DB8401B2190DF94920494972D9336AB6ECC906E4720A527EB0E97F814B0EA813
+  Commands:
+  - opam
+ManifestType: installer
+ManifestVersion: 1.6.0
--- a/manifests/o/OCaml/opam/2.2.0/OCaml.opam.installer.yaml
+++ b/manifests/o/OCaml/opam/2.2.0/OCaml.opam.installer.yaml
@@ -8,7 +8,6 @@ Platform:
 MinimumOSVersion: 10.0.18362.0
 InstallModes:
 - silent
-- interactive
 UpgradeBehavior: install
 Commands:
 - opam
--- a/manifests/o/OCaml/opam/2.2.0/OCaml.opam.locale.en-US.yaml
+++ b/manifests/o/OCaml/opam/2.2.0/OCaml.opam.locale.en-US.yaml
@@ -1,0 +1,27 @@
+# Created using wingetcreate 1.6.1.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.6.0.schema.json
+
+PackageIdentifier: OCaml.opam
+PackageVersion: 2.2.0
+PackageLocale: en-US
+Publisher: OCaml, Inc.
+PublisherUrl: https://www.ocaml.org/
+PublisherSupportUrl: https://github.com/ocaml/opam/issues
+PrivacyUrl: https://ocaml.org/policies/privacy-policy
+PackageName: opam
+PackageUrl: https://opam.ocaml.org/
+License: LGPL-2.1-only WITH OCaml-LGPL-linking-exception
+LicenseUrl: https://github.com/ocaml/opam/blob/master/LICENSE
+Copyright: Copyright 2012-2024 the opam authors
+CopyrightUrl: https://github.com/ocaml/opam/blob/master/README.md#copyright-and-license
+ShortDescription: A package manager for OCaml.
+Description: Opam is a source-based package manager for OCaml. It supports multiple simultaneous compiler installations, flexible package constraints, and a Git-friendly development workflow.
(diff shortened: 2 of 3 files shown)


ManasJayanth:
@microsoft-github-policy-service agree [company="Dining Philosophers, LLP."]

ManasJayanth:
@microsoft-github-policy-service agree company="Dining Philosophers, LLP."

mdanish-kh (changes requested):
@ManasJayanth Would you help me understand the change here? Why are we removing the existing EXE installers? Because of the installer type difference, this will break the upgrade flow for users that already have the EXE flavor installed. WinGet won't be able to upgrade from an existing `exe` installation to a new `portable` one

Additionally, the package ID for this package is `Diskuv.opam` and the URLs come from repo `diskuv/[...]`. I'm not aware if this is the official channel for the distribution of the package. The new URL however, comes from the repo `ocaml/opam`. Seems to me that this change belongs in a new package ID `OCaml.Opam`

ManasJayanth:
> Additionally, the package ID for this package is Diskuv.opam and the URLs come from repo diskuv/[...]. I'm not aware if this is the official channel for the distribution of the package. The new URL however, comes from the repo ocaml/opam. Seems to me that this change belongs in a new package ID OCaml.Opam

Indeed. I just spoke with one of the maintainers and they're want this too.

> Would you help me understand the change here? 

This change was suggested by the current maintainer of the manifest - opam no longer needs an installer per say and a zip file extract is good enough (reducing points of failure)

> Because of the installer type difference, this will break the upgrade flow for users that already have the EXE flavor installed. WinGet won't be able to upgrade from an existing exe installation to a new portable one

Thanks for pointing this out. I didn't realise this. But, I guess, once a new manifest is created under the `OCaml` namespace, this wouldn't be an issue. Feel free to correct me if I'm wrong.

mdanish-kh:
> But, I guess, once a new manifest is created under the OCaml namespace, this wouldn't be an issue. Feel free to correct me if I'm wrong.

If you add the manifest in `OCaml.Opam` and set `Publisher: OCaml` (current value is `Diskuv, Inc.`) in the locale manifest, WinGet _should_ treat both packages as separate. The `exe` installs would get mapped to PackageId `Diskuv.opam` and wouldn't get the portable offered as an upgrade. Users will have to install through the new PackageId to get the portable version. The `PackageName` value being `opam` in both _may_ cause collision when WinGet tries to correlate installed package information to available manifests in this repository, but I'm hoping we wouldn't face the issue as `Publisher` value would be different

> This change was suggested by the current maintainer of the manifest - opam no longer needs an installer per se and a zip file extraction is good enough (reducing points of failure)

Btw, if the EXE installer is no longer needed, would it be a good idea to remove `Diskuv.opam` manifests entirely from the repo? If we want new users to only discover and install portable version of the package, removing the manifest would make it so that `winget search/install opam` would only ever map to PackageId `OCaml.Opam`

ManasJayanth:
@jonahbeckford Would love to hear your thoughts on removing `Diskuv.opam` from the registry.

jonahbeckford:
> @jonahbeckford Would love to hear your thoughts on removing `Diskuv.opam` from the registry.

* In our instructions for installing opam, we should always specify the Publisher.Package for minimal security. That is, `winget install --id OCaml.opam`. So just as a forcing function it may be best to keep `Diskuv.opam` around until the ocaml.org docs change. After that we can do the removal.
* I've always included `opam-installer.exe` since a) it is a trivial thing to build, b) it is documented in the official opam manual, and c) I've used `opam-installer.exe` before and probably will again. But we don't distribute `opam-installer` in the Unix package managers (ex. https://rpmfind.net/linux/RPM/fedora/devel/rawhide/x86_64/o/opam-2.1.6-3.fc41.x86_64.html) so yeah, it can be dropped.
* I can understand dropping `opam-putenv.exe` if 32-bit needs to be dropped.

---

For now, let's get a `OCaml.opam` version `2.0.0` available with just 64-bit `opam.exe` inside it.

Note: I very much don't like the changing capitalization of `OCaml.opam`, but I think `opam` is officially lowercase (confer: https://discuss.ocaml.org/t/capitalization-of-opam/1686) and `OCaml` has always been mixed-case.

manifests/o/OCaml/opam/2.2.0/OCaml.opam.installer.yaml
```diff
@@ -0,0 +1,23 @@
...
+UpgradeBehavior: install
+Commands:
+- opam
+InstallerType: portable
+Installers:
+- Architecture: x64
+  InstallerUrl: https://github.com/ocaml/opam/releases/download/2.2.0/opam-2.2.0-x86_64-windows.exe
+  InstallerSha256: DB8401B2190DF94920494972D9336AB6ECC906E4720A527EB0E97F814B0EA813
+  Commands:
+  - opam
```
mdanish-kh, line 21:
```suggestion
```
Not needed here as it's specified at the root level at L13, L14

manifests/o/OCaml/opam/2.2.0/OCaml.opam.installer.yaml
```diff
@@ -0,0 +1,23 @@
...
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+
+PackageIdentifier: OCaml.opam
+PackageVersion: 2.2.0
+Platform:
+- Windows.Desktop
+MinimumOSVersion: 10.0.18362.0
+InstallModes:
+- silent
+- interactive
```
mdanish-kh, line 11:
```suggestion
```
Portables don't technically support an "interactive" installation mode

manifests/o/OCaml/opam/2.2.0/OCaml.opam.locale.en-US.yaml
```diff
@@ -0,0 +1,27 @@
...
+Copyright: Copyright 2012-2024 the opam authors
+CopyrightUrl: https://github.com/ocaml/opam/blob/master/README.md#copyright-and-license
+ShortDescription: A package manager for OCaml.
+Description: Opam is a source-based package manager for OCaml. It supports multiple simultaneous compiler installations, flexible package constraints, and a Git-friendly development workflow.
+Tags:
+- foss
+- ocaml
+- open-source
+- package-manager
+- dependency-manager
```
mdanish-kh, line 24:
```suggestion
- dependency-manager
Moniker: ocaml
```
Adding `Moniker: opam` _should_ tell the client to install `OCaml.opam` when a user types `winget install opam` provided no other PackageId in the repo shares this moniker

manifests/o/OCaml/opam/2.2.0/OCaml.opam.locale.en-US.yaml
```diff
@@ -0,0 +1,27 @@
...
+InstallationNotes: See https://opam.ocaml.org/blog for the latest news on how to use opam.
```
mdanish-kh, line 25:
```suggestion
InstallationNotes: See https://opam.ocaml.org/blog for the latest news on how to use opam.
Documentations:
  - DocumentLabel: Opam 2.0 documentation
    DocumentUrl: https://opam.ocaml.org/doc/
```

Since there were other requested changes, figured I would optionally suggest this as well for a more complete manifest

mdanish-kh (changes requested):

ManasJayanth:
@mdanish-kh Thank you for helpful review comments. Really sped things up!

mdanish-kh:
[Policy] Reset Labels

mdanish-kh (approved):

mdanish-kh:
[Policy] Needs-Manual-Merge

stephengillie:
- https://github.com/microsoft/winget-pkgs/issues/142854

PR #164604, opened 2024-07-25
New version: Neovim.Neovim version 0.10.1

damn-good-b0t:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/164604)

### DIFF
--- a/manifests/n/Neovim/Neovim/0.10.1/Neovim.Neovim.installer.yaml
+++ b/manifests/n/Neovim/Neovim/0.10.1/Neovim.Neovim.installer.yaml
@@ -1,0 +1,21 @@
+# Created with WinGet Updater using komac v2.3.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+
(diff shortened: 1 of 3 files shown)


stephengillie (approved):

Trenly:
@stephengillie - You may need to update your automation to check for scope changes;

* #164944

stephengillie:
> @stephengillie - You may want to update your automation to check for scope changes from the previous manifest
> 
> * [[Package Issue]: Neovim.Neovim #164944](https://github.com/microsoft/winget-pkgs/issues/164944)
> 
> Edit: You should also be checking that if scope is specified that the uninstall registry entry matches the specified scope - HKLM for machine and HKCU for user. It's a known bug with wix installers that they always write to HKLM even when a user scope install was performed, so `machine` or unset are the only "correct" scopes for wix installers

This should probably be an Issue with the Issue-Feature tag. It's outside of the scope of my current assigned workload.

PR #165016, opened 2024-07-28
New version: SomePythonThings.WingetUIStore version 3.1.1

marticliment:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/165016)

### DIFF
--- a/manifests/s/SomePythonThings/WingetUIStore/3.1.1/SomePythonThings.WingetUIStore.installer.yaml
+++ b/manifests/s/SomePythonThings/WingetUIStore/3.1.1/SomePythonThings.WingetUIStore.installer.yaml
@@ -1,0 +1,25 @@
+# Created with WinGet Releaser v2 using Komac v1.11.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.5.0.schema.json
+
+PackageIdentifier: SomePythonThings.WingetUIStore
+PackageVersion: 3.1.1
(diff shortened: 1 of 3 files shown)


ClintonOffor:
Ok

marticliment:
Is there no way to retrieve logs from the testing pipelines?

vikingnope:
I would recommend removing the `/NoWinGet` flag. That is the only difference between the pre-release manifest and this one, and the pre-release was working fine during pipelines.

404NetworkError:
I don't know if removing the flag will fix it. Looking at other PRs, all of them since ~9 hrs ago are failing the same way.

marticliment:
> I would recommend removing the `/NoWinGet` flag. That is the only difference between the pre-release manifest and this one, and the pre-release was working fine during pipelines.

The Prerelease did pass? Then this release should also pass. Changes were introduced that shouldn't fail verification amymore. I have tested the pr both with the flag and without the flag, and it failed both of the times. However, if the 3.1.1-beta3 passed, this PR should also pass the checks, since no changes have been introduced since then to the installer

marticliment:
Yes, all new PRs are failing the tests

stephengillie (approved):

stephengillie (approved):

stephengillie:
This package installs and launches normally in a Windows 10 VM.

PR #172068, opened 2024-09-06
New package: SolarWinds.KiwiSyslogServer version 9.8.3

rakleed:
Checklist for Pull Requests

Manifests

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/172068)

### DIFF
--- a/manifests/s/SolarWinds/KiwiSyslogServer/9.8.3/SolarWinds.KiwiSyslogServer.installer.yaml
+++ b/manifests/s/SolarWinds/KiwiSyslogServer/9.8.3/SolarWinds.KiwiSyslogServer.installer.yaml
@@ -11,11 +11,15 @@ Installers:
 - Architecture: x86
   InstallerUrl: https://www.solarwinds.com/downloads/SyslogServerFree.zip
   InstallerSha256: 1FCE04035F51220B2A9D2C7481479E74E2598CBCBB5F8560C9DC1AB611474E39
+  InstallerSwitches:
+    Custom: /CURRENTUSER
   AppsAndFeaturesEntries:
   - DisplayVersion: '9.8.3  (Service Edition)'
 - Architecture: x86
   InstallerUrl: https://www.solarwinds.com/downloads/SyslogServerFree.zip
   InstallerSha256: 1FCE04035F51220B2A9D2C7481479E74E2598CBCBB5F8560C9DC1AB611474E39
+  InstallerSwitches:
+    Custom: /CURRENTUSER
   AppsAndFeaturesEntries:
   - DisplayVersion: '9.8.3  (Standard Edition)'
 ManifestType: installer
--- a/manifests/s/SolarWinds/KiwiSyslogServer/9.8.3/SolarWinds.KiwiSyslogServer.installer.yaml
+++ b/manifests/s/SolarWinds/KiwiSyslogServer/9.8.3/SolarWinds.KiwiSyslogServer.installer.yaml
@@ -20,7 +20,7 @@ Installers:
   InstallerUrl: https://www.solarwinds.com/downloads/SyslogServerFree.zip
   InstallerSha256: 1FCE04035F51220B2A9D2C7481479E74E2598CBCBB5F8560C9DC1AB611474E39
   InstallerSwitches:
-    Custom: /CURRENTUSER
+    Custom: /ALLUSERS
   AppsAndFeaturesEntries:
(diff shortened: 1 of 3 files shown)


stephengillie:
Verify:
image

Hi @rakleed,

It looks like the version number in Registry includes the "Service Edition", to be `9.8.3 (Service Edition)`. Either this needs a `DisplayVersion` to help with matching, or the `PackageVersion` should be updated to include "Service Edition". What do you think?

stephengillie:
> @ stephengillie It would probably be better to specify it in `DisplayVersion`, since otherwise you would probably need to change the folder name, which would look strange.

While it would be an odd folder name, it should reflect what's needed for program operation. 

> And if you run the installer manually, there will be an option to install in application mode, not as a service by default, and then the version number will have a different value -- `9.8.3 (Standard Edition)`.

Good point. 

> As far as I understand, `DisplayVersion` can only contain a string, you do not support regex.

This is also correct.

> If you specify `DisplayVersion: "9.8.3 (Service Edition)"`, could this cause problems with `winget update` if the user installs it as an application?

The `DisplayVersion` would have to be nested under each installer. This way, the appropriate `DisplayVersion` could be specified for both.

stephengillie:
>  Manifest Error: Duplicate installer entry found.

stephengillie:
Hi @rakleed,

I believe the `Custom` value is the one to populate here. This should be added in each `Installers` node, to specify the node-specific switch data. I like to put data such as this in the line below the `InstallerSha256`, but I believe it would be valid anywhere in the node.

```
  InstallerSwitches:
    Custom: /CURRENTUSER
```

manifests/s/SolarWinds/KiwiSyslogServer/9.8.3/SolarWinds.KiwiSyslogServer.installer.yaml
```diff
@@ -0,0 +1,26 @@
...
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+
+PackageIdentifier: SolarWinds.KiwiSyslogServer
+PackageVersion: 9.8.3
+InstallerType: zip
+NestedInstallerType: nullsoft
+NestedInstallerFiles:
+- RelativeFilePath: Kiwi_Syslog_Server_9.8.3.Freeware.setup.exe
+Installers:
+- Architecture: x86
```
stephengillie, line 11:
```suggestion
- Architecture: x86
  Scope: user
```

manifests/s/SolarWinds/KiwiSyslogServer/9.8.3/SolarWinds.KiwiSyslogServer.installer.yaml
```diff
@@ -0,0 +1,26 @@
...
+  InstallerUrl: https://www.solarwinds.com/downloads/SyslogServerFree.zip
+  InstallerSha256: 1FCE04035F51220B2A9D2C7481479E74E2598CBCBB5F8560C9DC1AB611474E39
+  InstallerSwitches:
+    Custom: /CURRENTUSER
+  AppsAndFeaturesEntries:
+  - DisplayVersion: '9.8.3  (Service Edition)'
+- Architecture: x86
```
stephengillie, line 18:
```suggestion
- Architecture: x86
  Scope: machine
```

manifests/s/SolarWinds/KiwiSyslogServer/9.8.3/SolarWinds.KiwiSyslogServer.installer.yaml
```diff
@@ -0,0 +1,26 @@
...
+  InstallerUrl: https://www.solarwinds.com/downloads/SyslogServerFree.zip
+  InstallerSha256: 1FCE04035F51220B2A9D2C7481479E74E2598CBCBB5F8560C9DC1AB611474E39
+  InstallerSwitches:
+    Custom: /CURRENTUSER
```
stephengillie, line 22:
```suggestion
    Custom: /ALLUSERS
```

stephengillie (changes requested):
The `Scope` data will differentiate the nodes, so the pipelines can progress past this message. 

For the installer switches, I'm using an installer's standard switches (I forget which) - and there's a chance that the switches will need to be modified.

stephengillie (approved):

Issue #173833, opened 2024-09-16
[Package Issue]: WinDirStat.WinDirStat - WinGet Forgets Installed Version of WinDirStat.WinDirStat After Upgrade, Causing Repeated Upgrades


Lonniebiz:
### Please confirm these before moving forward

- [X] I have searched for my issue and not found a work-in-progress/duplicate/resolved issue.
- [X] I have not been informed if the issue is resolved in a preview version of the winget client.

### Category of the issue

Other

### Brief description of your issue

WinGet forgets the installed version of WinDirStat.WinDirStat immediately after upgrading, leading to unnecessary repeated upgrades in subsequent runs.

### Steps to reproduce

1. Open PowerShell
2. Run `winget install windirstat --scope machine` to install WinDirStat if not already installed
3. After installation, run `winget upgrade --all --include-unknown`
4. Observe that WinDirStat.WinDirStat shows "Unknown" in the Version column, but has an "Available" version listed
5. Allow the upgrade to proceed and complete successfully
6. Immediately run `winget upgrade --all --include-unknown` again
7. Notice that WinDirStat.WinDirStat still shows as "Unknown" in the Version column and is again offered for upgrade to the same version that was just installed

### Actual behavior

Winget successfully upgrades WinDirStat.WinDirStat to the latest available version. However, it immediately "forgets" this version information. In subsequent runs, even immediately after an upgrade, the Version column remains "Unknown" for WinDirStat.WinDirStat. This results in winget repeatedly offering to upgrade WinDirStat.WinDirStat to the same version in every run, creating a loop of unnecessary upgrades.

### Expected behavior

After a successful upgrade, winget should remember and correctly display the installed version of WinDirStat.WinDirStat. Subsequent `winget upgrade` commands should not offer to upgrade WinDirStat.WinDirStat unless a newer version is actually available. The Version column should show the correct installed version, and WinDirStat.WinDirStat should only be listed for upgrade when there's a genuine new version available.

### Environment

```raw
Windows Package Manager v1.8.1911
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.19045.4894
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.23.1911.0

Winget Directories
-----------------------------------------------------------------------------------------------------------------------
Logs                               %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\Diag…
[...]

Issue #161769, opened 2024-07-06
[Update Request]: WinDirStat/WinDirStatt improperly detects v1.1.2 as "Unknown" (Installed Version:  Unknown - Update to 1.1.2 available)


ntman1:
### What type of update are you requesting?

A change to the package metadata

### Current Package Identifier

WinDirStat.WinDirStat

### Package Version

1.1.2

### Please describe the changes you would like to see

The version of WinDirStat.WinDirStat located at https://github.com/microsoft/winget-pkgs/tree/master/manifests/w/WinDirStat/WinDirStat fails to detect either manual or WinGet versions of WinDirStat v1.1.2 as being WinDirStat v1.1.2, and misidentifies the WinDirStat v1.1.21 installation as "Unknown" (aka Installed Version:  Unknown - Update to 1.1.2 available) I am assuming based upon the manifest that it is installing the current package from https://sourceforge.net/projects/windirstat/files/windirstat/1.1.2%20installer%20re-release%20%28more%20languages%21%29/windirstat1_1_2_setup.exe/download, however when WinGet sees the package (eventhough it's showing properly in Control Panel -> Programs and Features or Settings -> Apps -> Installed Apps as v1.1.2 (however the original author's packaging may not be properly inserting the version number into the registry).

ntman1:
I put in a bug report to the developers regarding their installer - [https://github.com/windirstat/windirstat/issues/75](url)

Trenly:
* Duplicate of #54490 
* Duplicate of #88994

Issue #88994, opened 2022-02-15
'winget upgrade' reports WinDirStat upgradable when latest version is installed


Hobart:
### Brief description of your issue

I have WinDirStat 1.1.2 installed.  `winget upgrade` lists it as upgradable.

### Steps to reproduce

```
C:\Windows\system32>winget upgrade
Name                Id                    Version Available Source
------------------------------------------------------------------
Mozilla Firefox ESR Mozilla.Firefox.ESR   45.9.0  91.6.0    winget
WinDirStat          WinDirStat.WinDirStat Unknown 1.1.2     winget
2 upgrades available.

C:\Windows\system32>
```

### Expected behavior

Not see an up to date product listed.

### Actual behavior

Product is listed with installed 'Version' of 'Unknown'

### Environment

```shell
C:\Windows\system32>winget --info
Windows Package Manager v1.1.13405
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.19044.1526
Package: Microsoft.DesktopAppInstaller v1.16.13405.0

Logs: %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir

Links
---------------------------------------------------------------------------
Privacy Statement   https://aka.ms/winget-privacy
License Agreement   https://aka.ms/winget-license
Third Party Notices https://aka.ms/winget-3rdPartyNotice
Homepage            https://aka.ms/winget
Windows Store Terms https://www.microsoft.com/en-us/storedocs/terms-of-sale
```

```
reg query hkcu\software\microsoft\windows\currentversion\uninstall\windirstat

HKEY_CURRENT_USER\software\microsoft\windows\currentversion\uninstall\windirstat
    UninstallString    REG_EXPAND_SZ    "C:\Program Files (x86)\WinDirStat\Uninstall.exe"
    InstallLocation    REG_EXPAND_SZ    C:\Program Files (x86)\WinDirStat
    DisplayName    REG_SZ    WinDirStat 1.1.2
    DisplayIcon    REG_SZ    C:\Program Files (x86)\WinDirStat\windirstat.exe,0
    dwVersionMajor    REG_DWORD    0x1
    dwVersionMinor    REG_DWORD    0x1
    dwVersionRev    REG_DWORD    0x2
    dwVersionBuild    REG_DWORD    0x4f
    URLInfoAbout    REG_SZ    http://windirstat.info/
[...]

Issue #54490, opened 2022-03-15
[Package Issue]: WinDirStat.WinDirStat


fghamsary:
### Please confirm these before moving forward

- [X] I have searched for my issue and not found a work-in-progress/duplicate/resolved issue.
- [X] I have not been informed if the issue is resolved in a preview version of the winget client.

### Category of the issue

Other

### Brief description of your issue

I have already had this application installed before, and now winget propose an update for it, but can't detect the current version installed, but proposes an update (a phantom update, because I am on the same version)
I've already uninstalled it, and installed it again via winget, always the same.

### Steps to reproduce

Just install WinDirStat and afterwards check with a winget upgrade.
You can see that winget can't determine the currently installed version, so it proposes the update which doesn't work as well.
May be it has something to do with the language of my windows which is in French.

### Actual behavior

Can't determine the currently installed version so proposes the update always!

### Expected behavior

Determine the current version correctly and not proposing any update for it.

### Environment

```shell
PS C:\Users\fghamsary> winget --info
Windows Package Manager (Preview) v1.2.3411-preview
Copyright (c) Microsoft Corporation. Tous droits réservés.

Windows: Windows.Desktop v10.0.22000.556
Package: Microsoft.DesktopAppInstaller v1.17.3411.0
```

### Screenshots and Logs

PS C:\Users\fghamsary> winget install WinDirStat.WinDirStat
Trouvé WinDirStat [WinDirStat.WinDirStat] Version 1.1.2
La licence d’utilisation de cette application vous est octroyée par son propriétaire.
Microsoft n’est pas responsable des paquets tiers et n’accorde pas de licences à ceux-ci.
Downloading https://sourceforge.net/projects/windirstat/files/windirstat/1.1.2%20installer%20re-release%20%28more%20languages%21%29/windirstat1_1_2_setup.exe/download
  ██████████████████████████████   630 KB /  630 KB
Le code de hachage de l’installation a été vérifié avec succès
Démarrage du package d’installation... Merci de patienter.
Installé correctement (This means installed correctly => it is in French)

PS C:\Users\fghamsary> winget upgrade
Nom                ID                    Version       Disponible    Source
---------------------------------------------------------------------------
WinDirStat         WinDirStat.WinDirStat Unknown       1.1.2         winget
Microsoft .NET SDK Microsoft.dotnet      5.4.622.11901 6.1.322.12413 winget

jedieaston:
[...]

PR #174213, opened 2024-09-19
Add recommendation for users to move away from WinDirStat

Trenly:
Checklist for Pull Requests

Manifests

---

There have now been a few issues filed indicating confusion around the behavior of WinDirStat and the version being unknown.
* https://github.com/microsoft/winget-pkgs/issues/161769
* https://github.com/microsoft/winget-pkgs/issues/54490
* https://github.com/microsoft/winget-pkgs/issues/88994
* https://github.com/microsoft/winget-pkgs/issues/173833

Because the last release of WinDirStat was in 2016, it should be safe to assume that it is not in active development and no new updates for the package will be coming. However, another package - WizTree - provides the same functionality as WinDirStat but is newer, much faster, and is still being patched and developed upon. 

This PR adds an installation note to help reduce confusion and point users towards the newer software if they should choose to install it.

###### Related to microsoft/winget-cli#1899

### DIFF
--- a/manifests/w/WinDirStat/WinDirStat/1.1.2/WinDirStat.WinDirStat.installer.yaml
+++ b/manifests/w/WinDirStat/WinDirStat/1.1.2/WinDirStat.WinDirStat.installer.yaml
@@ -1,5 +1,5 @@
-# Created with YamlCreate.ps1 v2.0.7 $debug=AUSU.7-2-1
-# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.1.0.schema.json
+# Created with YamlCreate.ps1 v2.4.1 $debug=AUSU.CRLF.7-4-5.Win32NT
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
 
 PackageIdentifier: WinDirStat.WinDirStat
 PackageVersion: 1.1.2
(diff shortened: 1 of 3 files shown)


ItzLevvie (approved):

Lonniebiz:
> Because the last release of WinDirStat was in 2016, it should be safe to assume that it is not in active development and no new updates for the package will be coming. However, another package - WizTree - provides the same functionality as WinDirStat but is newer, much faster, and is still being patched and developed upon.
> 
> This PR adds an installation note to help reduce confusion and point users towards the newer software if they should choose to install it.

While WizTree may be faster, and more actively developed, it isn't open source (like WinDirStat is). Users like myself may prefer WinDirStat for this reason alone (even if it is slower).

Trenly:
That is true, which is why I just added it as a note so that users can make their own decisions about the software they choose to install

cvrebert:
> it should be safe to assume that it is not in active development and no new updates for the package will be coming

They've been making alpha/beta releases periodically this year (https://github.com/windirstat/windirstat-next/releases ). Likewise for commits (https://github.com/windirstat/windirstat/commits/master/ ), including one yesterday.

> There have now been a few issues filed indicating confusion around the behavior of WinDirStat and the version being unknown.

Presumably they'll have that fixed when they finally make their next major stable release: https://github.com/windirstat/windirstat/issues/46
Seems they're switching to a Wix installer.

Trenly:
Perfect; Once a new version is released I'll update the manifest and remove this note

NoMoreFood:
@Trenly The next two months is my goal. Would you like me to take over updating this when the release is ready?  I would appreciate you removing the note now since it is false.

Lonniebiz:
@NoMoreFood: Couldn't you do a quick release of WinDirStat, that is exactly like the current version provided in winget except that it merely sets that version registry entry that winget needs to remain aware of the installed version of WinDirstat?

Perhaps that alone is more work than it might appear me, and it is just better to wait.

As for WinDirStat and "active development", I've used it for years and it doesn't bother me much that it hasn't changed since 2016. When something works well for its purpose it doesn't require too much active development. Plus, it is open source, and I typically favor that over closed source alternatives (that's more of a priority to me than performance in this case).

While I'm glad to be informed of the existence WizTree, I don't like a package manager recommending a proprietary application as an alternative to an open source one unless this difference is explicitly made clear. I guess that's just my preference, but now that the note has been discovered untrue, that's probably a more agreeable reason to go ahead and remove it.

aisgbnok:
FYI, @Trenly improved the accuracy of the note last week:

- #174360

As I understand it, the note aimed to minimize user confusion and reduce the amount of issues being opened in this [microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs) repository. While I agree that the previous note may have seemed biased, the updated note appears to be a reasonable compromise until 2.x.x lands.

Note as of 2024-09-27:

> **Installation Notes:** Because this software does not write a version to the registry, WinGet is not able to determine the installed version. This will be fixed in the next release of WinDirStat. See https://github.com/windirstat/windirstat/issues/75 for more info.

PR #175251, opened 2024-09-26
WoowaBrothersCorp.BaeminOrder version `(0.8.2880.0)`

alexkoala:
Checklist for Pull Requests

Manifests

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/175251)


stephengillie:
Verify:
image
image

Hi @alexkoala,

The installer adds parentheses around the version number that the package and manifest omit. Should these be added to the `PackageVersion`, or should a `DisplayVersion` be used? What's your opinion?

stephengillie:
> @stephengillie I think it would be better to follow `DisplayVersion` for ARP match. What's your opinion?

At first, I agree - it would be the simplest way to make the `PackageIdentifier` match the application version while still matching to the Registry.

But the `DIsplayVersion` can cause numerous issues if not used correctly, from blocking other manifest versions, to the "upgrade always available" situation, to restricting the specific data that the package manager can give about package versions that aren't in the repo. So I'm hesitant to suggest using it. 

Especially since the version in Registry is visually similar to the `PackageVersion`. I believe that most users can visually ignore the parentheses when comparing these version numbers. Were they wildly different (such as `1.5.3.2` vs `5.3.2`) then it would be necessary. But here, I think the `PackageVersion` should be changed to match the Registry.

stephengillie (approved):

PR #188360, opened 2024-11-03
Use the dependencies Zip file for bootstrapping in SandboxTest

Trenly:
cc @ItzLevvie / @stephengillie  / @SpecterShell / @denelon / @OfficialEsco / @ImJoakim 

Checklist for Pull Requests
  * https://github.com/microsoft/winget-cli/issues/4916
  * https://github.com/microsoft/winget-cli/issues/4903
  * https://github.com/microsoft/winget-cli/issues/4778

## Description
With the latest release of WinGet, the minimum VCLibs UWP version has been bumped up. The VCLibs team is no longer publishing the new versions to the `aka.ms` URL, causing the Sandbox Test script to always fall back on the PowerShell Module and `Repair-WinGetPackageManager`, which takes significantly longer than caching the dependencies on the user filesystem and mapping them into the sandbox.  

This PR addresses a few issues with SandboxTest and how it manages dependencies

#### 1. Adds the new Dependencies Zip file
Since the new versions of VCLibs UWP does not have a publicly available download, the CLI team has started publishing a Zip file containing the dependencies with the CLI releases. This PR adds this zip file as a dependency and updates the install script inside the sandbox so that it extracts the APPX files from the ZIP and installs them. This ensures that the install does not fall back to using the PowerShell Module.

#### 2. Adds Version-Based Dependency Management
Because the dependency on Microsoft.UI.Xaml was bumped from 2.7 to 2.8, instead of installing whichever one was needed, the script would simply copy and install both inside the sandbox. In this PR, the behavior is changed so that only the required dependencies are installed.

#### 3. Known Issue - New Dependency on `DesktopAppInstaller_Dependencies.zip`
With the new dependency, if any newer releases of the WinGet CLI do not include the zip file, then it won't be able to be installed. This *should* fall back to using the PowerShell Module still, but is an untested edge case

-----

Thanks to @mdanish-kh for helping test and finding bugs in my initial implementation!

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/188360)

### DIFF
--- a/Tools/SandboxTest.ps1
+++ b/Tools/SandboxTest.ps1
@@ -158,21 +158,20 @@ if ($useNuGetForMicrosoftUIXaml) {
   $dependencies += $uiLibs_nuget
 }
 
-Write-Output $versionTag
-
-if ([System.Version]$versionTag.Trim('v') -ge [System.Version]'1.9.25180') {
+$compareVersion = $versionTag -replace '(^v)|(-preview$)'
+if ([System.Version]$compareVersion -ge [System.Version]'1.9.25180') {
   # As of WinGet 1.9.25180, VCLibs no longer publishes to the public URL and must be downloaded from the WinGet release
   $dependencies += $dependencies_inRelease
   $script:dependencySource = 'InRelease'
 } else {
   $dependencies += $vcLibsUwp
-   # As of WinGet 1.7.10582, the dependency on uiLibsUwP was bumped from version 2.7.3 to version 2.8.6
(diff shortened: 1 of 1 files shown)


mdanish-kh (approved):

denelon:
I added the SHA256 in a .txt file in the latest release.

vedantmgoyal9 (approved):

ItzLevvie (approved):

Trenly:
I'll have to update the PR to get the content from the hash file then, and remove the "If no hash" logic. I've got some other changes I want to test out too

ItzLevvie (approved):

ItzLevvie (approved):

PR #183845, opened 2024-10-17
Update SandboxTest.ps1

lars-nvhgroup:
Added "--accept-package-agreements".
It will fail if tested manifest depend on a package with an EULA.

Checklist for Pull Requests
- [x] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)?
- [ ] Is there a linked Issue?

Note: `<path>` is the directory's name containing the manifest you're submitting.

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/183845)

PR #188570, opened 2024-11-04
Refactor SandboxTest

Trenly:
cc @ItzLevvie / @stephengillie / @SpecterShell / @denelon / @OfficialEsco / @ImJoakim / @mdanish-kh 

Checklist for Pull Requests
  - https://github.com/microsoft/winget-cli/issues/4778
  - Replaces #188360 
  - Closes #183845 

## Description
When creating #188360 I had a lot of trouble following the code due to having to try and keep track of which files were where, the mapped path in the sandbox, and going through the logic of all the conditional statements. This PR completely refactors SandboxTest so that it is a bit more dynamic, has logic that is a bit easier to follow, and makes future maintenance easier.  

### 1. Parameter Validation
The original code would allow any value for the parameters, and then validate it later. This logic has been moved to the `ValidateScript` for the respective parameter. This ensures the parameters are validated before anything from the script is called and will always exit cleanly, providing better outputs for automated scenarios

### 2. Output Methods
The original script used `Write-Host` for almost everything. This means that the outputs couldn't really be piped anywhere, at least not in a meaningful way. This has been changed throughout the entire script. Instead, `Write-Information` is now used, and defaults to `Continue`.  This means that the built-in `-InformationAction` parameter now does something! By setting `-InformationAction SilentlyContinue` the outputs from the script will be suppressed. The same also works for `-WarningAction`. Note that this does not apply to the bootstrap script created inside the sandbox.

### 3. Verbose and Debug Outputs
The original script had no verbose output or debug output, meaning that anyone using the script had to be very familiar with it to know what was going on behind the scenes. This helps ensure that if a user encounters issues with the script, verbose and debug output can be requested without someone else having to replicate the issue, making it easier to fix any issues that are reported. Note that this does not apply to the bootstrap script created inside the sandbox.

### 4. Unique Exit Codes
When running a script in an automated fashion, having exit codes that actually mean something is very helpful. With these unique exit codes, it's possible to build retry or error handling into upstream callers. It also helps with debugging when trying to figure out where to start looking for an issue, since if there is a specific exit code returned it's generally a well-known error that occurred. Although this script isn't heavily used in automations (as far as I know), this change should help for anyone who wishes to use it in the future. Note that this does not apply to the bootstrap script created inside the sandbox.

### 5. Bootstrapping Script Improvement
The SandboxTest script writes a separate `bootsrap.ps1` script that is mapped into the sandbox and is what actually runs the test in the sandbox. This has been one of the larger areas of refactoring

#### 5.1 Single Output Block
In the original SandboxTest, the bootstrap script was built up piece by piece, with each piece being added in based upon the arguments the user provided, variables from within the script, and to accommodate multiple different behaviors. This has been changed to a single multi-line string. This means that when trying to add new behavior to the bootstrapping script, it is much easier to understand and know that the changes to the script will, in fact, make it into the sandbox.

#### 5.2 Better Bootstrapping Logic
The original bootstrapping script, with all its conditionals, required the paths within the sandbox for each file that needed to be installed. This made the logic of the bootstrap script very hard to follow, especially when trying to understand how a file from the host was mapped into the sandbox. Almost all of that has been removed, along with the corresponding code that determined each file's location in the sandbox. Instead, the bootstrap script now performs a sequence of operations which will install the required files without having to know exactly where they are. Simply by knowing where the mapped folder containing all the test data is, everything can be bootstrapped using this sequence of operations:

1. Switch the present directory to the test data folder
2. If there are any Zip files, extract them to the current folder - This takes care of both the case where the UI.Xaml NuGet is being used and the case where the `DesktopAppInstaller_Dependencies.zip` is being used
3. Get all `*.appx` files that are for `x64` and install them - This way we don't need to know the exact path to the installer files, even if they were extracted from a Zip
4. Install WinGet
5. If anything in steps 2-4 failed, fall back to using the PowerShell Module to install WinGet

### 6. Cleaner Dependency Management / Version-Based Dependency Management
In the original script, all the dependencies were built up as objects, and then selectively added to the dependency list. Additionally, when the dependency on Microsoft.UI.Xaml was bumped from 2.7 to 2.8, instead of installing whichever one was needed, the script would simply copy and install both inside the sandbox.

Now, the dependencies are not pre-defined and then added to the dependency list. Instead, they are defined at the moment they are added to the list. This means that it is easier to step through the logic to see what exactly is being added to the dependency list and why. This is further supported by debug messages (see above) 

### 7. Adds the new Dependencies Zip file
Since the new versions of VCLibs UWP does not have a publicly available download, the CLI team has started publishing a Zip file containing the dependencies with the CLI releases. If the release is v1.9.25180 or newer, that dependencies file will be used. This does mean that there is a dependency that each new release contains the respective Zip file and SHA256 txt file. However, even if the new releases don't have these, the bootstrap script will fall back to using the PowerShell Module

### 8. More Variables
I hesitate on whether to call this an improvement, since it does add a layer of obfuscation and complexity to the script. However, this also changes it so that every variable is extremely descriptively named. While not every hardcoded value has been made into a variable, many of the items which are built up from other variables are. 

#### 9.1 File Cleanup
The original script would leave around many of the temporary files that it downloaded for checking SHAs. The script now cleans these up when it exits. 

#### 9.2 Dependencies Caching
In the original script, the dependencies were cached in the same location as all the test data. This meant they were explicitly excluded from being cleaned up with each run. They have now been moved to their own directory. This way, the test data folder can be completely deleted, and upon re-creation, the dependencies are copied from the cache over into the test data folder. This ensures that the test data is always fully cleaned and there are no rogue files that could interfere. 

#### 9.3 Location of Files
All of the sandbox test files used to be located in the TEMP folder. Although this isn't wrong, I didn't feel that was the best place for them. Instead, they now share the same APPDATA folder as WinGet. This was done because the msixbundle installers for WinGet are already being stored there. The new `DesktopAppInstaller_Dependencies.zip` file is stored in the same location as its respective WinGet msixbundle, and the other files are stored at the root of the WinGet AppData folder (`%LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe`).  

This also provides the benefit that if WinGet is uninstalled from the host system, all of the cached files are also removed.

### 10. Comments 
Lots more comments. Nothing more to explain
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/188570)


ItzLevvie (approved):

Tools/SandboxTest.ps1
```diff
@@ -1,269 +1,518 @@
...
+function Invoke-CleanExit {
+    param (
+        [Parameter(Mandatory = $true)]
+        [int] $ExitCode
+    )
+    Invoke-FileCleanup -FilePaths $script:CleanupPaths
+    $script:WebClient.Dispose()
+    Write-Debug "Exiting ($ExitCode)"
+    exit $ExitCode
 }
```
vedantmgoyal9, line 118: The description of this function doesn't seem right. It is of `Initialize-Folder`.

vedantmgoyal9 (changes requested):

vedantmgoyal9 (approved):

ItzLevvie (approved):

Tools/SandboxTest.ps1
```diff
@@ -1,269 +1,518 @@
...
+$script:DependenciesZipFileName = "$script:DependenciesBaseName.zip" # This should exactly match the name of the file in the CLI GitHub Release
+
+# Download Urls
+$script:VcLibsDownloadUrl = 'https://aka.ms/Microsoft.VCLibs.x64.14.00.Desktop.appx'
+$script:UiLibsDownloadUrl_v2_7 = 'https://github.com/microsoft/microsoft-ui-xaml/releases/download/v2.7.3/Microsoft.UI.Xaml.2.7.x64.appx'
+$script:UiLibsDownloadUrl_v2_8 = 'https://github.com/microsoft/microsoft-ui-xaml/releases/download/v2.8.6/Microsoft.UI.Xaml.2.8.x64.appx'
+$script:UiLibsDownloadUrl_NuGet = 'https://globalcdn.nuget.org/packages/microsoft.ui.xaml.2.8.6.nupkg?packageVersion=2.8.6'
+
+# Expected Hashes
+$script:VcLibsHash = 'B56A9101F706F9D95F815F5B7FA6EFBAC972E86573D378B96A07CFF5540C5961F'
```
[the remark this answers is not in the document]
Trenly: Thanks; This was from my testing, forgot to remove

Tools/SandboxTest.ps1
```diff
@@ -1,269 +1,518 @@
...
-  Add-AppxPackage -Path '$($vcLibsUwp.pathInSandbox)' -ErrorAction Stop
-  Add-AppxPackage -Path '$($uiLibsUwp_2_7.pathInSandbox)' -ErrorAction Stop
-  Add-AppxPackage -Path '$($uiLibsUwp_2_8.pathInSandbox)' -ErrorAction Stop
-  Add-AppxPackage -Path '$($desktopAppInstaller.pathInSandbox)' -ErrorAction Stop
+    Get-ChildItem -Filter '*.zip' | Expand-Archive
+    Get-ChildItem -Recurse -Filter '*.appx' | Where-Object {`$_.FullName -match 'x64'} | Add-AppxPackage -ErrorAction Stop
+    # This path is set explicitly instead of using Get-ChildItem as an error prevention measure
+    Add-AppxPackage './$($script:AppInstallerPFN).msixbundle' -ErrorAction Stop
 } catch {
   Write-Host -ForegroundColor Red 'Could not install from cached packages. Falling back to Repair-WinGetPackageManager cmdlet'
```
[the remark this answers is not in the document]
Trenly: I like the idea; Should be easy to implement, will do!

Tools/SandboxTest.ps1
```diff
@@ -1,269 +1,518 @@
...
+    # Script
+    [Parameter(Position = 1, HelpMessage = 'The script to run in the Sandbox.')]
+    [ScriptBlock] $Script,
+    # MapFolder
+    [Parameter(HelpMessage = 'The folder to map in the Sandbox.')]
+    [ValidateScript({
+            if (-Not (Test-Path -Path $_ -PathType Container)) { throw "$_ is not a folder." }
+            return $true
+        })]
+    [String] $MapFolder = $pwd,
```
mdanish-kh, line 28: If I launch the script from `C:\`, something like `& C:\Users\<user>\winget-pkgs\Tools\SandboxTest.ps1`, this will give sandbox read & write access to my entire system, which may not be obvious to the user running the script, and can allow an app inside sandbox to do anything it wants.

Should we disallow calling the script this way, or warn user when a parent path of `winget-pkgs` folder is marked as read-or-write?

mdanish-kh: I realize this is not strictly relevant to the refactor changes here as it was the same behavior before, but I thought it was worth calling out

Trenly: I like the idea of disallowing it, but that could cause issues in some use cases. Take for example my setup - I have everything inside of a Dev drive. It could be possible that I want to open the sandbox with read-write into the dev drive. 

I do think that adding a warning if the Map-Folder doesn't contain `winget-pkgs` would be a good idea though

mdanish-kh (approved):
We should probably update [SandboxTest.md](https://github.com/microsoft/winget-pkgs/blob/master/doc/tools/SandboxTest.md) with the new params and behaviors introduced by the script, but that can be a separate PR. The script is looking in a good state to take in now

cc @denelon

Trenly:
> We should probably update [SandboxTest.md](https://github.com/microsoft/winget-pkgs/blob/master/doc/tools/SandboxTest.md) with the new params and behaviors introduced by the script, but that can be a separate PR. The script is looking in a good state to take in now
> 
> cc @denelon

The only new prameter is actually `-Clean`, everything else stayed the same. Was trying to keep everything backwards compatible so I didn’t break anyones workflows

denelon (approved):

PR #191690, opened 2024-11-15
New package: JPSoft.tcmd version 33.0.18.0

### LINKED ISSUE #191609: [Package Request]: TCMD v32 x64
### How can we help?

I would like someone else to build the manifest.

### Please read and ensure the following

- [x] The installer meets the above requirements

### Please provide the following information

I am honestly not sure if the installer meets the above requirements. It requires Windows Installer service. I 

Download Page Url: https://jpsoft.com/all-downloads/all-downloads.html
Publisher: JPSoft
Package Name: TCC (TakeCommand) Command Line v32 x64
Description: advanced 64-bit command line predating Norton Utilities in the 1980s and under constant development for 40-ish years.
Package Version: <whatever is latest>
Installer Url: https://jpsoft.com/downloads/v33/tcmd.exe

These are the options on the TCMD.exe installer:
Image

Dragon1573:
## Checklist for Pull Requests

- [x] Resolve #191609

## Manifests

## Manual validation

Image
Image
Image

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/191690)

### DIFF
--- a/manifests/j/JPSoft/tcmd/33.0.18.0/JPSoft.tcmd.installer.yaml
+++ b/manifests/j/JPSoft/tcmd/33.0.18.0/JPSoft.tcmd.installer.yaml
@@ -1,0 +1,23 @@
+# Created with komac v2.6.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+
+PackageIdentifier: JPSoft.tcmd
+PackageVersion: 33.0.18.0
+InstallerType: exe
+Scope: machine
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+InstallerSwitches:
+  Silent: /exenoui /qn
+  SilentWithProgress: /exebasicui /qb
+UpgradeBehavior: install
+ReleaseDate: 2024-11-05
+ProductCode: '{3F8B0D52-1415-4ED1-A23A-78148BBB1138}'
+Installers:
+- Architecture: x64
+  InstallerUrl: https://jpsoft.com/downloads/v33/tcmd.exe
+  InstallerSha256: CC89B35A18991CCD619FC8BD506793A874606C0250D80003877250468F9D2465
+ManifestType: installer
+ManifestVersion: 1.6.0
--- a/manifests/j/JPSoft/tcmd/33.0.18.0/JPSoft.tcmd.locale.en-US.yaml
+++ b/manifests/j/JPSoft/tcmd/33.0.18.0/JPSoft.tcmd.locale.en-US.yaml
@@ -1,0 +1,19 @@
+# Created with komac v2.6.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.6.0.schema.json
+
+PackageIdentifier: JPSoft.tcmd
+PackageVersion: 33.0.18.0
+PackageLocale: en-US
+Publisher: JP Software
+PublisherUrl: https://jpsoft.com/
+PublisherSupportUrl: https://jpsoft.com/help
+PrivacyUrl: https://jpsoft.com/company/privacy.html
+PackageName: Take Command 33
+PackageUrl: https://jpsoft.com/products/take-command.html
+License: Proprietary
+LicenseUrl: https://jpsoft.com/company/license.html
+Copyright: Copyright (C) 2024 JP Software
+ShortDescription: Windows console replacement, includes TCC and CMDebug.
+PurchaseUrl: https://jpsoft.com/buy-now.html
+ManifestType: defaultLocale
+ManifestVersion: 1.6.0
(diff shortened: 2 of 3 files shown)


JoeC4281:
Will a new manifest have to be created for each new release of TCMD? 

What if I want to download a previous release of TCMD?

Ref: https://jpsoft.com/all-downloads/all-downloads.html 

Joe

Dragon1573:
> Will a new manifest have to be created for each new release of TCMD?

YES. Manifests in the repository are version-related. Each manifest will only point to one exact version of the package.

I can't sure if the current [`InstallerUrl`](https://jpsoft.com/downloads/v33/tcmd.exe) is vanity. If it is, there can only be one manifest associated to the URL, any outdated versions with the same `InstallerUrl` should be removed from the repository.

> What if I want to download a previous release of TCMD?

Since this pull request is the first which related to this package, so there will be no "previous version" available via `winget.exe`.

In my opinion, any outdated (or deprecated) version should be ignored when a contributor or volunteer submit manifests of a package. If an outdated version is already exists in the repository, and the `InstallerUrl` is still valid, it can be kept.

If you insist on using an outdated version, don't hesitate to create another issue asking for it. ❤️

ItzLevvie (approved):

ClaireCJS:
Wow, it works! 

But currently only for new installs. 

It can't upgrade.

But there is a misunderstanding. One doesn't really upgrade. One simply installs a new version in a new folder.  One can have many different versions "installed" at the same time.  

Sure, only one gets associated in the GUI with BTM/BAT files, but from the command line, that doesn't matter.

Any way to make this one install to c:\TCMD32\ ?  

I have c:\TCMD31, c:\TCMD30, c:\TCMD29, etc. I just symlink \TCMD\ to the correct one.  Doesn't really matter where it's put. Multiple versions can co-exist just fine.

JoeC4281:
Why not just symlink to the default install location?

C:\Program Files\JPSoft\TCMD33

Joe

JoeC4281:
I really don't see a problem with upgrading.

It need not be upgraded via winget.

Option /U is all that is needed to be issued from the command line to upgrade an existing copy of TCMD.

Note well, though, that when Rex posts a new build version (the next release will probably be 33.0.19), 
it will be downloaded from the same URL as 33.0.18, 
as the download is still the installer file TCMD.EXE

Will this be a problem for winget?

We will have to wait and see, 
as I am not sure how the manifests for winget work,
that is, are they specific to 33.0.18,
or will it work with 33.0.19,
since it is the same installer file (TCMD.EXE)

Only when the version becomes 34.x.x, 
will the download URL be different. 

Joe

ClaireCJS:
I try to keep all old versions, particularly for back-finding-out when a newly-discovered bug started, or if i decide to build a dosbox and need an older version.  I never delete one, just install to a new folder.   a lot of holistic stuff over here and it can be a month before finding out something is broken, at which point a simple junction command reverts to the old version

Dragon1573:
> Any way to make this one install to c:\TCMD32\ ?

I think it can be achieved by manually configuring the install location via `-l` or `--location` switch of `winget.exe`.

> I have `C:\TCMD31`, `C:\TCMD30`, `C:\TCMD29`, etc. I just symlink `\TCMD\` to the correct one. Doesn't really matter where it's put. Multiple versions can co-exist just fine.

Currently, they are all matched by ARP entries to exactly the same package with different versions. So `winget.exe` will ask you to upgrade and keep just the latest version.

In issue panel of this repository, "multiple version co-exists" is a problem instead of an acceptable case. If "co-exists" feature should be enabled, they might need to be split into many packages with different package identifiers. That may become a mess.

CC @stephengillie.

Dragon1573:
> Note well, though, that when Rex posts a new build version (the next release will probably be 33.0.19), it will be downloaded from the same URL as 33.0.18, as the download is still the installer file TCMD.EXE
> 
> Will this be a problem for winget?

Sometimes. 🫤

The URL in this case is called "vanity URL". It's a stable, static URL which always points to the latest version of an installer file.

When a new installer file is released, the file is totally different from that before and may cause https://github.com/microsoft/winget-pkgs/labels/Error-Hash-Mismatch.

There're several packages facing exactly the same problem. We just create a new manifest of the new version with exactly the same installer URL, and remove the outdated one. For this case, each major version will keep only 1 manifest.

Or until we find a version-related static URL for each minor or patch release.

-----

### Current situation

```text
# Only one manifest of version v33.* can be kept in the repository.
https://example.com/v33/installer.exe ---> All version >=33.0.0.0,<34.0.0.0
```

### Our target

```text
# Figure out an installer URL which point to the exact version.
https://example.com/v33/33.0.0.0/installer.exe ---> Exactly version ==33.0.0.0
https://example.com/v33/33.0.0.1/installer.exe ---> Exactly version ==33.0.0.1
https://example.com/v33/33.0.1.0/installer.exe ---> Exactly version ==33.0.1.0
https://example.com/v33/33.1.0.0/installer.exe ---> Exactly version ==33.1.0.0
```

PR #196898, opened 2024-12-07
Don't explicitly set ArchiveBinariesDependOnPath: false

mdanish-kh:
cc @Trenly

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/196898)


Trenly (approved):
I know the CLI defaylts to false - thought that having the field set made it easier to change if someone beeded to change it later and would help reviewers when looking at new zip packages being added.

Either way, this change looks fine

mdanish-kh:
I see your point - My line of reasoning was to keep the manifest as simple to understand as possible for contributors, since once we add this field (either true or false), it'll likely be carried forward in all subsequent versions

Trenly:
@denelon - can we get a merge?

denelon (approved):

PR #201999, opened 2024-12-24
New package: GovernmentOfJapan.MIC.miChecker version 3.1

DuckDuckStudio:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/201999)

### DIFF
--- a/manifests/g/GovernmentOfJapan/MIC/miChecker/3.1/GovernmentOfJapan.MIC.miChecker.installer.yaml
+++ b/manifests/g/GovernmentOfJapan/MIC/miChecker/3.1/GovernmentOfJapan.MIC.miChecker.installer.yaml
@@ -1,0 +1,20 @@
+# Created with komac v2.9.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.9.0.schema.json
+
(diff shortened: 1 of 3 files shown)


DuckDuckStudio:
All of them are https://www.soumu.go.jp/ , how come there are https://github.com/microsoft/winget-pkgs/labels/Validation-Domain label? 🤔

stephengillie:
Verify:
image
Requires JDE or JDK. Does the package comes with its own? If not...
Since the package manager doesn't support wildcards in dependencies, we usually omit the dependency in packages that require `JAVA`. (`GIT` and similar packages are handled likewise.)

Retrying with `Oracle.JavaRuntimeEnvironment`:
image
After install, I got a different error. Log: 
[1735155793928.log](https://github.com/user-attachments/files/18247051/1735155793928.log)

DuckDuckStudio:
image

WHAT???

DuckDuckStudio:
Based on the documentation, it appears that this package requires jre to be placed directly into the installation directory, which winget may not be able to do.
image

PR #206620, opened 2025-01-04
Remove Package: Fort Firewall

ghost:


Emi-Emi-Emi:
I decided am not going to keep this, and I doubt anyone else will do it, so it's better if it gets removed because it will be too outdated eventually since Fort Firewall gets updated frequently.

Thank you and have a good day.

asihd0as8hd8shad:

stephengillie:
URL: https://github.com/tnodir/fort/releases/download/v3.15.6/FortFirewall-3.15.6-windows10-x86_64.exe 
OK

(Automated message - build 924)

stephengillie:
Hi @asihd0as8hd8shad,

(Automated response - build 934.)

asihd0as8hd8shad:
> I decided am not going to keep updating this, and honestly, I doubt anyone else will do it, since nobody really cares about the winget version, so it's better if it gets removed because it will be too outdated eventually since Fort Firewall gets updated frequently.
> 
> Thank you and have a good day.

Trenly:
@asihd0as8hd8shad - Even if you aren't maintaining it, there are scans that the community runs to update packages

asihd0as8hd8shad:
@Trenly Well, that's the issue, what if not. the development slowed a little because of Christmas and New year, but in one day there were 3 updates, and then, x86 link doesn't work so I had to use the previous version.

So if FF gets behind updates, since this is a firewall, it could mean issues or security risks, and while FF offers and self update mechanism, there is always a risk of some new features missing or something that could be essential, so the official version / github, in my opinion, would be the best thing.

But it is your choice, there can be merge or close and I was just waiting for this to delete my account and rest a while, which is the reason why I even changed the username and all.

Take care and have a good day.

Trenly:
I'll leave the decision up to @stephengillie, since he's the one who originally added the label

PR #212073, opened 2025-01-14
New version: dnGrep.dnGrep-4.3.3.0

doug24:
Checklist for Pull Requests

Manifests
- [X] Have you [validated](https://github.com/microsoft/winget-pkgs/blob/master/doc/Authoring.md#validation) your manifest locally with `winget validate --manifest <path>`?
- [X] Does your manifest conform to the [1.9 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.9.0)?

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/212073)

### DIFF
--- a/manifests/d/dnGrep/dnGrep/4.3.3.0/dnGrep.dnGrep.installer.yaml
+++ b/manifests/d/dnGrep/dnGrep/4.3.3.0/dnGrep.dnGrep.installer.yaml
@@ -1,0 +1,42 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.9.0.schema.json
+
+PackageIdentifier: dnGrep.dnGrep
+PackageVersion: 4.3.3.0
+InstallerLocale: en-US
+Installers:
+- Architecture: x86
+  InstallerType: wix
+  Platform:
+  - Windows.Desktop
+  InstallerUrl: https://github.com/dnGrep/dnGrep/releases/download/v4.3.3.0/dnGREP.4.3.3.x86.msi
+  InstallerSha256: df633c5b63dad7c5b396a4d297308fbef2011844a944efc04ff180fda2e73456
+  ProductCode: '{59D5D2E9-0EF6-4536-8999-B96893D50DA0}'
+  Scope: machine
+  InstallerSwitches:
+    Silent: /quiet
+    SilentWithProgress: /passive
+    Custom: /norestart
+  UpgradeBehavior: install
+  ReleaseDate: 2025-01-14
+  AppsAndFeaturesEntries:
+  - DisplayName: dnGrep 4.3.3 (x86)
+    UpgradeCode: '{4CF55558-B642-482E-9FBF-06EA5AEC8276}'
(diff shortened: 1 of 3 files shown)


stephengillie (approved):

Trenly:
@doug24 - If you're able, would you be able to use a tool like Komac, YamlCreate, or WingetCreate for submitting new manifests in the future? I noticed that items like the `Platform`, `InstallerType`, `InstallerSwitches`, etc. are common between both installers. Having these at the root level of the manifest instead of at the installer level would make the manifest easier to maintain, and tools like Komac or YamlCreate will automatically move these fields to the root level whenever they can

doug24:
@Trenly Thank-you for pointing this out, I wasn't aware the manifest wasn't correct. 

The project used to run Winget-Releaser github action, and it worked great for a while. Then it started getting random errors calling Komac and I would have to rerun the action five times before it worked. Then the action just stopped running at all. I created issues on that project, but they were not resolved, and I deleted the action.  

I created the manifest manually for a couple of releases then I wrote my own tool to generate them. Looks like I missed something in my template.  I'll try running Komac again and use that or update my tool to fix these issues.

Trenly:
It's all good @doug24! The manifest technically isn't incorrect, it will still work just fine, I had simply noticed while performing some maintenance that many of the items were duplicated and could be moved so they only need to be specified once. 

If you're looking to use a GitHub action, I recommend using `wingetcreate`. It's super easy - just needs the URLs, the PackageVersion, and a GitHub access token. There's a great example at dotenvx - https://github.com/dotenvx/dotenvx/blob/2ce11585136fffcc08136abb9cb686d105fe3f62/.github/workflows/ci.yml#L375-L389

doug24:
@Trenly Help, please!

I tried to follow your guidance and change the format of the manifest file in my new release today: https://github.com/microsoft/winget-pkgs/pull/227650

But as you can see it failed with a BlockingDetectionFound 4001.  I've never had that issue before., and the code is scanned before it is signed by SignPath.

Then I looked and see that you have been changing other dnGrep manifests and one that had previously passed also failed with BlockingDetectionFound 4001.

What is going on here?  As far as I can see there is no information about what was detected and the links to more information are worthless:  

> If you get a Binary-Validation-Error label on your pull request, update your software to remove the code detected as PUA.

Without any information about what was detected, I can't possibly change the application to fix it!

Is this related to the manifest format?  Or is it some new test that is being run?

Trenly:
As far as I know, it is not related to the manifest version or format at all. 

@denelon - This is another case where we need help from whichever team is responsible for the scans

denelon:
Yep, we're trying to get in touch with the right folks who can look into the "instances" of validations where we're getting blocking detections and figure out what they are. There are a bunch of different blocking detections, and it's not obvious with the response we get today what the "inner" error is as opposed to the 4001 top level result.

PR #200010, opened 2024-12-20
Remove version: CiKSolutions.LogTagAnalyzer version 3.2.5.2

DuckDuckStudio:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/200010)

PR #222988, opened 2025-02-05
Modify the reply message for label xxx-Version-Removal

DuckDuckStudio:
The content of the response here is copied from (part of) Stephengillie's automated response. 

<details>
<summary>Outdated</summary>
<p><del>When a PR author makes a removal request, the reason has usually been stated (somewhere).</del></p>
<p><del>This may duplicate Stephengillie&#39;s automated reply, but other moderators seem to just add labels to the corresponding PRs without replying.</del></p>
</details>

<details>
<summary>Resolved</summary>
<p>For the Highest-Version-Removal and Last-Version-Removal labels, while they are <a href="https://github.com/microsoft/winget-pkgs/pull/155074#pullrequestreview-2078182152">semantically distinct</a>, this can lead to a situation like in <a href="https://github.com/microsoft/winget-pkgs/pull/200010">#200010</a>, where Last-Version-Removal was added about a week after it was added and then Highest- Version-Removal. i think we can keep only one of them, normally we don&#39;t remove the highest version and keep the lower one (if it&#39;s a mistake it should be Modify not Remove, the version is already unavailable when the installer hash changes/installer removed, no need to remove, just waiting it fixed and submit a Modify PR)</p>
</details>

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/222988)


.github/policies/labelAdded.highestVersionRemoval.yml
```diff
@@ -20,13 +20,10 @@ configuration:
         then:
           - addReply:
               reply: >-
-                Hello @${issueAuthor},
+                Hi @${issueAuthor},
```
Trenly, line 23: Most other responses use `Hello`. This should be kept consistent

.github/policies/labelAdded.highestVersionRemoval.yml
```diff
@@ -20,13 +20,10 @@ configuration:
...
 
 
-                This PR removes the highest available version of the package from the repository.
-
-
-                Please provide the reason for removal to help our moderators understand the context.
```
Trenly, line 29: Can we keep this portion of the message? There are several cases where users submit removals without putting anything in the body to indicate why it is being removed

.github/policies/labelAdded.highestVersionRemoval.yml
```diff
@@ -20,13 +20,10 @@ configuration:
...
+                This manifest has the highest version number for this package. Is it available from another location? (This might be in error if the version is switching from semantic to string, or string to semantic.)
```
Trenly, line 26: I know you mentioned the semantic distinction between *Highest*- and *Last*- version removals in your post, but I think the distinction here needs to be very clear, and this new message doesn’t seem to capture that difference

If a user is removing the highest version, there should be other versions in the repo that users could use. This is usually the case when a publisher unpublishes the latest version, or a pre-release package is mistakenly submitted under a release identifier. In these cases, it doesn’t make sense to ask if the package is available from a different location. However, it helps the moderators a lot if the removal reason is provided so that the PR can be assessed properly.

If a user is removing the last version in the repo, the package will become entirely unavailable. This is usually the case when a link goes 404 and nobody has checked to see if a new version is available. In this case, asking if it is available from another location does make sense.

-----

I know that I apply the incorrect label occasionally (using highest- when it should be last-), but Stephens automation is usually pretty good about determining the right one to use. 

I agree the messaging could be clearer, but I don’t think this PR makes the distinction clear enough

Trenly (changes requested):

.github/policies/labelAdded.highestVersionRemoval.yml
DuckDuckStudio: Keep this sentence?
```
(This might be in error if the version is switching from semantic to string, or string to semantic.)
```

Trenly: I think that makes sense

DuckDuckStudio: Done.

.github/policies/labelAdded.lastVersionRemoval.yml
```diff
@@ -23,10 +23,13 @@ configuration:
                 Hello @${issueAuthor},
 
 
-                This PR removes the last version of the package from the repository.
+                This PR removes the last version of this package in the repository, meaning it will completely remove this package from the repository.
```
Trenly, line 26:
```suggestion
                This PR removes the last version of this package in the repository, meaning it will completely remove this package from WinGet.

```

Trenly (approved):

Trenly:
LGTM @denelon

denelon (approved):

Issue #189849, opened 2024-11-08
[New Feature]: Avoid "API rate limit"


Dragon1573:
### Description of the new feature/enhancement

This is a re-request of a feature. It was implemented before, but soon opted-out with a reversion. I want to re-add this feature back. ❤

- Previously requested by #167997
- First attempt (but rejected) by #168000
- Previously implemented by #168384
- Previously reverted by #168747
- Previously "revert back" (but rejected) by #170577
  - It was rejected because https://github.com/microsoft/winget-pkgs/pull/168384#issuecomment-2318919294

Dragon1573:
> [!NOTE]
>
> Before `winget.exe` version `1.9.x` generally available, I can simply perform #170577 on my local repository and keep syncing with upstream by using rebase action. Then `Tools/SandboxTest.ps1` changed a lot and I met huge merge conflicts.
>
> "API rate limit" problem really affects me. I'd request an official support resolving this problem. 🙏🏼

Trenly:
Close with reason: Addressed by above PR;

PR #230208, opened 2025-02-16
Add function for checking and caching GitHub Tokens

Trenly:
Checklist for Pull Requests
  * #189849

## Description
This PR re-adds authentication using GitHub tokens for `SandboxTest.ps1`

This is achieved by adding a function which caches the provided GitHub token to avoid making unnecessary calls to the GitHub API. The token cache is located inside WinGet's AppData folder so that uninstalling WinGet will also remove the cache. This also ensures that the tokens are cached in a well-known location so this function can be ported to other tools such as PRTest.ps1 and YamlCreate.

For security, the cache does not contain the users actual GitHub token, but instead contains the HASH of the token. Where possible, the expiration of the token is also cached. After a token has been in the cache for more than 30 days, it is re-validated. 

## Reviewers
* @mdanish-kh 
* @denelon
* @stephengillie 

## Watchers
* @dragon1573

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/230208)

### DIFF
--- a/Tools/SandboxTest.ps1
+++ b/Tools/SandboxTest.ps1
@@ -321,7 +321,7 @@ function Test-GithubToken {
         $cachedTokenAge = [Math]::Round($cachedTokenAge, 2) # We don't need all the precision the system provides
         Write-Debug "Token has been in the cache for $cachedTokenAge days"
         $cacheIsExpired = $cachedTokenAge -ge $script:CachedTokenExpiration
-        $cachedTokenContent = Get-Content $cachedToken
+        $cachedTokenContent = (Get-Content $cachedToken -Raw).Trim() # Ensure any trailing whitespace is ignored
         $cachedTokenIsEmpty = [string]::IsNullOrWhiteSpace($cachedTokenContent)
 
         # It is possible for a token to be both empty and expired. Since these are debug and verbose messages, showing both doesn't hurt
@@ -369,12 +369,13 @@ function Test-GithubToken {
 
     Write-Verbose "Checking Token against $($requestParameters.Uri)"
     $apiResponse = Invoke-WebRequest @requestParameters # This will return an exception if the token is not valid; It is intentionally not caught
-    $rateLimit = $apiResponse.Headers['X-RateLimit-Limit']
-    $tokenExpiration = $apiResponse.Headers['github-authentication-token-expiration'] # This could be null if the token is set to never expire
+    # The headers can sometimes be a single string, or an array of strings. Cast them into an array anyways just for safety
+    $rateLimit = @($apiResponse.Headers['X-RateLimit-Limit'])
+    $tokenExpiration = @($apiResponse.Headers['github-authentication-token-expiration']) # This could be null if the token is set to never expire
     Write-Debug "API responded with Rate Limit ($rateLimit) and Expiration ($tokenExpiration)"
 
-    if (!$rateLimit) { return $false } # Something went horribly wrong, and the rate limit isn't known. Assume the token is not valide
-    if ($rateLimit -le 60) {
+    if (!$rateLimit) { return $false } # Something went horribly wrong, and the rate limit isn't known. Assume the token is not valid
+    if ($rateLimit[0] -le 60) {
         # Authenticated users typically have a limit that is much higher than 60
         Write-Warning "You may encounter 'API rate limit exceeded' errors! Please consider adding `GITHUB_TOKEN` to your environment variables"
         return $false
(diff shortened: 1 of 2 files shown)


Tools/SandboxTest.ps1
```diff
@@ -273,6 +305,132 @@ function Test-FileChecksum {
...
+            }
+        }
+        else {
+            # Either the token was empty, or the cached token is expired. Remove the cached token so that re-validation
+            # of the token will update the date the token was cached if it is still valid
+            Invoke-FileCleanup -FilePaths $cachedToken.FullName
+        }
+    }
+    else {
+        Write-Verbose 'Token was not found in the cache'
```
mdanish-kh, line 400: It's possible that a hash-file exists in the directory, but the user is using a different / new token. It may make sense to clean up all other files in the directory in this condition

Tools/SandboxTest.ps1
```diff
@@ -273,6 +305,132 @@ function Test-FileChecksum {
...
+    if ([int]$rateLimit[0] -le 60) {
+        # Authenticated users typically have a limit that is much higher than 60
+        return $false
+    }
+
+    Write-Verbose 'Token validated successfully. Adding to cache'
+    # If the token doesn't expire, write a special value to the file
+    if (!$tokenExpiration -or [string]::IsNullOrWhiteSpace($tokenExpiration[0])) { $tokenExpiration = @([System.DateTime]::MaxValue) }
+    # When datetime objects are converted to string, they have extra whitespace that should be trimmed off
+    $tokenExpiration = $tokenExpiration[0].ToLongDateString().Trim()
```
mdanish-kh, line 428:
```suggestion
    $tokenExpiration = [DateTime]::Parse($tokenExpiration[0]).ToLongDateString().Trim()
```

I'm really not sure why you and I get different GitHub API responses, but `$tokenExpiration[0]` is a string for me.


image

Tools/SandboxTest.ps1
```diff
@@ -273,6 +305,132 @@ function Test-FileChecksum {
...
+
+        # It is possible for a token to be both empty and expired. Since these are debug and verbose messages, showing both doesn't hurt
+        if ($cachedTokenIsEmpty) { Write-Verbose 'Cached token had no content. It will be re-validated' }
+        if ($cacheIsExpired) { Write-Verbose "Cached token is older than $script:CachedTokenExpiration days. It will be re-validated" }
+
+        if (!$cacheIsExpired -and !$cachedTokenIsEmpty) {
+            # Check the content of the cached file in case the actual token expiration is known
+            Write-Verbose 'Attempting to fetch token expiration from cache'
+            # Since Github adds ` UTC` at the end, it needs to be stripped off. Trim is safe here since the last character should always be a digit
+            $cachedExpirationForParsing = $cachedTokenContent.TrimEnd(' UTC')
```
mdanish-kh, line 365: Since we use `ToLongDateString()` when saving the token, it shouldn't have ` UTC` anymore at the end

Tools/SandboxTest.ps1
```diff
@@ -273,6 +305,132 @@ function Test-FileChecksum {
...
+            $cachedExpirationDate = [System.DateTime]::MinValue
+            [System.DateTime]::TryParse($cachedExpirationForParsing, [ref]$cachedExpirationDate)
```
mdanish-kh, line 367:
```suggestion
            # Pipe to Out-Null so that it doesn't get captured in the return output
            [System.DateTime]::TryParse($cachedExpirationForParsing, [ref]$cachedExpirationDate) | Out-Null
```

The output of `Test-GitHubToken` is an Object instead of a boolean because the output of this TryParse was getting captured in the return output. Explicitly pipe to null to get a boolean output

https://stackoverflow.com/questions/21866021/boolean-variable-gets-returned-as-an-object


image

mdanish-kh (changes requested):

Tools/SandboxTest.ps1
Trenly: Some users (like me) may have multiple tokens. For example, I have an account through GitHub EMU and my personal account (this one). It's also possible for a sinle user to have multiple tokens with different permissions.

Or, its possible that some users may share a single PC and a single Windows login, but use different GitHub tokens.

This implementation specifically allows for multiple token hashes to be in the cache to account for these scenarios

Tools/SandboxTest.ps1
Trenly: True; I'll lleave this just in case though

mdanish-kh (approved):

denelon:
@AmelBawa-msft I wanted to call this one out since we've been looking at how we handle GitHub credentials over at winget-create.

ItzLevvie:
[Policy] Needs-Manual-Merge

stephengillie (approved):

stephengillie (approved):

PR #231729, opened 2025-02-20
Remove version: 3.3Kontakt version 5.10.5.9143

kris6673:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/231729)


vcodst:
I'm working for Dstny, and this package is indeed creating problem for us as we are publishing the same application for different brands with the same product code.
So I've asked kris6673 to remove this package to avoid people using other brands to be automatically updated with 3Kontakt application.

stephengillie (approved):

Issue #500, opened 2020-05-20
Auto generate manifests for newer versions


felipecassiors:
# Disclaimer

I tried to find an issue like this in both repositories, but I didn't.

# Description of the new feature/enhancement

To support some kind of CI that, based on previously set logic, can auto-generate manifests for newer versions of the packages.

# Proposed technical implementation details

A very good example is the Chocolatey Automatic Package Updater Module: https://github.com/majkinetor/au

Based on custom rules, such as scripts for finding newer versions, URLs for downloading them, a CI can run continuously based on a scheduled trigger, and if it finds a newer version, it try to build a new manifest automatically, test it, and submit a PR (or push directly).

So then, we don't need to manually create manifests for every new version of an existing package.

superusercode:
Scoop also has their own implementation of this: https://github.com/lukesampson/scoop/wiki/App-Manifest-Autoupdate

felipecassiors:
If we settle on an implementation archtecture I think the community would be excited to help. **I also think this is something to put on the v1 backlog.**

I was thinking of something like:

Every package (which has its own folder inside the `Publisher`) can have a file called `update.yml`. This file can be like

``` YAML  
template:
  Id: GitHub.cli
  Name: GitHub CLI
  AppMoniker: gh
  InstallerType: MSI
updaters:
  - arch: x64
     uses: updater-github-releases
     with: 
       regex: '*_windows_amd64.msi'
  - arch: x86
     uses: updater-github-releases
     with:
       regex: '*_windows_386.msi'
```

And then, `updater-github-releases` would be a plugin for the packages delivered through GitHub Releases. 

This enables people to create every kind of plugin, and this way making them reusable without having to share code. People could create plugins which are applicable for a single distributor, but for that case, we can also have an extremely generic plugin, such as `updater-html-webscrap`:

``` YAML
updaters:
  - arch: x64
     uses: updater-html-webscrap
     with:
       versionFilter: '.list-group-item:contains(version)'
       url: https://contoso.com/download
       downloadUrl: "https://contoso.com/download/contoso_${{ version }}.msi"
```

These plugins must return some values, such as `version`, `x64.url`, and `x64.hash` for instance.

These plugins could be `npm` modules for easy management.

felipecassiors:
And then, a new project called `winget-updater` (could be written in TypeScript for example). 

It needs to be a CLI which can be called passing the `update.yml` or an option for recursively check a folder for `update.yml`'s, and then:
[...]

PR #236105, opened 2025-03-09
[Spec] Auto generate manifests for newer versions

vedantmgoyal9:
I've drafted a spec on how Winget automation could work based on what I’ve built so far, but the field names and other details may need a complete overhaul. Suggestions and feedback are more than welcome!

cc @denelon @stephengillie @Trenly @ImJoakim @ItzLevvie @jedieaston @KaranKad @mdanish-kh @OfficialEsco @quhxl @russellbanks @SpecterShell @PckgrBot

- Related issue: #500


doc/specs/#500 - Auto generate manifests for newer versions.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+The above JSON defines automation metadata and will be stored in a `.auto_updates.json` file alongside `.validation` files.
+
+### Automation Workflow
+
+The automation consists of a PowerShell script that:
+
+1. Loads variables specified in the `AdditionalInfo` object.
+2. Executes `Invoke-RestMethod` or `Invoke-WebRequest` with the defined arguments (supporting multiple sources for different APIs).
```
Trenly, line 57: How would software from RSS feeds be handled?

vedantmgoyal9: https://devblogs.microsoft.com/scripting/use-windows-powershell-to-parse-rss-feeds - like a normal XML

doc/specs/#500 - Auto generate manifests for newer versions.md
```diff
@@ -0,0 +1,87 @@
...
+    "ForceUpgrade": false
+  },
+  "Skip": {
+    "Skip?": false,
+    "Reason": null
+  }
+}
+```
+
+The above JSON defines automation metadata and will be stored in a `.auto_updates.json` file alongside `.validation` files.
```
Trenly, line 50: It would probably be better to have a separate directory structure specifically for auto-updates instead of including metadata in the `manifests` folder. While having the data right next to all of the manifests makes it easy for users to find, it greatly increases the complexity cost of a solution since whatever is built will have to recursively filter through all the directories to find the metadata files instead of filtering to only a known sub-tree

doc/specs/#500 - Auto generate manifests for newer versions.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Automation Workflow
+
+The automation consists of a PowerShell script that:
+
+1. Loads variables specified in the `AdditionalInfo` object.
+2. Executes `Invoke-RestMethod` or `Invoke-WebRequest` with the defined arguments (supporting multiple sources for different APIs).
+3. Runs the `PostResponseScript` to process API responses.
+4. Calls `wingetcreate update` in non-interactive mode:
+   - Uses `--submit` if no additional metadata is required.
```
Trenly, line 60: In these cases, wouldn't it be better to have the ISV's integrate wingetcreate into their CI/CD pipeline? My worry would be that if an official automation is added, publishers would just do nothing and let the automation handle it instead of supporting their own applications.

While polling for a new version works at small scale, it can easily chew up a ton of bandwidth. There are currently ~8,000 package identifiers in the repository; If even 20% of them were polled, that would be 1,600 packages every few hours that need to be queried for updates, and that will only grow as the service grows. 

If the manifest is simple enough that `wingetcreate update --submit` can handle it, then it is trivial for an ISV to integrate `wingetcreate` into their CI/CD process - turning the flow from a pull to a push

doc/specs/#500 - Auto generate manifests for newer versions.md
```diff
@@ -0,0 +1,87 @@
...
+
+This PowerShell script will be executed in an Azure Pipeline on a set schedule.
+
+## UI/UX Design
+
+N/A. This change does not impact the WinGet CLI directly.
+
+## Security
+
+Security remains unaffected. PRs generated by the automation will go through the same validation pipelines and require manual moderator approval before merging.
```
Trenly, line 71: If scripts are being added to the repository, the scripts will need to be validated somehow to ensure they are safe. There are security implications to running scripts in an automated environment - for example, what if the script tries to leak some process information that could contain secrets?

Trenly, line 71: Additionally, how are the endpoints that are being called validated for security? It wouldn't be good to have an azure agent talking to a URL that is MitM'ing all the traffic over to a threat actor, just to try and get information about Package Update.

Do domains need to be tested and whitelisted? Does there need to be some form of SmartScreen trust level check?

doc/specs/#500 - Auto generate manifests for newer versions.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Compatibility
+
+No breaking changes. This automation can coexist with the existing @wingetbot installer hash updates.
+
+## Potential Issues
+
+- **GitHub API Rate Limits**: Applications distributed via GitHub Releases may face API rate limits. Encouraging project maintainers to directly update WinGet manifests (push-style automation) could help mitigate this issue.
```
Trenly, line 79: Not just GitHub API Rate Limits, but what about other sites that have rate limits such as Mozilla? If they get hit with a few hundred requests quickly enough, they could throttle the connections or toss back 403 or 429 error

Trenly, line 79: What happens if a publisher removes their API or blocks the user agent or changes the response or the ISV changes the format of the file name? How do downstream breaking changes get surfaced to the repo here? It wouldn't be good to have bad PRs being submitted, but it also wouldn't be good to fail silently

doc/specs/#500 - Auto generate manifests for newer versions.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Future Considerations
+
+- Expanding automation to support distribution methods other than GitHub Releases.
```
Trenly, line 83: Isn't this already covered by the schema above, where any endpoint can be used?

doc/specs/#500 - Auto generate manifests for newer versions.md
```diff
@@ -0,0 +1,87 @@
...
+1. Loads variables specified in the `AdditionalInfo` object.
+2. Executes `Invoke-RestMethod` or `Invoke-WebRequest` with the defined arguments (supporting multiple sources for different APIs).
+3. Runs the `PostResponseScript` to process API responses.
+4. Calls `wingetcreate update` in non-interactive mode:
+   - Uses `--submit` if no additional metadata is required.
+   - If additional metadata is needed, modifies the manifests manually before calling `wingetcreate submit`.
+
+This PowerShell script will be executed in an Azure Pipeline on a set schedule.
+
+## UI/UX Design
```
Trenly, line 65: What is the UI / UX of the PRs that are submitted using the automation? Do they appear as normal PRs? Is there something linking them to the automation run? Are there any logs that are generated?

doc/specs/#500 - Auto generate manifests for newer versions.md
```diff
@@ -0,0 +1,87 @@
...
+
+N/A. This change does not impact the WinGet CLI directly.
+
+## Security
+
+Security remains unaffected. PRs generated by the automation will go through the same validation pipelines and require manual moderator approval before merging.
+
+## Compatibility
```
Trenly, line 73: What is the expected "correctness" of the automation? One thing that comes to mind is the request to be able to push changes onto wingetbot PRs since it isn't always accurate - this would likely have the same issue where if something was inaccurate the PR would have to be either merged or closed and a second PR would have to be opened to address it (as well as potentially a third to actually fix the script)

Trenly (commented):
You have detail in here about what the schema looks like, but there isn't anything describing the function of the application itself.  

You mention that it runs on a set schedule, how often does it run? Is that something which is configurable per package? Is the application something where a package update can triggered by a user making an API call or will it only operate on the schedule? 

Not all of the fields are described in the specification either.
What does the `Skip` field do? How and when should it be set? Is `Reason` required, or optional? Is there any length constraints? Any Regex validation? Are any of the fields in the specification enums?

Another consideration - how long should scripts be? How are they validated? Is there a maximum length?  Maybe it makes more sense to have the script be a separate file in the same directory as the auto-update metadata, that way `PSScriptAnalyzer` can be run on it. Should the script be something like the format of a DSC resource where there is a `Get`, a `Set`, and a `Test` method (or something similar) where there is /a function specifying how to `Get` the URLS, a `Set` function for specifying how to call wingetcreate, and a `Test` method for checking the manifest is valid and performing any other needed updates before submission?

This is a good start, but I think it needs a lot more thought and detail around implementation.

ItzLevvie (approved):

ItzLevvie (approved):

ItzLevvie (approved):

ItzLevvie (approved):

Trenly (changes requested):
Marking as needs changes based on my comments above, so that this doesn't get accidentally merged due to Levvie's approvals

denelon:
We definitely need the schema for these files to be explicitly built and included for JSON schema validation. We're also going to need similar documentation to what we have for manifests so authors can look at these files.  There are also going to be extra concerns regarding folks submitting new files and attempting to modify them.

doc/specs/#500 - Auto generate manifests for newer versions.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Problem Statement
+
+Currently, winget-pkgs relies on community contributors to update package versions, leading to delays between a new software release and its availability in WinGet. This delay is problematic when updates address security vulnerabilities, leaving users exposed. Faster updates would make WinGet more reliable, particularly for businesses that depend on timely and verified application availability.
+
+## Solution Design
+
+The proposed automation will automatically open pull requests (PRs) for new versions of applications when a new release is detected via the software’s API.
+
+```jsonc
```
pl4nty, line 22: Instead of JSON, could the manifests use YAML like app manifests? It would improve readability and simplify multi-line strings, but need an additional PowerShell module

ItzLevvie (changes requested):
Keeping as suggested changes

https://github.com/microsoft/winget-pkgs/pull/236105#pullrequestreview-2873242305

ItzLevvie:
[Policy] Reset Feedback

stephengillie:
Merging failed with:
> Error: 
{
  "message": "Repository rule violations found\n\nAll comments must be resolved.\n\nAt least 1 approving review is required by reviewers with write access.\n\nNew changes require approval from someone other than the last pusher.\n\n",
  "documentation_url": "https://docs.github.com/rest/pulls/pulls#merge-a-pull-request",
  "status": "405"
} - Url  - Body: {"merge_method":"squash","sha":"73ddb8a4e477a03f7356442b016b786261755730"} 

(Automated response - build 1327.)

denelon:
I'm going to turn this into a draft. We need to have deeper discussions internally about this. I have lots of conflicting feedback about whether we want to take the responsibility for this automation mechanism. Once that has been decided upon, we'll either close this and the associated issue with the reason, or bring more resources to bear.

G0rocks:
Just for my two cents if I'm understanding correctly then this can already be done.
See how CalcpadCE did this after I posted this issue:
https://github.com/imartincei/CalcpadCE/issues/100

They did it with this automation:
https://github.com/vedantmgoyal9/winget-releaser

PR #236847, opened 2025-03-12
Remove version: ImaginaryInfinity.Squiid version 1.2.2 (Auto)

DuckDuckStudio:
### This PR is automatically created by [Sundry](https://github.com/DuckDuckStudio/Sundry/), please apply any changes requests directly🙏.
It returns a 404 status code in GitHub Action and has been automatically verified.

image

image

Higest-Version-Remove

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/236847)


TabulateJarl8:
As a note, anything older than the latest version is also going to return a 404 if you'd like to remove those as well, as we don't preserve old package versions. I never bothered to remove them to avoid complications when releasing a new version.

If someone could let me know if this is something we should be doing when releasing a new version, that'd be great

DuckDuckStudio:
If the old version is still available when you release a new version (the artifact has not expired), this is not necessary. This removal PR is only to clean up invalid versions in the repository (it seems that this version will no longer be valid in the future).

stephengillie (approved):

PR #230915, opened 2025-02-18
New version: KhronosGroup.VulkanSDK version 1.4.304.1

Dvd-Znf:
### Pull request has been created with [komac](https://github.com/russellbanks/Komac) v2.10.1 :rocket:

Closes #230914
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/230915)

PR #243331, opened 2025-03-28
 VulkanSDK: Add actual ARM64 installer support

MarijnS95:
When #230915 got merged to close #230914, it didn't actually add the installer for ARM64 but only used the existing `x64` one (and proceeded to use the old license links, likely copied from an older release too).

Strangely `winget` didn't complain about this and seemed to continue to install the `x64` version of both the `VCRedist` dependency and the Vulkan SDK itself (was everything being emulated?).

I could run `vkcube.exe` from the SDK just fine, yet a native Aarch64 app calling `LoadLibraryEx()` for `C:\Windows\System32\vulkan-1.dll` failed with it not being a "valid win32 application" (in my Dutch locale).

After adding and running with the correct installer, my app could once again load `vulkan-1.dll` without errors.

### DIFF
--- a/manifests/k/KhronosGroup/VulkanSDK/1.4.304.1/KhronosGroup.VulkanSDK.locale.en-US.yaml
+++ b/manifests/k/KhronosGroup/VulkanSDK/1.4.304.1/KhronosGroup.VulkanSDK.locale.en-US.yaml
@@ -11,14 +11,14 @@ PrivacyUrl: https://vulkan.lunarg.com/sdk/home#content/popup/privacy/lg
 PackageName: Vulkan SDK
 PackageUrl: https://vulkan.lunarg.com/sdk/home
 License: MIT
-LicenseUrl: https://vulkan.lunarg.com/software/license/vulkan-1.3.296.0-windows-license-summary.txt
+LicenseUrl: https://vulkan.lunarg.com/software/license/vulkan-1.4.304.1-windows-license-summary.txt
 Copyright: Copyright (c) 2015-2022 LunarG, Inc.
-CopyrightUrl: https://vulkan.lunarg.com/software/license/vulkan-1.3.296.0-windows-license-summary.txt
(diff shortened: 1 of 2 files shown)


manifests/k/KhronosGroup/VulkanSDK/1.4.304.1/KhronosGroup.VulkanSDK.locale.en-US.yaml
```diff
@@ -11,14 +11,14 @@ PrivacyUrl: https://vulkan.lunarg.com/sdk/home#content/popup/privacy/lg
 PackageName: Vulkan SDK
 PackageUrl: https://vulkan.lunarg.com/sdk/home
 License: MIT
-LicenseUrl: https://vulkan.lunarg.com/software/license/vulkan-1.3.296.0-windows-license-summary.txt
+LicenseUrl: https://vulkan.lunarg.com/software/license/vulkan-1.4.304.1-windows-license-summary.txt
 Copyright: Copyright (c) 2015-2022 LunarG, Inc.
-CopyrightUrl: https://vulkan.lunarg.com/software/license/vulkan-1.3.296.0-windows-license-summary.txt
+CopyrightUrl: https://vulkan.lunarg.com/software/license/vulkan-1.4.304.1-windows-license-summary.txt
```
MarijnS95, line 16: This file has a long list of copyrights, and I'm unsure if we should update the one with a more recent date, these all seem to be pulled from code with wild formatting.

stephengillie (approved):

manifests/k/KhronosGroup/VulkanSDK/1.4.304.1/KhronosGroup.VulkanSDK.locale.en-US.yaml
stephengillie: Hi @MarijnS95, 

Unfortunately, you're probably the most informed regarding this topic, so you would probably be the best person to make this decision. Generally, a newer copyright is better, but you would know more here than I do.

PR #246011, opened 2025-04-06
New version: Prometheus.WindowsExporter version 0.30.6

damn-good-b0t:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/246011)

### DIFF
--- a/manifests/p/Prometheus/WindowsExporter/0.30.6/Prometheus.WindowsExporter.installer.yaml
+++ b/manifests/p/Prometheus/WindowsExporter/0.30.6/Prometheus.WindowsExporter.installer.yaml
@@ -1,0 +1,52 @@
+# Created with WinGet Updater using komac v2.11.2
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.9.0.schema.json
+
+PackageIdentifier: Prometheus.WindowsExporter
+PackageVersion: 0.30.6
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+UpgradeBehavior: install
+ReleaseDate: 2025-04-06
+Installers:
+- Architecture: x64
+  InstallerType: portable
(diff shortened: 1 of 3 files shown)


stephengillie:
>  Executable C:\Program Files\windows_exporter\windows_exporter.exe returned exit code: 1
 
 | 00000001 | 1 | -4294967295 | ERROR_INVALID_FUNCTION | Incorrect function. | 


(Automated response - build 1037.)

stephengillie (approved):

jkroepke:
@stephengillie What does it mean, the Automatic Validation ended exit code 1? I'm author of the project ask if I can do something to prevent that?

Utesgui:
@jkroepke the validation tries to start the exe, but since the installation adds it as service and starts it, the default port is already in use so the program exits
image
is it possible to check if the application that is listening on the port you want to use is "you" and exit gracefully if so?
after stopping the service it asks me for the firewall rules and works once they are added:
image

jkroepke:
@Utesgui I see

We have 2 options here:

* run the msi installer with LISTEN_PORT=9183 in tests only. This ensures that the service listen on different service
* run the exe with `--web.listen-address=127.0.0.1:9183` as additional CLI parameter in tests only. This ensure that the exe is listen on a different port. Listen on 127.0.0.1 prevents the firewall warning.

Normally, the msi installer would fail, if the service wont start. However, there is a health endpoint on http://127.0.0.1:9182/health to validate if the service is up and running.

The firewall warning is indented. The exporter should bind on a public interface, because it's intended that the called by an external monitoring system.

stephengillie:
> @stephengillie What does it mean, the Automatic Validation ended exit code 1? I'm author of the project ask if I can do something to prevent that?

It's something of a known issue with the validation pipelines, which were written to expect Windows Forms, and run all EXEs without arguments. For most CLI applications, this is an erroneous condition, as arguments are expected, and so these return exit code 1, whch apparently means "Invalid function". This is so common that I routinely ignore all exit codes below 1000.

PR #253949, opened 2025-05-05
Retire "stacklok.minder" manifest

evankanderson:
The `stacklok.minder` package was renamed to `mindersec.minder` when the software was donated to the OpenSSF.  Having the old package name as well as the new one in `winget` is confusing to users, and has a chance of installing year-old software versions that will not be updated.

Retire the `stacklok/minder` package (but not the `frizbee` package, which is still owned by Stacklok).

Note: I couldn't find an official way to request removal, so I'm sending this PR, but I'm happy to follow a different process which hides or removes the older Stacklok version of the application.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/253949)


evankanderson:
@microsoft-github-policy-service agree [company="Custcodian"]

evankanderson:
@microsoft-github-policy-service agree company="Custcodian LLC"

evankanderson:
@ItzLevvie -- I think this might need some manual attention, as it's a package delete, not a new package or an update to an existing package.

SpecterShell:
You can only remove one version per PR.

PR #261270, opened 2025-05-28
Remove older portable versions of OpenRGB

Macleykun:
Checklist for Pull Requests
- [ ] Is there a linked Issue? (no)

Manifests
- [ x] Have you checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/change?
- [x ] This PR only modifies one (1) manifest (technically i touch 2 versions, but in this context it makes sense)
- [x ] Have you [validated](https://github.com/microsoft/winget-pkgs/blob/master/doc/Authoring.md#validation) your manifest locally with `winget validate --manifest <path>`? (not applicable because we're removing 2 portable versions, to favor the installer).
- [x ] Have you tested your manifest locally with `winget install --manifest <path>`? (again see previous comment)
- [x ] Does your manifest conform to the [1.10 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.10.0)? (again see previous comment)

---

Hello! As the previous PR went well and also installing goes well, i do notice even after uninstalling openrgb (and i reinstalled my windows last weekend, and only installed the installer variant). That winget keeps thinking it needs to update OpenRGB. Even tho it's already the latest version.

The previous versions are from 2023 and 2022 and are the portable versions. Because an installer is more prefered i think it's handier to delete the older once's so we don't confused the end-user when they upgrade/migrate to the installer.

Btw i did double check the version but it's really 1.0rc1. But if i additionally need to change the version to 0.9.1244. Do let me know and i'll add it to this PR, so that the current and future versions for OpenRGB with winget will garantuee a smooth install :).
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/261270)


Macleykun:
image

Picture of the current issue. Same happens if uninstalled and then freshly installed.

Macleykun:
Responding as it's intentional to remove older manifests, so we only offer the installer type, as currently there's an update loop.

Macleykun:
> Each manifest has now been removed separately with #261670 and #261671.

Thank you so much! Lovely pfp btw and have a great weekend! :)

PR #271843, opened 2025-07-07
Updated Version: FireDaemon OpenSSL 3.5.1

trueqbit:
Checklist for Pull Requests

Manifests
- [x] Does your manifest conform to the [1.10 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.10.0)?

Unfortunately SpecterShell's PR #271549 doesn't align with the vendor's manifest:
1. FireDaemon updated the trade mark name
2. The description is different.

It would be great if other contributors would consult the vendors first.

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/271843)


ItzLevvie (approved):

trueqbit:
> @trueqbit it looks like this is removing a bunch of expected return codes. Is that intentional?

@denelon Yes, the removal is intentional. First we never had them listed. Additionally I don't understand why standard return codes of an MSI-based installer should be listed in a manifest. This is very cumbersome to maintain, and the list is incomplete.

denelon:
Standard MSI codes shouldn't be strictly necessary. In general, they (return codes in the manifest) were added for executables with arbitrary return values. I didn't dig too deeply; I just did a cursory view of the manifest and saw a bunch of return codes being removed. 

We do "prefer" publishers to update their own manifests when new versions are released. That's one of the main reasons we built wingetcreate. The idea was to eliminate as much friction as possible so publishers could easily maintain their packages. I'm assuming you're affiliated with this publisher/package. Is this submission likely to be automated in the future?

trueqbit:
I am affiliated with FireDaemon - as a software engineer and also maintainer of distributions.

I have automated submission in the back of my mind, but it hasn't been an issue yet.
However, I find it difficult to automate because `wingetcreate` changes the manifest when I update it, so I have to post-process the resulting files manually before committing.

denelon:
@trueqbit, I just now got the e-mail notification for your comments.

Are there already issues filed for these over at https://github.com/microsoft/winget-create ?

PR #302619, opened 2025-10-11
Update version: Microsoft.GameInput version 3.0.26100.6154

LinXP:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/302619)


KarbitsCode:
> ##[error] Manifest Error: DisplayVersion declared in the manifest has overlap with existing DisplayVersion range in the index. Existing DisplayVersion range in index:  [ [10.1.26100.6154, 10.1.26100.6154]]

There is already other version using that DisplayVersion value, so that's why the pipeline complains https://github.com/microsoft/winget-pkgs/labels/Manifest-AppsAndFeaturesVersion-Error

Also, the policy bot keeps removing any error related labels recently, that's odd

PR #321647, opened 2025-12-10
Hashes in SQLite manifest are incorrect.

bucweat:
Hashes for SQLite 03.51.01 are not correct. You cannot use the hashes from https://sqlite.org/download.html as they are SHA3-256. I downloaded the files and calculated the SHA256 locally and updated the file.  
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/321647)


bucweat:
> If hashes for 03.51.01 are not correct, then why are you modifying 03.51.00 manifest?

I thought I was modifying the 03.51.01 manifest...if not then I'm sorry. I will attempt to fix and resubmit the push request.

bucweat:
Closing at this is 03.51.00 issue is with 03.51.01.

PR #326326, opened 2025-12-26
New package: NUnit.NUnit version 4.4.0

### LINKED ISSUE #333521: [Package Request]: NUnit.NUnit
### How can we help?

I would like help so I can submit the manifest.

### Please read and ensure the following

- [x] The installer meets the above requirements

### Please provide the following information

Download Page Url: https://github.com/nunit/nunit
Publisher: NUnit
Package Name: NUnit
Description: A unit-testing framework for all .NET languages. It can be used for a wide range of testing, from unit testing with TDD to full fledged system and integration testing.
Package Version: 4.4.0
Installer Url: https://github.com/nunit/nunit/releases/download/4.4.0/NUnit.Framework-4.4.0.zip
———————————————
Notes from me:
* I grew tired of seeing so many of my PRs never get processed (Even more so if they didn't get processed before the installers were removed upstream, as with many KDE programs), so I'm making issue reports for a handful of them now so I can check off the "Is there a linked Issue?" box in the PRs.

DandelionSprout:
Checklist for Pull Requests
- [x] Is there a linked Issue? Fix #333521.

Manifests
- [x] Does your manifest conform to the [1.10 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.10.0)?

---
Notes from me:
* Nuget's package search site claims that NUnit has had 559,683,719 all-time downloads of it. I allow myself to doubt the validity of Nuget's download statistics (Some packages there were listed with ≥1,000,000,000 downloads), but at the very least it indicates that Nunit is reasonably popular among Nuget users.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/326326)

### DIFF
--- a/manifests/n/NUnit/NUnit/4.4.0/NUnit.NUnit.installer.yaml
+++ b/manifests/n/NUnit/NUnit/4.4.0/NUnit.NUnit.installer.yaml
@@ -13,16 +13,14 @@ Installers:
   NestedInstallerFiles:
   - RelativeFilePath: bin\net8.0\nunitlite-runner.exe
     PortableCommandAlias: nunitlite-runner
-  - RelativeFilePath: bin\net6.0\nunitlite-runner.exe
-    PortableCommandAlias: nunitlite-runner6
-  - RelativeFilePath: bin\net462\nunitlite-runner.exe
-    PortableCommandAlias: nunitlite-runner4
+  Dependencies:
+    PackageDependencies:
+    - PackageIdentifier: Microsoft.DotNet.Runtime.8
 - Architecture: x86
   InstallerUrl: https://github.com/nunit/nunit/releases/download/4.4.0/NUnit.Framework-4.4.0.zip
   InstallerSha256: 6172949E6D3E5F5EAFB7D06A603C7132145BFA852669E579E4FCB6248D75736F
   NestedInstallerFiles:
   - RelativeFilePath: bin\net462\nunitlite-runner.exe
     PortableCommandAlias: nunitlite-runner
-ElevationRequirement: elevationRequired
 ManifestType: installer
 ManifestVersion: 1.12.0
--- a/manifests/n/NUnit/NUnit/4.4.0/NUnit.NUnit.installer.yaml
+++ b/manifests/n/NUnit/NUnit/4.4.0/NUnit.NUnit.installer.yaml
@@ -11,16 +11,16 @@ Installers:
   InstallerSha256: 6172949E6D3E5F5EAFB7D06A603C7132145BFA852669E579E4FCB6248D75736F
   NestedInstallerFiles:
   - RelativeFilePath: bin\net8.0\nunitlite-runner.exe
-    PortableCommandAlias: nunit
+    PortableCommandAlias: nunitlite-runner
   - RelativeFilePath: bin\net6.0\nunitlite-runner.exe
-    PortableCommandAlias: nunitnet6
+    PortableCommandAlias: nunitlite-runner6
   - RelativeFilePath: bin\net462\nunitlite-runner.exe
-    PortableCommandAlias: nunitnet4
+    PortableCommandAlias: nunitlite-runner4
 - Architecture: x86
   InstallerUrl: https://github.com/nunit/nunit/releases/download/4.4.0/NUnit.Framework-4.4.0.zip
   InstallerSha256: 6172949E6D3E5F5EAFB7D06A603C7132145BFA852669E579E4FCB6248D75736F
   NestedInstallerFiles:
   - RelativeFilePath: bin\net462\nunitlite-runner.exe
-    PortableCommandAlias: nunitnet4
+    PortableCommandAlias: nunitlite-runner4
 ManifestType: installer
 ManifestVersion: 1.12.0
--- a/manifests/n/NUnit/NUnit/4.4.0/NUnit.NUnit.installer.yaml
+++ b/manifests/n/NUnit/NUnit/4.4.0/NUnit.NUnit.installer.yaml
@@ -10,17 +10,17 @@ Installers:
   InstallerUrl: https://github.com/nunit/nunit/releases/download/4.4.0/NUnit.Framework-4.4.0.zip
   InstallerSha256: 6172949E6D3E5F5EAFB7D06A603C7132145BFA852669E579E4FCB6248D75736F
   NestedInstallerFiles:
-  - RelativeFilePath: "bin\net8.0\nunitlite-runner.exe"
+  - RelativeFilePath: bin\net8.0\nunitlite-runner.exe
     PortableCommandAlias: nunit
-  - RelativeFilePath: "bin\net6.0\nunitlite-runner.exe"
+  - RelativeFilePath: bin\net6.0\nunitlite-runner.exe
     PortableCommandAlias: nunitnet6
-  - RelativeFilePath: "bin\net462\nunitlite-runner.exe"
+  - RelativeFilePath: bin\net462\nunitlite-runner.exe
     PortableCommandAlias: nunitnet4
 - Architecture: x86
   InstallerUrl: https://github.com/nunit/nunit/releases/download/4.4.0/NUnit.Framework-4.4.0.zip
   InstallerSha256: 6172949E6D3E5F5EAFB7D06A603C7132145BFA852669E579E4FCB6248D75736F
   NestedInstallerFiles:
-  - RelativeFilePath: "bin\net462\nunitlite-runner.exe"
+  - RelativeFilePath: bin\net462\nunitlite-runner.exe
     PortableCommandAlias: nunitnet4
 ManifestType: installer
 ManifestVersion: 1.12.0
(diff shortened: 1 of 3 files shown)


DandelionSprout:
@denelon Would it be possible to look into potentially getting this PR merged? It has been marked as "Validation-No-Executables" for 19 days now, although the manifest does have executables.

manifests/n/NUnit/NUnit/4.4.0/NUnit.NUnit.installer.yaml
```diff
@@ -0,0 +1,28 @@
...
+- Architecture: x64
+  InstallerUrl: https://github.com/nunit/nunit/releases/download/4.4.0/NUnit.Framework-4.4.0.zip
+  InstallerSha256: 6172949E6D3E5F5EAFB7D06A603C7132145BFA852669E579E4FCB6248D75736F
+  NestedInstallerFiles:
+  - RelativeFilePath: bin\net8.0\nunitlite-runner.exe
+    PortableCommandAlias: nunitlite-runner
+  - RelativeFilePath: bin\net6.0\nunitlite-runner.exe
+    PortableCommandAlias: nunitlite-runner6
+  - RelativeFilePath: bin\net462\nunitlite-runner.exe
+    PortableCommandAlias: nunitlite-runner4
```
mdanish-kh: The issue is that with `ArchiveBinariesDependOnPath`, the `PortableCommandAlias` generation does not work. This is a CLI bug being tracked in:

- https://github.com/microsoft/winget-cli/issues/5105

So, when you install the package, the only referenced alias created is the file name itself `nunitlite-runner.exe` which is being taken from the `net8.0\` directory. `nunitlite-runner6` & `nunitlite-runner4` are not created.

Furthermore, the package requires DotNet to be installed, so we should have that as a dependency in the manifest...but the issue is that the different NestedInstallers require different versions of DotNet, something which we can not dynamically specify in the manifest.

Image



Because of all these issues, I suggest breaking this package down into multiple PackageIdentifiers `NUnit.NUnit.DotNet8`, `NUnit.NUnit.DotNet6` and so on. Feel free to pick a better naming scheme as I'm not sure what the traditional nomenclature is regarding this package. In each of the nested package, we should add the required version of DotNet as a dependency.

mdanish-kh (changes requested):

manifests/n/NUnit/NUnit/4.4.0/NUnit.NUnit.installer.yaml
DandelionSprout: * *"Furthermore, the package requires DotNet to be installed, so we should have that as a dependency in the manifest...but the issue is that the different NestedInstallers require different versions of DotNet, something which we can not dynamically specify in the manifest."*

Huh. Well spotted. I genuinely wasn't aware it needed a dependency outright.

I'll cook up something for the NET 8.0 version specifically.

* *"The issue is that with ArchiveBinariesDependOnPath, the PortableCommandAlias generation does not work."*

That explains a lot of the various PATH glitches I have previously run into. So now I guess I know about the cause of that too.

DandelionSprout: I've now made the x64 installer .NET 8-specific, and the x86 installer .NET 4-specific. So we'll see how this goes.

manifests/n/NUnit/NUnit/4.4.0/NUnit.NUnit.installer.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - RelativeFilePath: bin\net8.0\nunitlite-runner.exe
+    PortableCommandAlias: nunitlite-runner
+  Dependencies:
+    PackageDependencies:
+    - PackageIdentifier: Microsoft.DotNet.Runtime.8
+- Architecture: x86
+  InstallerUrl: https://github.com/nunit/nunit/releases/download/4.4.0/NUnit.Framework-4.4.0.zip
+  InstallerSha256: 6172949E6D3E5F5EAFB7D06A603C7132145BFA852669E579E4FCB6248D75736F
+  NestedInstallerFiles:
+  - RelativeFilePath: bin\net462\nunitlite-runner.exe
```
mdanish-kh, line 23: > I've now made the x64 installer .NET 8-specific, and the x86 installer .NET 4-specific. So we'll see how this goes.

As a typical WinGet user if I do something like `winget install <some-package> --architecture x64` or `winget install <some-package> --architecture x86`, I expect to logically get the "same" package, just for different instruction set/architectures. Currently, it's surprising that I get .NET4 variant in case of x86 & .NET8 variant in case of x64. And a typical user would be confused as to what decision we made for .NET6. Beyond even the `winget install` use case, some users may want to just download the packages for multiple variants with `winget download`. I think the multiple PackageIdentifier approach will make more sense for such use cases where a typical user wouldn't be expected to inspect the manifest itself to know where they could find the appropriate installers

manifests/n/NUnit/NUnit/4.4.0/NUnit.NUnit.locale.en-US.yaml
```diff
@@ -0,0 +1,36 @@
...
+PackageLocale: en-US
+Publisher: NUnit
+PackageName: NUnit 4 Framework
+PackageUrl: https://github.com/nunit/nunit
+License: MIT
+ShortDescription: A unit-testing framework for all .NET languages. It can be used for a wide range of testing, from unit testing with TDD to full fledged system and integration testing.
+Description: |-
+  A unit-testing framework for all .NET languages. It can be used for a wide range of testing, from unit testing with TDD to full fledged system and integration testing.
+
+  Note that due to the Winget package on x64 needing at least one .NET dependency for its x64 .exe-s to pass validation, the Winget package's x64 version automatically installs the .NET 8 Runtime, regardless of whether or not the user solely intends to use the .NET 6 or 4 .exe-s. The PATH shortcut will also lead to the .NET 8 .exe by default.
```
mdanish-kh, line 14: Personally, I don't think this information is apt for the `Description` field. That's supposed to be the extension of `ShortDescription` offering more information about the package itself, not how WinGet is dealing with said package. I think this information is more suited as a comment in the manifest

manifests/n/NUnit/NUnit/4.4.0/NUnit.NUnit.installer.yaml
DandelionSprout: Given I have many other PRs open at the moment already (32 at the time of writing), creating 3 separate PRs with 1 for each .exe would add to my current workload. So the options that I personally would do would be either:
* The current approach.
* One PR for the .NET 8 .exe only.
* Giving up.

manifests/n/NUnit/NUnit/4.4.0/NUnit.NUnit.locale.en-US.yaml
DandelionSprout: * *"(…), not how WinGet is dealing with said package."*

I've had people recommend me previously to put similar notes in `InstallerNotes`, but those are unfortunately only shown to most users *after* the package has already been installed. So there have been some difficult calls for me to figure out which sections to add such infos in.

manifests/n/NUnit/NUnit/4.4.0/NUnit.NUnit.installer.yaml
mdanish-kh: I understand, I'm not explicitly asking you to create those 3 PRs, but merely suggesting what the better approach may be. Given that you already have this PR open, we can make it only for .NET8

manifests/n/NUnit/NUnit/4.4.0/NUnit.NUnit.locale.en-US.yaml
mdanish-kh: Yes, this may be better suited for a PreInstallationNote, once winget-cli has support for it
- https://github.com/microsoft/winget-cli/issues/3483

Regardless, if we're splitting up this package into 3 variants, we may not even need the extended explanation

manifests/n/NUnit/NUnit/4.4.0/NUnit.NUnit.installer.yaml
DandelionSprout: I'll go with the current approach; unless a 4.4.1 or 4.5 gets released by NUnit upstream, in which case I'd go with .NET 8 if I were to make a new PR.

stephengillie:
Manual Validation ended with: 
> PS C:\Users\User\Desktop> C:\Users\User\AppData\Local\Microsoft\WinGet\Packages\NUnit.NUnit__DefaultSource\bin\net8.0\nunitlite-runner.exe --version
> NUnitLite 4.4.0 (Release)
> Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
> 
> PS C:\Users\User\Desktop> Get-ARPTable nunit
> 
> 
> DisplayName    : NUnit 4 Framework
> DisplayVersion : 4.4.0
> Publisher      : NUnit
> ProductCode    : NUnit.NUnit__DefaultSource
> Scope          : User
> 
> 
> 
> PS C:\Users\User\Desktop>

(Automated response - build 1217.)

stephengillie (approved):

PR #329779, opened 2026-01-13
New version: aelassas.Servy version 5.2

aelassas:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/329779)


stephengillie (approved):

stephengillie:
Yes, we had 2 new packages added by 2 different submitters, with different caps for the same developer's name. This wasn't caught by the pipelines as both went through, then sat for a couple of days, before I could validate them. And the pipelines are currently the only gate which checks for caps mismatches, which led to this Issue: 
- https://github.com/microsoft/winget-pkgs/issues/330103

PR #337341, opened 2026-02-07
New package: SaversPlanet.ColorMix version 2.0.0

vkryukin:
Added new freeware screensaver.
Updated to multi-file manifest format as requested.

Manifest validated locally.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/337341)


manifests/s/SaversPlanet/ColorMix/1.0.0/SaversPlanet.ColorMix.installer.yaml
stephengillie, line 7:
```suggestion
    InstallerType: portable
```

stephengillie (changes requested):
Hi @vkryukin,

This appears to be a portable package:
image

manifests/s/SaversPlanet/ColorMix/1.0.0/SaversPlanet.ColorMix.installer.yaml
vkryukin: @stephengillie Thanks for the review!

However, this is NOT a portable package.
The file at the URL is an Inno Setup Installer (.exe), not a standalone application.

It copies the .scr file to the Windows System directory.
It registers the screensaver in the registry.
It creates uninstaller entries.
If we mark it as portable, Winget will treat the setup.exe itself as the application, which is incorrect.

The correct type is Inno, and the silent switches provided (/VERYSILENT ...) are standard for Inno Setup.

Please allow me to keep InstallerType: inno.

stephengillie:
> Hello @stephengillie,
> 
> screensaver opens Windows screensaver settings dialog after installation. I can not remove it now, because if i get new code signing certificate it will not be green in Windows SmartScreen. Current old digital signature is green in SmartScreen (thousands of downloads in past years).
> 
> Can you please approve this screensaver?
> 
> Thank you!

Hi @vkryukin,

Our policy of having a fully-unattended mode is meant to make a very quiet experience for our users. If we allow this package, then it will be precedent to allow other installers without a fully-unattended mode. And quickly, we would lose one of our most-appreciated features. 

If you're not able to make this update, what will happen when the next version is released? Or has this software had the last update it will ever receive?

PR #345727, opened 2026-03-06
New version: Klocman.BulkCrapUninstaller version 6.1

Klocman:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/345727)


stephengillie (approved):

Klocman:
@knw257 Right, I did not realize it worked like that. I restored the original version and added the new one with a different filename. It is a quick fix for an issue that I believe shouldn't affect users installing through winget, so leaving the original file up is not a problem (https://github.com/Klocman/Bulk-Crap-Uninstaller/issues/864).

PR #349184, opened 2026-03-17
New version: axpnet.AeroFTP version 2.9.9

axpnet:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/349184)

PR #349729, opened 2026-03-18
New version: axpnet.AeroFTP version 3.0.2

axpnet:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/349729)


axpnet:
Hi, this is the expected STATUS_DLL_NOT_FOUND (0xC0000135) error. AeroFTP requires the Microsoft Visual C++ 2015-2022 Redistributable, which is not present in the validation sandbox.

The NSIS installer auto-downloads and installs it at runtime via NSISdl, and the manifest declares Microsoft.VCRedist.2015+.x64 as a PackageDependency. The download fails silently in the sandbox due to no internet access.

This has been the case for all previous versions and has been manually approved each time (see #349540, #349184). The application installs and runs correctly on real Windows systems.

PR #358985, opened 2026-04-12
Add Needs-Author-Feedback with Needs-CLA

Trenly:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/358985)

### DIFF
--- a/.github/policies/labelAdded.needsCLA.yml
+++ b/.github/policies/labelAdded.needsCLA.yml
@@ -16,6 +16,8 @@ configuration:
           - labelAdded:
               label: Needs-CLA
         then:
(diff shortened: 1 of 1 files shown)


ItzLevvie (approved):

Trenly:
After further thought - I don't think this is a valuable change; while it does ensure that old PRs that need CLA would be closed automatically, it would mean that they wouldn't auto-merge after a user agrees to the CLA.

Since users are tagged anyways on the Needs-CLA message, they should already be aware of the need to provide feedback

PR #357996, opened 2026-04-11
New package: Mozilla.MaintenanceService version 149.0.2rc1

DandelionSprout:
Checklist for Pull Requests
- [x] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)?
- [ ] Is there a linked Issue?  If so, fill in the Issue number below.
   
  - Resolves #[Issue Number]

Manifests
- [x] Have you checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/change?
- [x] This PR only modifies one (1) manifest
- [x] Have you [validated](https://github.com/microsoft/winget-pkgs/blob/master/doc/Authoring.md#validation) your manifest locally with `winget validate --manifest <path>`?
- [x] Have you tested your manifest locally with `winget install --manifest <path>`?
- [x] Does your manifest conform to the [1.12 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.12.0)?

Note: `<path>` is the directory's name containing the manifest you're submitting.

---
Notes from me:
* Last time I submitted (and successfully merged) a `Mozilla.MaintenanceService` package last autumn, I got absolutely hammered by "Security first" end-users. But this time the installer comes from a 100× more stable source.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/357996)

PR #361091, opened 2026-04-15
More accurate estimate of its version number + Implemented web feedback: Mozilla.MaintenanceService version 94.0

DandelionSprout:
Checklist for Pull Requests
- [ ] Is there a linked Issue?  If so, fill in the Issue number below.
   
  - Resolves #[Issue Number]

Manifests

---
Notes from me:
* Pretty much everyone on the planet (Me included, just to have said that) have been completely confused as to how Mozilla Maintenance Service is versioned, as Mozilla's documentation of it is non-existent. Thanks to another Winget package, `MiTeC.ExeExplorer`, I could tell that at least 2 versions of Maintenance Service exist: One from 2016, and one from 2021. As such, I couldn't number it as `1` after all, and instead numbered it for the Firefox version it was first bundled with that I could tell.
* The various changes first attempted in #361084 are carried over into this PR, with the (current?) exception of the "Not intended for upgrades" specifyings, as people seeing `Mozilla.MaintenanceService` when running `winget update` will be drastically reduced.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/361091)

### DIFF
--- a/manifests/m/Mozilla/MaintenanceService/94.0/Mozilla.MaintenanceService.installer.yaml
+++ b/manifests/m/Mozilla/MaintenanceService/94.0/Mozilla.MaintenanceService.installer.yaml
@@ -1,0 +1,16 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: Mozilla.MaintenanceService
+PackageVersion: "94.0" # The earliest Firefox version known to contain the 2021-07-24 Maintenance Service timestamp version.
+ReleaseDate: 2021-11-02
+InstallerType: zip
+NestedInstallerType: nullsoft
+NestedInstallerFiles:
+- RelativeFilePath: firefox\maintenanceservice_installer.exe
+Installers:
+- Architecture: neutral
+  InstallerUrl: https://archive.mozilla.org/pub/firefox/candidates/149.0.2-candidates/build1/win32/en-US/firefox-149.0.2.zip
+  InstallerSha256: d25b7ceb295fb5c2b9d48e1c3b52d7243a3f16542636c42da8148e05741d6734
+RequireExplicitUpgrade: true
+ManifestType: installer
+ManifestVersion: 1.12.0
--- a/manifests/m/Mozilla/MaintenanceService/94.0/Mozilla.MaintenanceService.locale.en-US.yaml
+++ b/manifests/m/Mozilla/MaintenanceService/94.0/Mozilla.MaintenanceService.locale.en-US.yaml
@@ -1,0 +1,33 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.12.0.schema.json
+
+PackageIdentifier: Mozilla.MaintenanceService
+PackageVersion: "94.0" # The earliest Firefox version known to contain the 2021-07-24 Maintenance Service timestamp version.
+PackageLocale: en-US
+Publisher: Mozilla
+PackageName: Mozilla Maintenance Service (Standalone)
+PackageUrl: https://archive.mozilla.org/pub/firefox/candidates/
+License: Unclear, presumably MPL-2.0
+ShortDescription: Standalone installer for Mozilla Maintenance Service, a tool used by Mozilla to more easily update Firefox and Thunderbird installations.
+Description: |-
+  Standalone installer for Mozilla Maintenance Service, a tool used by Mozilla to more easily update Firefox and Thunderbird installations.
+
+  The installer comes from Mozilla's well-hid but publicly available "Release Candidates" builds of Firefox Stable, as official Firefox portable packages published by Mozilla exist for those.
+
+  The package is mostly intended for anyone wanting re-installs after having previously uninstalled Maintenance Service, or who use very old Firefox versions. Further oddities are listed in Release Notes below.
+ReleaseNotes: |-
+  Due to the total lack of Mozilla changelogs for Maintenance Service, and in fact that it's unclear if it's open source or not since the location of its source code is unknown, figuring out its version number relies on qualified guesses. There are however at least 2 versions known to exist (1 from 2016 and 1 from 2021), of which this version is the newest of them.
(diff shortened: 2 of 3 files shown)


DandelionSprout:
Both of those pipeline notes are intentional on my end, especially as I'll make sure to delete the `149.0.2rc1` manifest after (but at no point before) this `94.0` manifest gets merged.

seanbright:
As pointed out in the original PR, there is no reason to have this be separately installable from Firefox/Thunderbird. I don't know policy here (just an annoyed end-user), but can you just remove the maintenance service from the winget repository entirely?

DandelionSprout:
To date I have never heard of a package that was considered prohibited on Winget-pkgs because a subset of Winget users didn't like the package.

(Asking an admin for a second opinion on it:) @denelon Surely end-user opinions aren't a blockage against a package being in Winget-pkgs?

denelon:
I don't know enough about Mozilla Maintenance Service to have much of an informed opinion myself.

Generally, folks may add packages to the community repository as long as no terms of service or use violations occur.

At the end of the day, however, we intend to honor the wishes of the publisher. It is their product, and their customer experience using WinGet.

Without knowing the implications of adding this package to the community repository it's hard for me to have a concrete answer.
* Does it create a broken install/upgrade/list experience?
* What is the value in adding it?
* Does it cause confusion for users?

DandelionSprout:
From my perspective:

* *"I don't know enough about Mozilla Maintenance Service to have much of an informed opinion myself."*

Pretty much no one do (Me included), so you're not alone in that.

* *"folks may add packages to the community repository as long as no terms of service or use violations occur."*

Indeed how I see it.

* *"Does it create a broken install/upgrade/list experience?"*

The current manifest with the version number 149.0.2rc1 is somewhat broken. That particular version had been submitted before I learned more about Maintenance Service (which Mozilla themselves aren't keen to teach people about), so it shows up in `winget update` as an update to the vast majority of Firefox (non-MSIX) users. However, the app isn't broken in its functionality, as the update will install the same .exe of Maintenance Update as the one that 99% of users already have, but with a higher version number.

It is my current belief that this 94.0 manifest will not be affected by it, also as Firefox ≤93 shipped with an older version of Maintenance Service.

* *"What is the value in adding it?"*

The logic behind it is mostly for the following cases:
1) Someone installs `Mozilla.Firefox`. Mozilla's installer installs Mozilla Maintenance Service as a separate app by default, especially unless some extra installer parameters (if any such exist) get added by the truckload to Spectopo's Firefox PR scripting.
2) A user uninstalls Mozilla Maintenance Service as they are unsure what it does, while keeping Firefox installed.
3) The user then decides they want to re-install Mozilla Maintenance Service, to complement their still-installed Firefox.
4) There are no widely known ways online to re-install only Maintenance Service without having to run an upgrade or in-place upgrade of Firefox. Trust me, I've searched a lot online for that in the past 2 years. Firefox does however include a `maintenanceservice_installer.exe` post-installation, but Mozilla's lack of documentation of it makes it unsure if they are new versions or not (They usually aren't, unless they update directly from Firefox 85 or such).
5) That's where this package comes into play, which is the only currently broadly documented way to do a separate install of only Maintenance Service, which I am confident that surely there must be a user demographic for out there, even if not universally?

* *"Does it cause confusion for users?"*

Yes. This too I chalk up to Mozilla's non-existent documentation of what it is and does. Recent cases:
* https://github.com/microsoft/winget-pkgs/pull/357996
* https://new.reddit.com/r/firefox/comments/1sm28m4/how_do_i_update_mozilla_maintenance_services_mms/

vitaminj:
As I understand it, step 3 above should be "Firefox, when detecting the next update will prompt the user to update it differently because the service isn't present, and in doing so will get the service re-installed"

Also the confusion caused is only by OP unilaterally adding this package to winget for a second time, after the first time also caused confusion and potential problems.

I think @cpdupuis commented previously from Mozilla, or could perhaps arrange for a representative to give an official verdict?

As I understand it the maintenance service is only there to make an update more seamless (removing the need to ask for elevation) when one is triggered by the host app, it doesn't run or do anything by itself until asked, so reinstalling it if it has been manually removed doesn't win a user anything.

seanbright:
> To date I have never heard of a package that was considered prohibited on Winget-pkgs because a subset of Winget users didn't like the package.

It has nothing to do with not liking it. It adds no value and only serves to confuse users.

There is also an installer for it in the Firefox and Thunderbird install directories if a user uninstalls it in error.

I’ll stay out of it from now on but it’s OK to admit a mistake was made and just undo it.

iDolmatov:
There is no obvious need to have this entry in the winget database.
As already stated, this is not a standalone product, so installing it without the target product will only cause misunderstanding among users and clutter the system.
If the user has removed a component and wants to restore it, it is enough to run "maintenanceservice_installer" from the directory of the main program (such as Firefox).
In the release version of Firefox, the registry entry is numbered 149.0.2. This means that installing a different version has the potential to break component compatibility. For example, if for some reason the user failed to update the main program (the browser) and then "reinstalled" a child component from a third-party source.
According to the reference, even this method is not recommended, although it is acceptable.

https://support.mozilla.org/en-US/kb/managing-firefox-updates

> Reinstall the Mozilla Maintenance Service
> If the service is missing or corrupted, it can be reinstalled using the Firefox installer:
>
> Download the Firefox installer matching your deployment (ESR or standard release).
> Run the installer as Administrator.
> Ensure that Install Mozilla Maintenance Service is selected.
> After installation, recheck the service status.

___

> [Moderator](https://support.mozilla.org/en-US/user/cor-el/)
> [9/22/18, 6:35 AM](https://support.mozilla.org/en-US/questions/1234880#answer-1157484)
> The MMS service (maintenanceservice.exe) is installed in its own folder (%PROGRAMFILES%\Mozilla Maintenance Service).
> Once you have grated UAC access for this application when you first installed it, no further approval is necessary.

vitaminj:
Mozilla dev @cpdupuis has filed #361213 - this PR should not be landed

denelon:
Closing based on publisher request:
* https://github.com/microsoft/winget-pkgs/issues/361213

PR #366138, opened 2026-04-28
Improve the Microsoft.Office package, part 1/2.

DandelionSprout:
Checklist for Pull Requests
- [ ] Is there a linked Issue?  If so, fill in the Issue number below.
   
  - Resolves #[Issue Number]

Manifests

---
Notes from me:
* This PR primarily aims to accomplish 2 things:
* * Exclude `Microsoft.Office` from also installing OneDrive and Teams, since those have standalone installers in Winget.
* * Set up distinctly separate 64- and 32-bit installation modes.
* In order to accomplish this, the package would need to have its configuration file changed, and at least I lack write permissions for https://res.cdn.office.net/aacux/cloudconfigs/wingeta.xml. So presumably the best solution long-term would be for Winget to host its own Office configuration files.
* Part 2/2 would be to change the installation parameters in the `Microsoft.Office` manifest from `/configure https://aka.ms/fhlwingetconfig` to approximately (and for the by-then-new x64 InstallerUrl) `/configure https://raw.githubusercontent.com/microsoft/winget-pkgs/refs/heads/master/Tools/ConfigurationFiles/Microsoft/Office/StandardSet/Config64.xml`

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/366138)

### DIFF
--- a/Tools/ConfigurationFiles/Microsoft/Office/StandardSet/Config32.xml
+++ b/Tools/ConfigurationFiles/Microsoft/Office/StandardSet/Config32.xml
@@ -1,0 +1,30 @@
+<Configuration ID="1521D833-CF38-49D7-8B2A-A65E95EAA630">
+  <Add OfficeClientEdition="32">
+    <Product ID="O365ProPlusRetail">
+      <Language ID="MatchOS"/>
+      <Language ID="MatchPreviousMSI"/>
+      <ExcludeApp ID="Groove"/>
+      <ExcludeApp ID="Lync"/>
+      <ExcludeApp ID="OneDrive"/>
+      <ExcludeApp ID="Teams"/>
+    </Product>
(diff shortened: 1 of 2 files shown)


Trenly (commented):
It feels like the Office team should host these. Adding configuration files to WinGet's repository seems like it could set an unwanted precedent, but I'll let @denelon review - especially since this is a MSFT owned package

denelon:
We have a work item to handle supporting package configuration files, but as @Trenly stated, we wouldn't host them here in this repository. They should be hosted by the publisher and match the publisher's intent for the "default" or "base" configuration(s).

DandelionSprout:
In that case, what should we do if we know about things that the publishers don't? In this case, how to prevent the installer from also installing OneDrive/Teams.

PR #366566, opened 2026-04-29
Add tsuchim.ImeKeysForUS version 0.1.3

tsuchim:
Adds tsuchim.ImeKeysForUS version 0.1.3.

### DIFF
--- a/manifests/t/tsuchim/ImeKeysForUS/0.1.0/tsuchim.ImeKeysForUS.installer.yaml
+++ b/manifests/t/tsuchim/ImeKeysForUS/0.1.0/tsuchim.ImeKeysForUS.installer.yaml
@@ -1,0 +1,17 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.9.0.schema.json
+PackageIdentifier: tsuchim.ImeKeysForUS
+PackageVersion: 0.1.0
+InstallerType: wix
+Scope: machine
+InstallModes:
+  - interactive
+  - silent
+  - silentWithProgress
+UpgradeBehavior: install
+ReleaseDate: 2026-04-29
+Installers:
+  - Architecture: x64
+    InstallerUrl: https://github.com/tsuchim/ime-keys-for-us/releases/download/v0.1.0/IME-Keys-for-US-0.1.0-x64-unsigned.msi
+    InstallerSha256: 523D9C9233BF75AE9D92B3949A786639B1B3FFE45C2170F83D8B7D73A5E5057D
+ManifestType: installer
+ManifestVersion: 1.9.0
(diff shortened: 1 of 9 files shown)


stephengillie:
> 
> Executable C:\Program Files\ImeKeysForUS\ime-keys-for-us.exe returned exit code: -1073741515 
 
  | C0000135 | -1073741515 | 3221225781 | STATUS_DLL_NOT_FOUND | Possibly due to a missing dependency. (The most common dependency is `Microsoft.VCRedist.2015+.x64` / `.x86`) | 


(Automated response - build 1280.)

tsuchim:

tsuchim:
Added Microsoft.VCRedist.2015+.x64 as a package dependency because validation reported STATUS_DLL_NOT_FOUND for the installed executable in the clean validation environment. Keeping the app dynamically linked for v0.1.0.

tsuchim:
Updated the submission to use the v0.1.2 locally Authenticode-signed MSI. The executable inside the MSI and the MSI itself are signed with the local CN=tsuchim code-signing certificate. The VC Runtime dependency remains because the app is dynamically linked.

stephengillie:
Hi @tsuchim,

The application seems to have installed normally, but didn't seem to run when launched. No tray icon, and no response to alt button presses. Is this expected? 

Verify while here:
image

tsuchim:
Thank you for testing. No, missing tray icon is not expected.

I updated the submission to v0.1.3 to address this more directly.

Changes in v0.1.3:
- The MSVC runtime is now statically linked, so the Visual C++ Redistributable is no longer required.
- Startup diagnostics were added.
- Startup stages are logged to `%LOCALAPPDATA%\ImeKeysForUS\ime-keys-for-us.log`.
- Fatal initialization failures now show a visible message box.
- If the tray icon cannot be added, the app also shows a warning instead of silently appearing to do nothing.
- The README now explains that this is a tray utility for Japanese IME users on US keyboards, and that systems without a Japanese IME may not show visible IME changes even though the tray icon should still appear.

The PR now points to the v0.1.3 signed MSI:
https://github.com/tsuchim/ime-keys-for-us/releases/download/v0.1.3/IME-Keys-for-US-0.1.3-x64.msi

stephengillie:
Thank you for adding the log - I was able to use it to confirm that it was running. (A common bug is that the scan step might sometimes remove the tray app without fully closing the EXE. Restarting the EXE fixed it here.)

Verify:
image
image

Edit: Not seeing a cause for the [Policy 2.7](https://learn.microsoft.com/en-us/windows/package-manager/package/windows-package-manager-policies#27-adult-content) check.

stephengillie (approved):

PR #367350, opened 2026-05-01
New version: Argente.DuplicateFinder version 2.0.3.3

voyager2saturn:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/367350)


voyager2saturn (commented):
Validated

voyager2saturn:
I have removed the old version. Let me know if I have to create a separate PR in this case.

stephengillie:
Hi @voyager2saturn,

> I have removed the old version. Let me know if I have to create a separate PR in this case.

Yes, it will need to be in a separate PR. Ideally, we would merge the new version before removing the old version, to keep that data in the repo. 

Edit: Looking at this again, there are really 3 versions involved:
- 2.0.3.1 removal (upper manifest)
- 2.0.3.2 move from (lower manifest)
- 2.0.3.3 move to (lower manifest)

image


 If the upper manifest removal is reverted or moved to another PR, both PRs should proceed past this check.

voyager2saturn:
Thanks @stephengillie, I will do as per the guidelines.

PR #367586, opened 2026-05-01
Make all triggers case insensitive and non-capturing

Trenly:
Checklist for Pull Requests

This PR updates the policy bot rules to minorly simplify the triggers used to apply labels to packages and perform other tasks. While this PR has minimal effect on the actual triggers themselves, it does a few things.

First, it makes all of the triggers fully case-insensitive. This helps to ensure triggers are correctly interpreted even with casing typos. For example, it used to be that `duplicate of #<id>` would not trigger because the `D` was not capitalized, or that `[Policy] Needs ATtention` would not trigger because the  `T` was capitalized. This should make it a little easier for moderators to no longer have to worry about being so precise with the casing of the triggers. 

Second, it adds the non-capturing flag. While most of the expressions do not use capture groups, a few of them do. Since the capturing groups aren't needed for anything other than qualifying a group for quantification. This very minorly reduces the compute load by reducing the need for backtracking and storing the capture group. While minor, across several issues all at the same time, it should save a few fractions of a second.

Finally, it completely re-writes the trigger for opening or closing a PR so that the reason can include newlines, hyperlinks, emojis, and basically anything through the use of the `s` flag, which should treat the comment to be treated as a single-line, allowing `.+` to match with newline characters as part of the match.

---

@denelon @mdanish-kh

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/367586)

### DIFF
--- a/.github/policies/labelManagement.issueOpened.yml
+++ b/.github/policies/labelManagement.issueOpened.yml
@@ -154,7 +154,7 @@ configuration:
           - isAction:
               action: Opened
           - bodyContains:
-              pattern: (?i)hash[\s-](mismatch|does\snot\smatch)
+              pattern: (?in)hash[\s-](mismatch|does\snot\smatch)
               isRegex: True
         then:
           - addLabel:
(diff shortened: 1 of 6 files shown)


denelon (approved):

Trenly:
[PoLiCY] NeTwOrk-BlOCKer

Trenly:
It works :)

PR #378270, opened 2026-05-22
Reduce Number of Synced Labels

Trenly:
## 📖 Description
This PR reduces the number of labels that get synced from an issue to a PR when the PR is linked to the issue. This addresses a concern / confusion raised in #329218 and obsoletes #333533 (as without the syncing, the labels can only be applied by the pipeline results). 

## ✅ Checklist

  - Obsoletes #333533

@denelon
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/378270)

### DIFF
--- a/.github/policies/labelManagement.issueUpdated.yml
+++ b/.github/policies/labelManagement.issueUpdated.yml
@@ -144,16 +144,6 @@ configuration:
         then:
           - labelSync:
               pattern: Blocking-Issue
-          - labelSync:
-              pattern: Hardware
-          - labelSync:
-              pattern: Area-External
-          - labelSync:
-              pattern: Installer-Issue
-          - labelSync:
-              pattern: Interactive-Only-Installer
-          - inPrLabel:
-              label: In-PR
       - description: Remove "Help-Wanted" and "Needs-Triage" labels when an issue goes into PR
         if:
           - payloadType: Issues


denelon (approved):

mdanish-kh:
Personal take: I'd only remove `Interactive-Only-Installer` since I believe the pipeline can, in most cases, detect if an installer requires interactivity.

- `Hardware`: It's possible that the requirement is fulfilled on the VM which runs the validation, but may not cover a big portion of existing users. Say maybe a driver for AMD / NVIDIA GPU? I'd want https://github.com/microsoft/winget-pkgs/labels/Hardware to remind moderators to take another look at the package
- `Area-External` / `Installer-Issue`:  Similar take, I think we use these labels as a catch-all for any particular issue, and to remind us to take another look at the issue is valuable. I think we use these labels sparingly anyway so it shouldn't be that big of a pain

mdanish-kh:
My reviews are mostly not on-time, 100% of the time....

Trenly:
> My reviews are mostly not on-time, 100% of the time....

I went nuclear figuring we can always add back :)

Trenly:
> * `Hardware`: It's possible that the requirement is fulfilled on the VM which runs the validation, but may not cover a big portion of existing users. Say maybe a driver for AMD / NVIDIA GPU? I'd want https://github.com/microsoft/winget-pkgs/labels/Hardware to remind moderators to take another look at the package

Hardware is a dangerous one, since it will automatically close the PR or issue with a comment that support for custom hardware is not planned. So, in theory, it should never even be possible for that to be synced.

> * `Area-External` / `Installer-Issue`:  Similar take, I think we use these labels as a catch-all for any particular issue, and to remind us to take another look at the issue is valuable. I think we use these labels sparingly anyway so it shouldn't be that big of a pain

I was on the fence about these, since it is true that we use them for a variety of reasons. I'm not opposed to adding them back, but I also saw there were only 11 PRs with https://github.com/microsoft/winget-pkgs/labels/Installer-Issue.

Another point to consider is that if the user is linking an issue, it generally means they should have looked at the issue (since how else would they know to link it?).  The contributions to the repo have changed since the syncing was first added, so I'm also questioning if it's valuable at all - since if there is an issue, the pipelines will generally catch it

PR #379541, opened 2026-05-26
New version: Nehoray.TranslationManager version 1.1.0

nehorayc04:
Initial submission.

Project: https://hebrew-translation-hub.com/
GitHub: https://github.com/nehorayc04/translation-launcher
SHA-256 verified against the live release asset.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/379541)

### DIFF
--- a/manifests/n/Nehoray/TranslationManager/1.1.0/Nehoray.TranslationManager.installer.yaml
+++ b/manifests/n/Nehoray/TranslationManager/1.1.0/Nehoray.TranslationManager.installer.yaml
@@ -14,11 +14,19 @@ InstallModes:
   - silent
   - silentWithProgress
 InstallerSwitches:
-  # WinGet's default for InstallerType: inno is already
-  # `/VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP-`; we set them
-  # explicitly so a future schema default change can't surprise us.
-  # /SP- skips the "This will install..." pre-install dialog.
-  Silent: /VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP-
+  # `/SILENT` (not /VERYSILENT) is deliberate. Inno Setup 6.7's
+  # RedirectionGuard handoff (level 2 SL5 → level 3 protected Setup)
+  # fails silently with /VERYSILENT in sandboxed contexts — the
+  # protected temp dir gets created with only `_setup64.tmp` + an
+  # empty `_isetup\` and the install never proceeds, leaving the
+  # install dir half-populated. Same root cause as the failure mode
+  # we hit during the launcher's in-app self-update. /SILENT keeps
+  # the small progress dialog visible, which is what RG's protected
+  # handoff needs to complete; the dialog is dismissed automatically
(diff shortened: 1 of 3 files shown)


nehorayc04:
***@***.*** agree

בתאריך יום ג׳, 26 במאי 2026, 12:13, מאת
microsoft-github-policy-service[bot] ***@***.***>:

nehorayc04:

nehorayc04:
Fixed in [054ea78](https://github.com/microsoft/winget-pkgs/pull/379541/commits/054ea7849f) — switched the `Silent` switches from `/VERYSILENT` to `/SILENT`.

Root cause: this installer is built with Inno Setup 6.7.2, which has [RedirectionGuard](https://jrsoftware.org/ishelp/topic_redirectionguard.htm) enabled in *enforcing* mode by default. When the installer is invoked with `/VERYSILENT` from certain elevated/sandboxed parent contexts, the RG handoff from SL5 (level 2) to the protected Setup process (level 3) aborts silently — the protected temp dir gets created with only `_setup64.tmp` and an empty `_isetup\`, no files reach `{app}`, and the installer exits with no useful log. We hit the exact same failure mode in the launcher's own in-app self-update flow and `/SILENT` is the proven workaround (RG's protected handoff completes when a small progress UI thread is alive).

User-facing impact: a small Inno Setup progress dialog appears briefly during install. No clicks required — `/SUPPRESSMSGBOXES /NORESTART /SP-` keep all other prompts suppressed.

Please re-run the validation when convenient.

nehorayc04:
Additional fixes pushed in [3dfee37](https://github.com/microsoft/winget-pkgs/pull/379541/commits/3dfee3730c):

Looking at the run #3 pipeline log shared with me, the installation phase itself reported `Status: Completed` (the `/SILENT` switch resolved the Inno 6.7 RedirectionGuard handoff issue from run #2). The failure was in the post-install verification step. Two changes meant to address that:

1. **Explicit `AppsAndFeaturesEntries`** — declares the exact `DisplayName` / `DisplayVersion` / `Publisher` / `ProductCode` the verification step should expect to find under `HKLM\Software\Microsoft\Windows\CurrentVersion\Uninstall\{B0D4F2A7-3CCE-4A1A-9C44-7E1A1B6F0001}_is1`.
2. **`InstallerSuccessCodes: [1641, 3010]`** — defensively whitelists Inno's documented restart-needed codes in case `/NORESTART` doesn't downgrade them to 0 in the sandbox.

If the post-install verification continues to fail, the most likely remaining cause is the installer's `[Run]` postinstall entry (`Filename: {app}\TranslationManager.exe; Flags: nowait postinstall runasoriginaluser`) — it tries to launch the GUI launcher and may not have an interactive session in the sandbox. That's a rebuild of the installer itself (recompile installer.iss adding `skipifsilent`), which I can do if this run is still rejected.

stephengillie:
Verify:
image
Edit: Internet Translation:
image
image
Internet Translation:
image

stephengillie (approved):

PR #385080, opened 2026-06-08
Remove package: LMCP.LocalMCP

lanchuske:
### Remove package: `LMCP.LocalMCP` (all versions: 3.0.159, 3.0.171)

LMCP is now a **macOS-only** product. The Windows build is no longer offered to new users — Windows users are directed to a waitlist on the publisher's site (local-mcp.com).

As the publisher/owner of this package (Magno Partners Consulting), I'm delisting the Windows package from winget so users don't install an unsupported build. Existing installations are unaffected.

- [x] I am the publisher/owner of this package, or have permission to remove it.
- [x] This is a package **removal** (deletes all manifest versions).

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/385080)


Trenly:
Each version requires it's own separate PR for removal. The `.validation` file also requires it's own separate PR after all the individual versions have been removed

###### Close with reason: PR spans multiple versions;

PR #387756, opened 2026-06-14
Remove version: nukeop.nuclear version 1.40.0

nukeop:


stephengillie:
Hi @nukeop,

(Deterministic automation - build 1385.)

stephengillie:
Close with reason: Duplicate of #387833;

PR #392201, opened 2026-06-23
Delete mis-versioned Xbox Live developer tools package

aaron-stebner:
This is part 1 of 3 changes that I'm working on:

1. Remove the existing Xbox Live Developer Tools package - it doesn't match the versioning scheme we want to use going forward
2. Re-add the existing Xbox Live Developer Tools package with updated version + metadata
3. Add an Xbox Live Developer Tools package for the 2604 release

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/392201)


aaron-stebner:
Requesting review from @MattHyman and @dandelionsprout

DandelionSprout:
Parts 2 and 3 sound good to me. So while I'm not immediately sold on part 1 in and of itself, I have to assume that the overall 3-part plan as a whole will go well.

aaron-stebner:
> Parts 2 and 3 sound good to me. So while I'm not immediately sold on part 1 in and of itself, I have to assume that the overall 3-part plan as a whole will go well.

Thank you for taking a look!  We want to use semver for these packages going forward and want the version format to be 1.YYMM.NN, where YYNN is something like 2512 for December 2025, 2604 for April 2026. The problem is the existing package starts with major version 2025, so if we kept that package we would need all future versions to have a major version >= 2025.  We don't want the major part of the semver to reset each calendar year because the semver definition (https://semver.org/) says that a major version revision is a new incompatible API revision.

The downside of doing this is that people who have the existing package installed will be orphaned there and will need to uninstall that and then re-install the one with the new version that we'll be submitting next. There aren't that many users who will be affected though so we decided it would be OK to go forward with that.

stephengillie:
Hi @aaron-stebner,

(Deterministic automation - build 1386.)

stephengillie (approved):

PR #392302, opened 2026-06-24
fix(CharacterArc): correct x64 installer URL, add arm64 entries for v1.10.1 and v1.10.2

baoyu0:
The v1.10.1 and v1.10.2 manifests declared `Architecture: x64` but pointed to the ARM64 installer URL and SHA256. This caused x64 users to receive the ARM64 build.

Changes:
- **1.10.1**: x64 URL → `CharacterArc-1.10.1-x64-setup.exe`, add arm64 entries
- **1.10.2**: x64 URL → `CharacterArc-1.10.2-x64-setup.exe`, add arm64 entries

Verification:
- x64 SHA256 confirmed against GitHub release assets
- arm64 entries use the previously incorrect URLs (which were actually correct for arm64)
- arm64 was missing entirely, now added
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/392302)

### DIFF
--- a/manifests/u/uu201/CharacterArc/1.10.1/uu201.CharacterArc.installer.yaml
+++ b/manifests/u/uu201/CharacterArc/1.10.1/uu201.CharacterArc.installer.yaml
@@ -12,11 +12,23 @@ ReleaseDate: 2026-06-21
 Installers:
 - Architecture: x64
   Scope: user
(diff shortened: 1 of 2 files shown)


baoyu0:

PR #394476, opened 2026-06-27
New version: Amerganim.AmaderBanglaKeyboard version 0.11.0

amerganim:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/394476)


amerganim:
@wingetbot run

Fixed: this package is a single universal Inno installer that serves x64/x86/arm64 (it installs the architecture-matched DLL at runtime), so all three architectures point to the same installer — matching the published 0.10.2 manifest. Added the arm64 and x86 entries that were missing from the auto-generated manifest.

stephengillie (approved):

PR #403006, opened 2026-07-15
New package: BeamoINT.Claudex version 1.4.5

BeamoINT:
### New package submission

- Package identifier: `BeamoINT.Claudex`
- Version: `1.4.5`
- Installer: portable files from the checksum-verified official Windows release archive
- Upstream release: https://github.com/BeamoINT/Claudex/releases/tag/v1.4.5

Claudex is a cross-platform compatibility layer for using authenticated Codex/OpenAI GPT models through the Claude Code interface.


Trenly (changes requested):
Only exe files are supported as portable types

PR #407476, opened 2026-07-24
New package: MSI.MSICenter version 2.0.72 (Uses rotating hash)

DandelionSprout:
## 📖 Description
Added MSI Center.

## ✅ Checklist

  - Resolves #[Issue Number]

## 📦 Manifest Checklist

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/407476)


Trenly (commented):
Because of the existence of PackageFamilyName, it implies the exe actually installs an MSIX package - And this does seem to be the case for the Installer URL.

However, the image you provided indicates that the install was done using an MSI.

AFAIK - `AppsAndFeaturesEntries` won't solve the issue of the two different install technologies. It's kind of like the old issue that PowerShell had where the version that came pre-installed with Windows was MSIX based but the packages from GitHub were msi based and users were very frustrated that they kept getting the message about `An update is available but could not be installed because the install technologies are different`.  

I wonder if this would have the same issue?

PR #408531, opened 2026-07-27
New package: Stipple.VoiceNote version 0.1.16

Sketchjar:
### New package: Stipple.VoiceNote 0.1.14

VoiceNote is an on-device dictation and meeting-notes app for Windows. I'm the publisher.

- [x] Have you checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/addition?
- [ ] Have you tested your manifest locally with `winget install --manifest <path>`? — see note below
- [x] Does your manifest conform to the [1.9 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.9.0)?

### Notes

On the install checkbox: `winget install --manifest` needs `LocalManifestFiles`
enabled from an elevated shell, which I did not turn on. Instead I ran the exact
published installer directly with `/S` and then confirmed `winget list`
correlates the result — the evidence is below. Flagging it rather than ticking a
box I did not literally perform.

This supersedes #406308, which I closed. That submission passed the pipeline but
came back with `Validation-Executable-Error` and `Validation-Domain`. Both causes
are fixed at the source rather than worked around:

**`Validation-Executable-Error`** — the installer registered no `Publisher` in
Add/Remove Programs, so validation could not correlate the installed package.
The app's bundle config now sets a publisher, and this build writes a complete
ARP entry. Verified on a clean machine with no prior install:

```
DisplayName    : VoiceNote
Publisher      : Stipple AI
DisplayVersion : 0.1.14
InstallLocation: C:\Users\<user>\AppData\Local\VoiceNote
```

and `winget list` then correlates it:

```
Name      Id                       Version
VoiceNote ARP\User\Arm64\VoiceNote 0.1.14
```

`AppsAndFeaturesEntries` are included on both architectures as well, so the match
does not depend solely on heuristics.

**`Validation-Domain`** — the previous `InstallerUrl` pointed at the storage
bucket the binaries live in, which is not the publisher domain. Installers are
now served from `voicenote.au`, which matches `PublisherUrl`. Both URLs in this
manifest were checked end-to-end and return the exact published artifacts.

### Other details

- Per-user install (`Scope: user`), NSIS, x64 + arm64.
- The installer is not code-signed yet. Noting it explicitly since it's visible
  on download; the previous run raised no `Validation-Defender-Error` and
  installed cleanly in the sandbox.

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/408531)


stephengillie:
Verify:
image

stephengillie (approved):

PR #410515, opened 2026-07-31
New package: ChristofMueller.DeviceShelf version 1.9.6

wealthwallet:
### New package: ChristofMueller.DeviceShelf version 1.7.7

Replaces #407794, which was closed when its branch briefly held no manifests during an update. Same package, same branch, now at 1.7.7 instead of 1.6.24. The earlier revision passed validation (`Azure-Pipeline-Passed`, `Validation-Completed`).

`winget validate --manifest` was run on Windows against these exact files and returned *Manifest validation succeeded* (winget v1.29.280). The install test is unchecked because `--manifest` requires `LocalManifestFiles`, which needs an elevated session I did not have on that machine — rather than tick a box I cannot back up. The installer itself is the one the project ships and installs interactively on the same machine.

Notes for the reviewer:

- The installer is NSIS (`nullsoft`) and is **not code-signed** — a deliberate decision by the publisher, so SmartScreen may warn. winget accepts unsigned installers; flagging it so it is not mistaken for a packaging error.
- `InstallerSha256` was computed from the bytes downloaded from the URL in this manifest, not copied from a release note.
- The description states that Npcap is required and not bundled. The x64 build links `wpcap.dll` at load time, and the interactive installer prompts for it — but `winget install` runs silently, which skips that prompt, so the requirement belongs in the manifest text.


stephengillie:
Verify:
image

stephengillie (approved):

PR #412045, opened 2026-08-04
New package: StacksTech.TheSTACKSApp version 1.2.3

JJBrownHere:
### New package

- **Package**: `STACKS.TheSTACKSApp`
- **Version**: 1.2.3
- **Installer**: Inno Setup, x64, user scope

### Checklist

### What was and was not verified

Being explicit rather than ticking every box:

**Verified.** All three manifests parse, carry every field the 1.6 schema
requires for their manifest type, and agree on PackageIdentifier and
PackageVersion. The InstallerUrl and InstallerSha256 were checked against the
live artifact immediately before opening this PR: the URL returns HTTP 200 and
the served file hashes to
`427C01AC016F21FE9D5BAC15B40BE42292F22A053BCA5F0CC96BC87CADA51EFB`, matching
`InstallerSha256` byte for byte. The installer itself was built and tested on
Windows hardware.

**Not verified.** `winget validate` and `winget install --manifest` were not
run, because these manifests were prepared on macOS where winget is not
available. I did not want to tick those boxes without having done it. If CI
surfaces anything, I will fix it promptly.

**CLA** is left unchecked because I have not confirmed its status; happy to
complete it when the bot asks.

### Note on signing

The installer is not code-signed, so SmartScreen shows a reputation prompt on
direct download. Installing through winget is the route that avoids that, which
is the reason for this submission.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/412045)


stephengillie:
> 
> Installation didn't complete within the timeout
> 2026-08-04 07:07:00.062 <E> [FAIL] C:\__w\1\s\external\pkg\src\AppInstallerCommonCore\Downloader.cpp(500)\WindowsPackageManager.dll!00007FF801854883: (caller: 00007FF80184FF09) Exception(1) tid(2990) 80070003 The system cannot find the path specified.
> 2026-08-04 07:07:00.111 <W> [CORE] RemoveMotwIfApplicable failed before IAttachmentExecute::Save(). Result: -2147024893. Proceeding with Save()
> 2026-08-27 11:40:43.554 <E> [FAIL] C:\__w\1\s\external\pkg\src\AppInstallerCommonCore\Downloader.cpp(500)\WindowsPackageManager.dll!00007FF973854883: (caller: 00007FF97384FF09) Exception(1) tid(1ef4) 80070003 The system cannot find the path specified.
> 2026-08-27 11:40:43.554 <W> [CORE] RemoveMotwIfApplicable failed before IAttachmentExecute::Save(). Result: -2147024893. Proceeding with Save()

JJBrownHere:
Corrected installer published; manifest updated to its SHA-256. Ready for another validation run.

**What was failing:** the Inno installer had a `[Run]` entry that shelled out to winget to fetch Apple's device service (our iPhone USB reading depends on it). In an unattended install that cannot complete - it raises a UAC prompt nothing can answer, and it calls winget from inside an install that winget itself started, so the outer process holds the package lock. That produced the `Installation didn't complete within the timeout` above.

**Fix:** `skipifsilent` on that `[Run]` entry, plus a `WizardSilent` guard on the post-install message box (that box was a second, independent way to hang an unattended install - `SuppressibleMsgBox` still draws when given `/VERYSILENT` without `/SUPPRESSMSGBOXES`). Interactive users still get the offer; silent installs skip it and the app detects the missing service at runtime instead, which is the better place to ask.

Verified locally before pushing: `/VERYSILENT` now completes in ~3 seconds with exit code 0.

The installer at the existing URL has been replaced and the manifest's `InstallerSha256` updated to `072CE7D2696518B7E562A6F2749CB1A3FFCC692F868ABFADBAD4984AB38074A5`. The application payload is byte-identical to the previously submitted binary (all 54 files hash-match) - only the installer wrapper changed, so this is the same 1.2.3 application.

Apologies for the delay in getting the corrected build up.

JJBrownHere:
Thanks for running it again, and good to see the installer itself now passes the pipeline.

On the domain flag: I believe this is triggered by the publisher segment of the identifier rather than by the URLs themselves being inconsistent. Setting out what the manifest declares:

| Field | Value |
|---|---|
| `PublisherUrl` | `https://itscrazyamazing.com` |
| `PublisherSupportUrl` | `https://itscrazyamazing.com` |
| `PackageUrl` | `https://windows.itscrazyamazing.com` |
| `InstallerUrl` | `https://windows.itscrazyamazing.com/The-STACKS-App-Setup-1.2.3.exe` |

The installer URL is on `windows.itscrazyamazing.com`, which is a subdomain of the `PublisherUrl` and the same host as the `PackageUrl`, so the installer is served from the publisher domain declared in this manifest. The download site is ours and is the only place we distribute the Windows build.

What does not appear in the manifest is `stacks.tech`, which is what the `StacksTech` prefix in `StacksTech.TheSTACKSApp` implies. `STACKS` is the product and company name (stacks.tech is our corporate site); `itscrazyamazing.com` is the application's own domain, which is why the app and its downloads live there. Both are operated by us. Each site is publicly branded STACKS - `stacks.tech` titles as "STACKS | Software for Phone Stores, Repair Shops & Resellers" and `itscrazyamazing.com` as "STACKS" - and the download page carries our company footer.

I am happy to provide whatever proof of control is most useful. I can:

1. Place a verification file at a path of your choosing on **either or both** domains (e.g. `https://stacks.tech/.well-known/<token>` and `https://windows.itscrazyamazing.com/.well-known/<token>`), or
2. Publish a link from `stacks.tech` to the Windows download page so the relationship between the two domains is publicly stated, or
3. Rename the package to use the distribution domain if you would prefer the identifier match the URLs directly.

Just tell me which you would accept and I will have it in place. Option 1 or 2 is quick on our side; I raise option 3 only because it is the one change that removes the ambiguity permanently, though I would rather not churn the identifier again if evidence is sufficient.

JJBrownHere:
Following up on my previous comment with a concrete offer, to save a round trip.

We would like to go with **option 1, a verification file**, and we can serve it directly from the installer host itself.

`windows.itscrazyamazing.com` is our own Cloudflare Worker which we deploy on demand, so we can publish a file at **any path you specify** on that exact hostname - the same origin the `InstallerUrl` points to. That seems like the most direct proof available, since it demonstrates control of the host serving the installer rather than of a related domain.

To do it in one pass, could you tell us:

1. the **exact path** (for example `/.well-known/microsoft-winget-verification.txt`, or whatever your tooling expects), and
2. the **exact contents** you expect - a token, the package identifier, or a specific string.

One detail worth flagging so it does not look like a failed check: unrecognised paths on that host currently return **404** by design, because the Worker only answers known routes. Adding the file is a small code change plus a redeploy rather than dropping a file into a bucket, which is why we need the precise path up front rather than guessing. Once we have it, we can have it live the same day.

We are also happy to serve the same file from `stacks.tech` if you would prefer the evidence come from the domain implied by the `StacksTech` identifier prefix - just say which host you want it on, or both.

For completeness on the earlier question: every URL in the manifest already resolves to one domain (`PublisherUrl` and `PublisherSupportUrl` are `itscrazyamazing.com`; `PackageUrl` and `InstallerUrl` are `windows.itscrazyamazing.com`). The only string referencing another domain is the `StacksTech` prefix of the identifier, which is our company name. So this is purely a proof-of-control question, and we are glad to satisfy it however you prefer.

JJBrownHere:
Renamed the package to `ItsCrazyAmazing.TheSTACKSApp` so the identifier matches the domain the installer is served from. That removes the mismatch this check was flagging rather than asking anyone to accept evidence for it.

Every field is now on one domain:

| Field | Value |
|---|---|
| `PackageIdentifier` | `ItsCrazyAmazing.TheSTACKSApp` |
| `PublisherUrl` | `https://itscrazyamazing.com` |
| `PublisherSupportUrl` | `https://itscrazyamazing.com` |
| `PackageUrl` | `https://windows.itscrazyamazing.com` |
| `InstallerUrl` | `https://windows.itscrazyamazing.com/The-STACKS-App-Setup-1.2.3.exe` |

The previous identifier used our company name (STACKS, stacks.tech) while the application and its downloads live on itscrazyamazing.com. Both are ours, but the identifier implied a domain that appeared nowhere in the manifest, which is what the domain check was reacting to. Aligning the identifier with the distribution domain is the cleaner fix, and it will not recur on future submissions.

Manifests have been moved from `manifests/s/StacksTech/...` to `manifests/i/ItsCrazyAmazing/...`; the old path is deleted in this branch. `PackageName` is unchanged ("The STACKS App"), so discovery by name is unaffected.

Sorry for the churn on this one, and thanks for the patience.

manifests/i/ItsCrazyAmazing/TheSTACKSApp/1.2.3/ItsCrazyAmazing.TheSTACKSApp.installer.yaml
```diff
@@ -0,0 +1,16 @@
+﻿# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+PackageIdentifier: ItsCrazyAmazing.TheSTACKSApp
```
mdanish-kh, line 2: Your PR description and body mention the PackageIdentifier `StacksTech.TheSTACKSApp` / `STACKS.TheSTACKSApp`, but the actual manifest uses `ItsCrazyAmazing` for the Publisher part of the PackageId. 

Is that intentional? I think the former ones seem more suitable

mdanish-kh (changes requested):

PR #414217, opened 2026-08-08
New package: SerZhyAle.FastMediaSorter.Server version 26.8.8.2305

SerZhyAle:
## Description

**New package: `SerZhyAle.FastMediaSorter.Server`** - "Fast Media Sorter Folder Share Server", version `26.8.8.2305`.

This is the always-on edition of the folder sharing already shipped by the existing package `SerZhyAle.FastMediaSorter`. It publishes user-chosen folders to an Android phone over SFTP from a **Windows service** (`FastMediaSorterCompanionSFTP`, delayed automatic start, running as LOCAL SERVICE), so the folders stay reachable from boot, with nobody signed in, and through a logoff.

**Why a separate package rather than a new version of the existing one:** the two are different products with different installers, different Inno `AppId`s, different ARP display names and different install directories. The existing `SerZhyAle.FastMediaSorter` package is a viewer that runs in the user session and needs no administrator rights - it must never start installing a service. A user who types this identifier has made that choice deliberately.

**Two things differ from the existing package's manifest, both deliberate:**

- `Scope: machine` - the product *is* machine state by construction (a service, a machine data directory with a restricted ACL, one program-scoped inbound firewall rule) and its installer is `PrivilegesRequired=admin`. Declaring the scope makes winget request elevation up front instead of discovering mid-install that it was needed.
- `MinimumOSVersion: 10.0.14393` - the management console is .NET 10 x64 and the whole point is a service that survives logoff, so the floor is real. (The other package deliberately omits this, because its setup also installs a 32-bit viewer for Windows 7/8.1.)

Everything else follows the same rules as the existing package: the Inno `setup.exe` is referenced directly (`InstallerType: inno`), with **no** `Dependencies` and **no** `NestedInstallerType`.

The silent switches include `/MIGRATEFROMUSER=yes`. The installer refuses a silent install when the user-session edition is present unless that switch is passed, so an unattended migration of existing state is never implicit - and the package passes it explicitly.

Links:
- Product page: https://serzhyale.github.io/FastMediaSorter_Lite/server.html
- Release: https://github.com/SerZhyAle/FastMediaSorter_Lite/releases/tag/v26.8.8.2305
- Changelog: https://github.com/SerZhyAle/FastMediaSorter_Lite/blob/main/CHANGELOG.md

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/414217)

### DIFF
--- a/manifests/s/SerZhyAle/FastMediaSorter/Server/26.8.8.2305/SerZhyAle.FastMediaSorter.Server.installer.yaml
+++ b/manifests/s/SerZhyAle/FastMediaSorter/Server/26.8.8.2305/SerZhyAle.FastMediaSorter.Server.installer.yaml
@@ -1,0 +1,43 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+#
+# The same hard-won rules as the User edition manifest apply here - read
+# docs/specifications/done/SPECIFICATION_WINGET_PUBLISHING.md before editing:
+#   * point at the Inno setup.exe directly (InstallerType: inno). Never at a ZIP.
+#   * NO `Dependencies` (sends winget into a resolution loop -> 0x8A150044).
+#   * NO `NestedInstallerType` (the "inconsistency" note is a non-blocking guide).
+#
+# Two things DIFFER from the User manifest, and both are deliberate:
+#
+#   * `Scope: machine`. The User manifest omits Scope entirely, because declaring
+#     `user` there forced a non-elevated flow that failed. Here the product is a
+#     Windows service, a machine state directory and a firewall rule - it is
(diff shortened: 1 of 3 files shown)


SerZhyAle:
Closing in favour of a fresh submission at the current release. This PR proposed 26.8.8.2305, which is now four releases behind; rather than re-point a PR that has been through validation, I am opening a new one for 26.8.15.1510 with the same manifest shape. Sorry for the extra round - and thank you for the URL-casing catch earlier, that fix is carried over.

SerZhyAle:
Superseded by #417819 (same package, version 26.8.15.1510).

PR #414775, opened 2026-08-10
New package: FernloopTechnologies.WhatsAppAIPro version 2.0.14

madison890000:
### New package: `FernloopTechnologies.WhatsAppAIPro` version `2.0.14`

WhatsApp AI Pro is a desktop app that adds AI auto-reply to an existing WhatsApp account via WhatsApp Web (QR code), without the WhatsApp Business API.

- **Publisher:** Fernloop Technologies Limited
- **Homepage:** https://waaipro.com
- **Installer:** NSIS (`nullsoft`), x64, 127.2 MB
- **License:** Proprietary — paid subscription (USD 49/month or USD 499/year) with a 7-day free trial

#### Why the last validation run failed, and what changed

The previous push submitted 2.0.13 with `Scope: user` and a bare `/S`. `08. Installation Validation` failed with `shellExecute`, and that failure was correct: **our installer really was broken.** Running `/S` under an elevated token crashed at `0xC0000005` roughly 6 seconds in and installed nothing.

The bug was upstream, in electron-builder's NSIS template. `multiUser.nsh` dereferenced an output pointer it never initialised:

```nsis
System::Call 'SHELL32::SHGetKnownFolderPath(... *p .r2)i.r1'
${If} $1 == 0
  System::Call '*$2(&w${NSIS_MAX_STRLEN} .s)'   # $2 = whatever was left in the register
  StrCpy $0 $1                                   # copies the HRESULT, not the path
```

Because `$2` was uninitialised, whether it crashed depended on leftover register contents — which is why it died under an elevated token but survived under a non-elevated one. electron-builder fixed this in 26.12.0 (`StrCpy $2 0`, `lstrcpynW` instead of the raw dereference, and `Push`/`Pop` instead of `System::Store`). **2.0.14 is built with 26.15.3.**

#### Measured install behaviour

Windows 11 26200 x64. Registry (`Uninstall\<APP_GUID>`, `Software\<APP_GUID>`) and install directory verified absent before **every** run. Non-elevated runs use a scheduled task with `/RL LIMITED`, i.e. a genuine non-elevated user token.

| Invocation | 2.0.13 (electron-builder 26.8.1) | **2.0.14 (26.15.3)** |
|---|---|---|
| Elevated `/S` | `0xC0000005`, 3/3, ~6 s, nothing installed | **exit 0, 2/2, 88.1 s / 81.9 s → HKCU, 6552 files** |
| Non-elevated `/S` | exit 0, 3/3 → HKCU | **exit 0, 2/2, 112.1 s / 105.9 s → HKCU, 6552 files** |

Every 2.0.14 run installed to `%LOCALAPPDATA%\Programs\WhatsApp AI Pro` (6552 files, 488.5 MB) and wrote a complete ARP entry to HKCU.

Because the installer is fixed, this manifest no longer needs the workarounds the 2.0.4 submission carried. `Scope: machine`, `ElevationRequirement: elevationRequired` and `Silent: /S /allusers` are all gone; `Scope: user` with a plain `/S` is now an accurate description of what the installer does.

`Publisher` is intentionally absent from `AppsAndFeaturesEntries`: the installer writes an empty `Publisher` to ARP, so declaring one would break correlation rather than help it. `ProductCode` `0e11ef54-7db6-5c68-a26b-5e856e20bd9a` was read back from the registry after an actual install.

#### Checklist

- [x] Manifests validate against the official `1.12.0` JSON schemas (`version`, `installer`, `defaultLocale`), checked with `jsonschema`
- [x] No other open PR exists for this package
- [x] `InstallerSha256` `EA02D443F3D7F6EA4E891C17E1C81E12A83FAD0D0F90ED17558BB80A14593D38` computed from the exact bytes served at `InstallerUrl`, independently on two machines
- [x] Installer is Authenticode signed and timestamped — issued by SSL.com to `FERNLOOP TECHNOLOGIES LIMITED`, matching the `Publisher` field
- [x] `InstallerUrl` is a permanent, version-pinned URL, not a "latest" redirect
- [x] Silent install verified from a clean machine under both elevated and non-elevated tokens — see table above
- [ ] `winget install --manifest` was attempted end to end but did not complete on my test machine. winget downloaded the installer and **verified the hash successfully**, then the launch was blocked by Microsoft Defender SmartScreen (`0x800704c7`), because winget's download carries Mark-of-the-Web (`ZoneId=3`) and this SSL.com OV certificate has not yet accumulated SmartScreen reputation. That is a reputation-age issue with a newly issued certificate, not a manifest problem — and it did not affect this pipeline, whose `08. Installation Validation` passed. My direct install measurements above used files fetched with `Invoke-WebRequest`, which does not attach Mark-of-the-Web, so they exercised the installer but not the SmartScreen path. Flagging both rather than checking a box I did not earn.

> The 2.0.4 and 2.0.13 directories have been removed; this PR now submits 2.0.14 only. Earlier review comments still apply — the Policy 1.8 subscription and pricing disclosure is carried over unchanged in `ShortDescription`, the opening paragraph of `Description`, and `PurchaseUrl`.


madison890000:
The dynamic validation failure was real — thanks for catching it. Our installer genuinely crashed on silent install, and I've fixed it rather than worked around it.

**Root cause (upstream):** electron-builder's NSIS `multiUser.nsh` dereferenced an output pointer it never initialised:

```nsis
System::Call 'SHELL32::SHGetKnownFolderPath(... *p .r2)i.r1'
${If} $1 == 0
  System::Call '*$2(&w${NSIS_MAX_STRLEN} .s)'   # $2 = leftover register contents
```

`/S` died at `0xC0000005` inside `System.dll` about 6 s in, installing nothing. Because `$2` was uninitialised, whether it crashed depended on leftover register state — it faulted under an elevated token and survived under a non-elevated one, which is exactly the asymmetry that made it hard to pin down. electron-builder fixed this in 26.12.0; **2.0.14 is built with 26.15.3.**

**Now measured on Windows 11 26200, clean registry and install dir verified before every run:**

| Invocation | 2.0.13 | 2.0.14 |
|---|---|---|
| Elevated `/S` | `0xC0000005`, 3/3 | exit 0, 2/2, → HKCU, 6552 files |
| Non-elevated `/S` (`schtasks /RL LIMITED`) | exit 0, 3/3 | exit 0, 2/2, → HKCU, 6552 files |

The PR now submits **2.0.14** only; the 2.0.4 and 2.0.13 directories are removed. Since the installer is fixed, the manifest drops the workarounds the 2.0.4 submission carried — `Scope: machine`, `ElevationRequirement: elevationRequired` and `/S /allusers` are gone, and `Scope: user` with a plain `/S` now accurately describes the installer.

One gap I want to state rather than paper over: I did not run `winget install --manifest` end to end, because it needs `winget settings --enable LocalManifestFiles` and I did not want to flip that admin-gated setting on the test machine. I verified the installer directly instead. Full detail is in the updated PR description.

Trenly:
cc @denelon @stephengillie

### Policy-Test-2.7 Analysis

2. **Flagged Substring:** None identified by the policy result; the engine only reports `triggered Adult Reference or Theme` for the full description, which describes paid WhatsApp automation and subscription billing.
3. **Recommendation:** waiver should be applied — the description contains no content a reasonable person would consider pornographic or sexually explicit. References to WhatsApp accounts, conversations, voice notes, and customer-message automation describe the product’s messaging functionality and are heuristic false positives under the Adult Content policy.

---
*Authored by Copilot*

PR #415077, opened 2026-08-10
New package: abooodbah.LeanRows version 0.1.3

abooodbah:
### Pull request has been manually verified

- [x] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/change?
- [x] Have you validated your manifest locally with `winget validate --manifest <path>`?
- [ ] Have you tested your manifest locally with `winget install --manifest <path>`?
- [x] Does your manifest conform to the [1.6 schema](https://github.com/microsoft/winget-cli/blob/master/schemas/JSON/manifests/v1.6.0)?

---

Two small MIT-licensed portable Windows utilities, published as zip archives with sidecar SHA256 files.

**abooodbah.LeanRows 0.1.3** — read-only viewer for large CSV, TSV, JSONL and log files. Indexes progressively and materializes only the rows in the current viewport, so peak memory tracks window size rather than file size. Measured on a 1.53 GB, 20,000,001-row CSV: 18.4 MB peak working set. 546 KB executable, no runtime dependency.
https://abooodbah.github.io/leanrows/

**abooodbah.LeanMark 0.2.0** — read-only Markdown viewer rendering GitHub-flavored Markdown and Mermaid entirely offline through the system WebView2 runtime, with raw HTML disabled and remote content blocked.
https://abooodbah.github.io/leanmark/

**Verification performed**

- `winget validate --manifest` passes for both packages (winget v1.29.280, manifest schema 1.6.0).
- SHA256 values verified against the published release archives by downloading and hashing them.
- Nested executable paths confirmed against the archive contents.

**Not performed:** `winget install --manifest` end-to-end. Both applications are already installed on this machine from their own installers, and I did not want a second portable copy shadowing them on PATH while testing. Happy to run it and report back if a maintainer would like that before merge.

Both installers are `zip` with `NestedInstallerType: portable`.

PR #415081, opened 2026-08-10
New package: abooodbah.LeanMark version 0.2.0

abooodbah:
### Pull request has been manually verified

- [x] Does your manifest conform to the [1.6 schema](https://github.com/microsoft/winget-cli/blob/master/schemas/JSON/manifests/v1.6.0)?

---

MIT-licensed portable Windows utility, published as a zip archive with a sidecar SHA256 file.

**abooodbah.LeanMark 0.2.0** — read-only Markdown viewer. Renders GitHub-flavored Markdown, local images and Mermaid diagrams entirely offline through the system WebView2 runtime rather than a bundled Chromium. Raw HTML is disabled, a strict Content Security Policy is enforced, and remote content and navigation are blocked.

Homepage: https://abooodbah.github.io/leanmark/

**Verification performed**

- `winget validate --manifest` passes (winget v1.29.280, manifest schema 1.6.0).
- SHA256 verified against the published release archive by downloading and hashing it.
- Nested executable path confirmed against the archive contents.

**Not performed:** `winget install --manifest` end-to-end. The application is already installed on this machine from its own installer and I did not want a second portable copy shadowing it on PATH. Happy to run it and report back if a maintainer would like that before merge.

Installer is `zip` with `NestedInstallerType: portable`.

Split out of #415077 following the validator note that a pull request must cover a single application.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/415081)

### DIFF
--- a/manifests/a/abooodbah/LeanMark/0.2.0/abooodbah.LeanMark.installer.yaml
+++ b/manifests/a/abooodbah/LeanMark/0.2.0/abooodbah.LeanMark.installer.yaml
@@ -1,0 +1,16 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+
+PackageIdentifier: abooodbah.LeanMark
+PackageVersion: 0.2.0
+InstallerType: zip
+NestedInstallerType: portable
+NestedInstallerFiles:
+- RelativeFilePath: LeanMark-v0.2.0-windows-x64\LeanMark.exe
+  PortableCommandAlias: leanmark
+ReleaseDate: 2026-08-08
+Installers:
+- Architecture: x64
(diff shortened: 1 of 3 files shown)


Trenly (changes requested):
The exe appears to depend on DLL's. `ArchiveBinariesDependOnPath` *or* a dependency on `Microsoft.EdgeWebView2Runtime` should be added to the installer manifest

Image

abooodbah:
@Trenly thanks for the review. I updated this to 0.3.0, the current release, and added both: `ArchiveBinariesDependOnPath: true`, since the archive ships `WebView2Loader.dll` and an `assets` folder that `LeanMark.exe` loads from its own folder, and a package dependency on `Microsoft.EdgeWebView2Runtime`, which LeanMark renders with. The manifests are on schema 1.12.0 now and `winget validate` passes.

PR #416473, opened 2026-08-12
New version: Hibbiki.Chromium version 151.0.7922.138

Hibbiki:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/416473)


Hibbiki:
> Please check to ensure the installer URL is correct and update the URL and the Hash if a change is made.
>
> If the package is mistakenly flagged as malware, you can submit the installer(s) to the Microsoft Defender team for analysis.

Would be nice if failing action actually showed error, instead of "Failed" with 0 details. Previously we could get actionable logs from Azure, not we're presented with useless message. What's even point of assigning `Needs-Author-Feedback` if author has 0 insight into what went wrong? Something did, but hell if I know what.

stephengillie (approved):

PR #416492, opened 2026-08-12
New package: JacobOptimiza.DevNav version 0.14.0

JacobOptimiza:
Update the existing New-Package PR to version 0.14.0 using the generated manifests from the published DevNav v0.14.0 release. Includes x64 and ARM64 user-scoped Inno installers, immutable versioned URLs, release SHA-256 hashes, and winget validate PASS.


JacobOptimiza:
Current PR state now contains exactly one multi-file manifest set, under
`manifests/j/JacobOptimiza/DevNav/0.14.0/`.

The superseded `0.9.5` files have been removed from the PR, and the current
diff contains only the three `0.14.0` manifest files. `winget validate` passes
locally.

The PullRequest-Error appears to have been produced while the previous PR state
was still being processed. Could a moderator please re-run validation on the
current HEAD (`449266ed50b0c10ced6970ca75d2c21f782e3856`)?

Trenly:
In my validation environment I didn't see the ability to run this application. Should the manifest declare a dependency on `Microsoft.PowerShell` (PowerShell 7) to ensure proper install?

<!--

PR #416963, opened 2026-08-13
New package: MSKazemi.YazSes version 2.18.2

MSKazemi:
### Package
`MSKazemi.YazSes` 2.18.1 — offline, on-device voice dictation (hold a key, speak, release; the text is typed into the focused window). Apache-2.0, https://github.com/MSKazemi/yazses

### Checklist
- [x] Manifests validated against the 1.6.0 schemas they declare.
- [x] `InstallerSha256` computed from the published release asset, not copied.
- [x] Installer is Inno Setup; silent install verified with `/VERYSILENT /SUPPRESSMSGBOXES /NORESTART`.
- [x] Per-user install (`PrivilegesRequired=lowest`), no elevation required.
- [x] `ProductCode` matches the Inno `AppId` + `_is1`, so `winget upgrade` detects an existing install.

### Note for moderators
This supersedes #415210 (version 2.17.0), which I am closing. That build has a defect that prevents the application from starting at all on a fresh install, so it should not reach users; 2.18.1 is the first release where the installed application runs.

Unattended install/uninstall of this exact artifact is exercised on a Windows runner in the project's own CI (install → run the CLI → uninstall → verify PATH is restored), so the silent switches above are tested rather than asserted.

### Not code-signed yet
The installer is unsigned; SmartScreen will warn. A SignPath Foundation application is in progress. Checksums for every artifact are published with the release and the verification procedure is documented at https://mskazemi.com/yazses/code-signing.html
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/416963)


MSKazemi:
Updated this PR from 2.18.2 to **2.35.0** — the current release — so there is still exactly one version in the diff and it is the latest. 2.18.2 has been removed from the branch.

Two substantive changes beyond the version stamp:

- **An arm64 installer entry.** The project has published `YazSes-<version>-windows-arm64.exe` since v2.22.0, but this manifest only ever listed x64, so Windows-on-ARM machines had no winget route. Both `InstallerSha256` values were computed from the released assets and cross-checked against the `SHA256SUMS.txt` published with the release.
- **2.18.2 was not a good version to publish.** It predates v2.34.0 (the Settings window, tray icon and overlay never appeared on Windows at all) and v2.35.0 (the in-app update check could not report that an update existed) — so anyone who installed 2.18.2 would have had no working update path out of it.

Everything else is unchanged: same Inno Setup installer, same silent switches, same per-user scope, same `ProductCode`.

@wingetbot run

This has been waiting on moderator approval since 13 August with all validation checks green. I know moderation is volunteer work and I'm not trying to jump the queue — if there's something in the submission I should fix to make it easier to approve, I'm happy to do it. Thanks for your time.

stephengillie:
Verify:
image
image

stephengillie (approved):

PR #417009, opened 2026-08-13
New package: Sparcle.BoltEnterprise version 0.1.150

sparcle-eng:
Bolt Enterprise 0.1.150, a local-first AI workspace published by Sparcle Inc.

Installer is the NSIS bundle from the official release at https://github.com/SparcleHQ/sparcle.app/releases/tag/v0.1.150.
SHA-256s in the manifest were taken from that release; the same values are published in its
SHA256SUMS, signed with minisign, and each artifact carries a VirusTotal report link.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/417009)


Trenly:
[Policy] CLA Signed

manifests/s/Sparcle/BoltEnterprise/0.1.150/Sparcle.BoltEnterprise.locale.en-US.yaml
```diff
@@ -0,0 +1,20 @@
+PackageIdentifier: Sparcle.BoltEnterprise
+PackageVersion: 0.1.150
+PackageLocale: en-US
+Publisher: Sparcle Inc.
```
Trenly, line 4:
```suggestion
Publisher: sparcle
```

Trenly (changes requested):
This package appears to have a dependency on `Microsoft.VCRedist.2015+.x64` that should be added to the manifest 

Image

sparcle-eng:
Thanks @Trenly — both points addressed, and pushed. Apologies for the slow turnaround.

**1. `Microsoft.VCRedist.2015+.x64` dependency — added.** You were right. Confirmed on our side: the binary is built with the MSVC toolchain and we do not set `crt-static`, so the runtime is linked dynamically and the dependency is real. It is now declared in the installer manifest.

**2. Publisher — fixed by moving the submission to 0.1.163 rather than by editing the field down.** Some detail, because your reading of the registry was correct and I do not want it to look like I am waving it away:

`0.1.150` genuinely writes `sparcle` to the registry, exactly as your screenshot shows. That build predates us setting a publisher, so our bundler derived the manufacturer from the application identifier `app.sparcle.bolt.enterprise` and took the second segment.

We corrected that on 2026-08-13 — the same day this PR was opened, which is why the fix missed the submitted artifact. From `0.1.151` onward the installer writes `Sparcle Inc.`, and we verified it on a clean Windows box: post-fix installs write `Software\Sparcle Inc.\Bolt`, where the pre-fix ones wrote `Software\sparcle`.

So taking the suggestion literally would have pinned our publisher name in the repository to a string that is an artifact of a bug we had already fixed. Instead this PR now submits **0.1.163**, where the installer writes `Sparcle Inc.` and the manifest's existing `Publisher: Sparcle Inc.` is both correct and matching. I have also added it explicitly under `AppsAndFeaturesEntries` so the ARP correlation is stated rather than inferred.

Happy to split this into a fresh PR and close this one if you would rather review a clean submission — just say which you prefer.

**What changed in this push**
- `0.1.150/` → `0.1.163/`
- `InstallerUrl` / `InstallerSha256` updated. The SHA-256 is from our release's signed `SHA256SUMS`, and matches the digest GitHub computed on upload.
- `Dependencies.PackageDependencies: Microsoft.VCRedist.2015+.x64`
- `AppsAndFeaturesEntries.Publisher: Sparcle Inc.`
- `ReleaseDate: 2026-09-21`

The branch is still named `...-0.1.150` — renaming it would close this PR, so I left it alone. The PR title is updated.

Thanks again for the review, and for the time you all put into this queue.

Trenly (approved):

Trenly:
[Policy] Reset Feedback - Inconsistent Label State

Trenly (approved):

Trenly:
[Policy] Needs Manual Merge - Automation Error

stephengillie:
@Trenly - Was merely waiting for [your earlier comment](https://github.com/microsoft/winget-pkgs/pull/417009#discussion_r4051426888) to be resolved.

PR #417273, opened 2026-08-14
New package: LoopTroopAI.LoopTroop version 0.5.2

looptroop-ai:
Adds LoopTroopAI.LoopTroop 0.5.2.

Replaces #417030, which we closed by accident: a force-push from a shallow clone left that branch without common history with `master`. Sorry for the noise — this branch is built directly on `master`.

One deliberate change from that pull request: `Architecture: x64` rather than `neutral`. The archive contains a single x64 executable, and `neutral` would have offered it on Windows on ARM, which we do not build.

Manifests are generated from the release manifest by the same code that renders this project's Homebrew, Scoop and Chocolatey descriptors, so the installer URL and checksum come from one source of truth.

The installer is a portable executable in a zip; it carries its own Node runtime, so only git is declared as a dependency.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/417273)


Trenly:
[Policy] CLA Signed

Trenly (approved):

PR #417549, opened 2026-08-14
New package:  OptimiseHealth.OptimiseBPDesktop 9.4.1

fraserbooth:
## 📖 Description
Adding OptBP Desktop to WinGet

## ✅ Checklist

## 📦 Manifest Checklist

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/417549)

### DIFF
--- a/manifests/o/OptimiseHealth/OptimiseBPDesktop/9.4.1/OptimiseHealth.OptimiseBPDesktop.installer.yaml
+++ b/manifests/o/OptimiseHealth/OptimiseBPDesktop/9.4.1/OptimiseHealth.OptimiseBPDesktop.installer.yaml
@@ -1,0 +1,12 @@
+# Created using wingetcreate 1.6.1.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: OptimiseHealth.OptimiseBPDesktop
+PackageVersion: 9.4.1
+InstallerType: nullsoft
+Installers:
+- InstallerUrl: https://www.optbp.com/dl/OptimiseBP-9.4.1.exe
+  Architecture: x86
+  InstallerSha256: 6dc56af99a29d879560c288a1cde320df77c414dc504bc82eba5449e6385353c
+ManifestType: installer
+ManifestVersion: 1.12.0
--- a/manifests/o/OptimiseHealth/OptimiseBPDesktop/9.4.1/OptimiseHealth.OptimiseBPDesktop.locale.en-UK.yaml
+++ b/manifests/o/OptimiseHealth/OptimiseBPDesktop/9.4.1/OptimiseHealth.OptimiseBPDesktop.locale.en-UK.yaml
@@ -1,0 +1,25 @@
+# Created using wingetcreate 1.6.1.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.12.0.schema.json
+
+PackageIdentifier: OptimiseHealth.OptimiseBPDesktop
+PackageVersion: 9.4.1
+PackageLocale: en-UK
+Publisher: OptimiseHealth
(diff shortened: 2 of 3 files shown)


fraserbooth:
Hi @stephengillie,
Sorry to bore you with this again...
I've improved the installer by upgrading the solution to .NET 10 and have versions with or without the runtime included available as per your previous suggestion.
This time the SHA256 seems to be doing something strange. I've downloaded directly from the URL and tried to validate directly via the WinGet Hash approach, but I'm none the wiser.
Thank you for sharing your wisdom...
Fraser

Trenly:
[Policy] CLA Signed

fraserbooth:
Hi,

I've investigated the hash mismatch further.

The manifest contains:

`6dc56af99a29d879560c288a1cde320df77c414dc504bc82eba5449e6385353c`

I have now downloaded the installer directly from:

`https://www.optbp.com/dl/OptimiseBP-9.4.1.exe`

using both a normal request and the `Microsoft-Delivery-Optimization/10.1` user agent.

Both downloads are exactly 3,080,160 bytes and both produce:

`6DC56AF99A29D879560C288A1CDE320DF77C414DC504BC82EBA5449E6385353C`

which exactly matches the manifest.

The response headers also show that the first request was fetched from the origin/CDN and the second was subsequently served from Cloudflare cache, with identical content.

Could you please let me know the SHA256 and file size actually received by the WinGet validation environment?

That should allow me to determine whether the validation environment is receiving a different CDN copy or another response entirely.

Thanks,
Fraser

Trenly:
> Could you please let me know the SHA256 and file size actually received by the WinGet validation environment?

```
2026-09-19 01:14:09.708 <I> [CORE] DeliveryOptimization downloading from url: https://www.optbp.com/dl/OptimiseBP-9.4.1.exe
2026-09-19 01:14:18.403 <I> [CORE] Download completed.
2026-09-19 01:14:18.485 <I> [CORE] Started applying motw to C:\Users\VALIDA~1\AppData\Local\Temp\WinGet\OptimiseHealth.OptimiseBPDesktop.9.4.1\6dc56af99a29d879560c288a1cde320df77c414dc504bc82eba5449e6385353c with zone: 3
2026-09-19 01:14:18.488 <I> [CORE] Finished applying motw
2026-09-19 01:14:18.488 <E> [CLI ] Package hash verification failed. SHA256 in manifest [6dc56af99a29d879560c288a1cde320df77c414dc504bc82eba5449e6385353c] does not match download [4022b9cac811125eec5a0113e25f2d661bfc66818144e39a31d729793d9af38c] with file size [3025970] and content type []
2026-09-19 01:14:18.489 <E> [CLI ] Terminating context: 0x8a150011 at C:\__w\1\s\external\pkg\src\AppInstallerCLICore\Workflows\DownloadFlow.cpp:244
```

<!--

fraserbooth:
Thank you ++

The failure was caused by Cloudflare/GitHub-Fastly edge regions serving different cached binaries for the same installer URL, so the WinGet validator received a different SHA-256.

I've now moved over to GitHub Releases for an immutable binary, and things seem to have passed this time :)

Trenly (changes requested):
It seems there's a missing dependency on a Dotnet runtime that should be added to the manifest

Image

Trenly:
It also seems this may be blocked by https://github.com/microsoft/winget-cli/issues/1665 as it appears the package specifically requires the x86 version of the Dotnet runtime and will not recognize the x64 version, meaning that on x64 systems the dependency won't actually be installed even if it is specified

fraserbooth:
Thank you for the advice, I didn't realise that Win11 didn't automatically pick the correct x86 version of .NET. I've shared the larger version that includes .NET with the installer.

Trenly (approved):

PR #417671, opened 2026-08-15
New package: sanlee-ys.telltale version 0.2.0

sanlee-ys:
New package: **sanlee-ys.telltale** version **0.2.0**

telltale is an open-source (MIT) Go CLI: a dispatch room where several coding-agent CLIs answer one brief side by side, over a statusline and a cross-vendor terminal HUD.

- Repository: https://github.com/sanlee-ys/telltale
- Release: https://github.com/sanlee-ys/telltale/releases/tag/v0.2.0

### Checklist

- [ ] Contributor License Agreement — not yet confirmed by me; I will sign if the bot asks.
- [x] Checked open pull requests for a competing add of this manifest — there are none.
- [x] Validated locally: `winget validate --manifest <path>` → **Manifest validation succeeded.**
- [ ] `winget install --manifest <path>` — **not run**, and I would rather say so than tick it. That path needs `LocalManifestFiles` enabled, which requires elevation this submission was not prepared with. What was done instead is below.
- [x] Conforms to the 1.12 schema.

### What was actually verified

- **The archive and its hash.** `InstallerSha256` comes from the release's own `checksums.txt` and was independently re-checked against the published asset. The download URL was confirmed to resolve, returning 3,984,461 bytes.
- **A real install of the identical archive.** The same zip and the same hash back the project's scoop manifest, and `scoop install telltale` was run end to end on a clean Windows 11 machine before this submission: scoop verified the hash itself, shimmed the binary, and `telltale version` reported `telltale 0.2.0`. That exercises the extract-and-shim path this manifest describes, though it is not winget's own installer.

### Notes on the manifest

**Package shape.** telltale ships as a zip containing a single portable executable, so this is `InstallerType: zip` with `NestedInstallerType: portable` — the same shape fzf, ripgrep and zoxide use. There is no installer.

**Architecture.** Only `x64` is listed. `windows/arm64` is not built by the project at all, so there is nothing to list rather than an untested entry.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/417671)

### DIFF
--- a/manifests/s/sanlee-ys/telltale/0.2.0/sanlee-ys.telltale.installer.yaml
+++ b/manifests/s/sanlee-ys/telltale/0.2.0/sanlee-ys.telltale.installer.yaml
@@ -1,0 +1,24 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+#
+# Shape: telltale ships as a zip holding a single portable exe, so this is
+# InstallerType: zip + NestedInstallerType: portable — the same shape fzf,
+# ripgrep and zoxide use, and the reason there is no installer to write.
+#
+# ONLY windows_amd64 is listed, and that is deliberate rather than an
+# oversight: winget installs on Windows, and windows/amd64 is the one target
+# this project continuously verifies. windows/arm64 is not built at all, so
+# there is nothing to list for it.
+PackageIdentifier: sanlee-ys.telltale
+PackageVersion: "0.2.0"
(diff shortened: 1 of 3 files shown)


sanlee-ys:
@microsoft-github-policy-service agree.

sanlee-ys:

Trenly (approved):

PR #417700, opened 2026-08-15
New package: bryramirezp.DictationTool version 0.1.0

bryramirezp:
### New package: `bryramirezp.DictationTool` version `0.1.0`

Push-to-talk dictation for Windows. Hold a key, speak, let go, and the text is
typed into whatever window has focus. Speech recognition runs locally through
faster-whisper, so it works with no internet connection and nothing is uploaded.

- Homepage: https://bryramirezp.github.io/dictation-tool/
- Source: https://github.com/bryramirezp/dictation-tool
- Release: https://github.com/bryramirezp/dictation-tool/releases/tag/v0.1.0
- License: MIT

I am the author of this package.

#### Checklist

- [ ] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)?
- [x] Have you checked that there aren't other open pull requests for the same manifest update/addition?
- [x] Have you validated your manifest locally with `winget validate --manifest <path>`?
- [ ] Have you tested your manifest locally with `winget install --manifest <path>`?
- [x] Does your manifest conform to the [1.12 schema](https://github.com/microsoft/winget-cli/blob/master/schemas/JSON/manifests/v1.12.0)?

On the CLA box: being agreed to in a comment below.

On the second unchecked box: `winget validate` passes, but `winget install --manifest`
needs `LocalManifestFiles` enabled, which requires an elevated shell I did not
have on this machine. Instead the released installer itself was installed and
uninstalled silently with the switches winget uses for `InstallerType: inno`
(`/VERYSILENT /SUPPRESSMSGBOXES /NORESTART`). Happy to run the manifest install
and report back if that is required.

#### Notes

Installs per user under `%LOCALAPPDATA%\Programs\DictationTool`, so it needs no
administrator rights; `Scope` is set to `user` accordingly.

`AppsAndFeaturesEntries` carries only the `ProductCode`, read from the registry
after a real install rather than guessed. Inno Setup registers the display name
as "Dictation Tool version 0.1.0", so pinning `DisplayName` would break matching
on the next release.

Verified on Windows 11: silent install creates the Start Menu entry and the
optional startup shortcut, silent uninstall removes the install directory, the
registry entry and that shortcut, and user settings under `%LOCALAPPDATA%` are
left alone. The installer SHA256 matches the release asset as downloaded.

[...]

PR #417713, opened 2026-08-15
New package: bryramirezp.Kara version 0.2.1

bryramirezp:
### New package: `bryramirezp.Kara` version `0.2.0`

Push-to-talk dictation for Windows. Hold a key, speak, let go, and the text is
typed into whatever window has focus. Speech recognition runs locally through
faster-whisper, so it works with no internet connection and nothing is uploaded.

- Homepage: https://bryramirezp.github.io/kara/
- Source: https://github.com/bryramirezp/kara
- Release: https://github.com/bryramirezp/kara/releases/tag/v0.2.0
- License: MIT

I am the author of this package.

This replaces #417700, which I opened for the same program under its old name,
`bryramirezp.DictationTool`. The project was renamed to Kara before it shipped,
so I closed that one rather than have you publish an identifier that would have
needed a move on its first update. Apologies for the extra review round.

#### Checklist

- [x] Have you checked that there aren't other open pull requests for the same manifest update/addition?
- [x] Does your manifest conform to the [1.12 schema](https://github.com/microsoft/winget-cli/blob/master/schemas/JSON/manifests/v1.12.0)?

On the unchecked box: `winget validate` passes, but `winget install --manifest`
needs `LocalManifestFiles` enabled, which requires an elevated shell I did not
have. Instead the released installer itself was installed and uninstalled
silently with the switches winget uses for `InstallerType: inno`
(`/VERYSILENT /SUPPRESSMSGBOXES /NORESTART`).

#### Notes

Installs per user under `%LOCALAPPDATA%\Programs\Kara`, so it needs no
administrator rights; `Scope` is set to `user` accordingly.

`AppsAndFeaturesEntries` carries only the `ProductCode`, read from the registry
after a real install rather than guessed. Inno Setup writes the display name
with the version in it, so pinning `DisplayName` would break matching on the
next release.

Verified on Windows 11: silent install creates the Start Menu entry and the
optional startup shortcut, silent uninstall removes the install directory, the
registry entry and that shortcut, and user settings under `%LOCALAPPDATA%` are
left alone. The installer SHA256 matches the release asset as downloaded.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/417713)


bryramirezp:
Updated this pull request from 0.2.1 — sorry, from 0.2.0 to 0.2.1, before review picked it up.

0.2.0 shipped with a contrast defect in the settings panel: the selected option in a segmented control was drawn with the accent colour behind it and the normal label colour on top, which measured 1.02:1. The option you had just chosen was the one you could not read. 0.2.1 fixes it and I would rather not have 0.2.0 be the first version published.

Manifests now point at the 0.2.1 release and carry its installer hash. `winget validate` passes.

PR #417719, opened 2026-08-15
Remove version: athrvk.Vayu version 0.16.0

athrvk:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/417719)


stephengillie:
Hi @,

(Deterministic automation - build 1808.)

athrvk:
Withdrawing this — it was filed in error, and 0.16.0 should not be removed.

Our release workflow used vedantmgoyal9/winget-releaser@v2 with max-versions-to-keep: 5. That action's cleanup step deletes the newest versions rather than the oldest: ConvertFrom-Json -NoEnumerate passes the version array to Sort-Object as a single object, so the sort is a no-op, -Descending never applies, and the tail slice picks the most recent releases. This is [vedantmgoyal9/winget-releaser#323](https://github.com/vedantmgoyal9/winget-releaser/issues/323), fixed on main in 19e706d but never released — the v2 tag still points at a commit from 2024-11-27, six months before the fix.

The companion PR [#417718](https://github.com/microsoft/winget-pkgs/pull/417718) (0.15.0) merged before we caught this; it was the same mis-fire. 0.17.0 is pending in [#417717](https://github.com/microsoft/winget-pkgs/pull/417717).

We have removed max-versions-to-keep from our release workflow so this cannot recur. Thanks @stephengillie for catching it.

Trenly:
Close with reason: Above;

PR #417819, opened 2026-08-15
New package: SerZhyAle.FastMediaSorter.Server version 26.8.15.1510

SerZhyAle:
## Description

**New package: `SerZhyAle.FastMediaSorter.Server`** - "Fast Media Sorter Folder Share Server", version `26.8.15.1510`.

This supersedes #414217, which proposed the same new package at `26.8.8.2305`. That version is now four releases behind, so rather than re-point a PR that had already been through validation I closed it and am submitting the current release. The manifest shape is unchanged from it, including the `PackageUrl` / `DocumentUrl` casing fix made there.

This is the always-on edition of the folder sharing already shipped by the existing package `SerZhyAle.FastMediaSorter`. It publishes user-chosen folders to an Android phone over SFTP from a **Windows service** (`FastMediaSorterCompanionSFTP`, delayed automatic start, running as LOCAL SERVICE), so the folders stay reachable from boot, with nobody signed in, and through a logoff.

**Why a separate package rather than a new version of the existing one:** the two are different products with different installers, different Inno `AppId`s, different ARP display names and different install directories. The existing `SerZhyAle.FastMediaSorter` package is a viewer that runs in the user session and needs no administrator rights - it must never start installing a service. A user who types this identifier has made that choice deliberately.

**Two things differ from the existing package's manifest, both deliberate:**

- `Scope: machine` - the product *is* machine state by construction (a service, a machine data directory with a restricted ACL, one program-scoped inbound firewall rule) and its installer is `PrivilegesRequired=admin`. Declaring the scope makes winget request elevation up front instead of discovering mid-install that it was needed.
- `MinimumOSVersion: 10.0.14393` - the management console is .NET 10 x64 and the whole point is a service that survives logoff, so the floor is real. (The other package deliberately omits this, because its setup also installs a 32-bit viewer for Windows 7/8.1.)

Everything else follows the same rules as the existing package: the Inno `setup.exe` is referenced directly (`InstallerType: inno`), with **no** `Dependencies` and **no** `NestedInstallerType`.

The silent switches include `/MIGRATEFROMUSER=yes`. The installer refuses a silent install when the user-session edition is present unless that switch is passed, so an unattended migration of existing state is never implicit - and the package passes it explicitly.

Links:
- Product page: https://serzhyale.github.io/FastMediaSorter_Lite/server.html
- Release: https://github.com/SerZhyAle/FastMediaSorter_Lite/releases/tag/v26.8.15.1510
- Changelog: https://github.com/SerZhyAle/FastMediaSorter_Lite/blob/main/CHANGELOG.md

---

### DIFF
--- a/manifests/s/SerZhyAle/FastMediaSorter/Server/26.8.15.1510/SerZhyAle.FastMediaSorter.Server.locale.en-US.yaml
+++ b/manifests/s/SerZhyAle/FastMediaSorter/Server/26.8.15.1510/SerZhyAle.FastMediaSorter.Server.locale.en-US.yaml
@@ -1,0 +1,66 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.12.0.schema.json
+
+PackageIdentifier: SerZhyAle.FastMediaSorter.Server
+PackageVersion: "26.8.15.1510"
+PackageLocale: en-US
+Publisher: SZA
+PublisherUrl: https://github.com/SerZhyAle
+PublisherSupportUrl: https://github.com/SerZhyAle/FastMediaSorter_Lite/issues
+Author: Serhii Zhyhunenko
+# FROZEN once published: it must keep matching the ARP DisplayName written by
+# FastMediaSorterServer.iss, which is how `winget upgrade` correlates this package.
+# It is deliberately NOT "FastMediaSorter LITE" - that name belongs to the User
+# edition package and sharing it would make the two indistinguishable to winget.
+PackageName: Fast Media Sorter Folder Share Server
+PackageUrl: https://serzhyale.github.io/FastMediaSorter_Lite/server.html
+License: MIT
+LicenseUrl: https://github.com/SerZhyAle/FastMediaSorter_Lite/blob/main/LICENSE
+Copyright: Copyright (c) 2025 SZA
+ShortDescription: Always-on folder sharing to Android over SFTP - a Windows service that starts at boot and serves with nobody signed in. Administrator rights required.
+Description: |-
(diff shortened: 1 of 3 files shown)


manifests/s/SerZhyAle/FastMediaSorter/Server/26.8.15.1510/SerZhyAle.FastMediaSorter.Server.locale.en-US.yaml
Trenly, line 6:
```
Publisher: SerZhyAle
```

Trenly, line 44: This implies that the installer manifest should contain either `ElevationRequirement: elevationRequired` or `ElevationRequirement: elevatesSelf`

Trenly (changes requested):

SerZhyAle:
@KalebLuedtke Thanks for the review - both points addressed:
- Publisher in the locale manifest now matches the registry value the installer writes (AppPublisher = SerZhyAle in the Inno script), changed from SZA to SerZhyAle.
- Added ElevationRequirement: elevatesSelf to the installer manifest - the installer is built with PrivilegesRequired=admin, so its own embedded manifest already requests admin execution level and Windows prompts UAC on launch without winget needing to elevate itself.

Ready for another look.

Trenly (approved):
Please mark all open comment threads as resolved

Trenly:
[Policy] Needs Manual Merge - Automation error

PR #418462, opened 2026-08-16
New package: Cwdio.CwdioFileManager version 0.1.5

cwdio:
## ✅ Checklist

  - Resolves #[Issue Number]

## 📦 Manifest Checklist

- [ ] Checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/change
- [ ] Validated manifest locally with `winget validate --manifest <path>` ([validation guide](https://github.com/microsoft/winget-pkgs/blob/master/doc/ValidationFailureGuide.md))

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/418462)


cwdio:

cwdio:
Updated this PR from 0.1.4 to **0.1.5**.

0.1.5 is the first code-signed release (SSL.com individual-validated certificate, `CN=Ihor Melnyk`), so this avoids introducing the package with an unsigned installer. The manifests point at the new release asset and its SHA256; everything else is unchanged.

manifests/c/Cwdio/CwdioFileManager/0.1.5/Cwdio.CwdioFileManager.installer.yaml
```diff
@@ -0,0 +1,16 @@
...
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+
+PackageIdentifier: Cwdio.CwdioFileManager
+PackageVersion: 0.1.5
+InstallerType: zip
+NestedInstallerType: portable
+NestedInstallerFiles:
+- RelativeFilePath: Cwdio File Manager-win32-x64\cwdio.exe
+  PortableCommandAlias: cwdio
+Installers:
```
Trenly, line 12:
```suggestion
ArchiveBinariesDependOnPath: true
Installers:
```

Trenly (changes requested):
Because of the DLLs, the application won't work if symlinked

cwdio:
@Trenly thanks for catching this — you're right. The portable build is an Electron app: `cwdio.exe` sits next to six DLLs plus `resources.pak`, `icudtl.dat`, `v8_context_snapshot.bin` and `locales/`, none of which it can resolve when it is launched through a link.

One thing your suggestion couldn't carry on its own: `ArchiveBinariesDependOnPath` was introduced in manifest version **1.9.0**, and these manifests were still on 1.6.0. Added there it doesn't fail — `winget validate` returns `Manifest Warning: Unknown field. [ArchiveBinariesDependOnPath]` and the setting is silently ignored, which is the worse outcome here, since the package would merge looking fixed and still install broken. So all three manifests moved to **1.12.0** in the same commit; nothing else changed, and `winget validate` is clean.

Pushed in eabe8ff. (Replying here rather than in the review thread — the reply endpoint is returning 502 on this PR.)

Trenly (approved):

PR #418530, opened 2026-08-16
Remove version: Pantelis23.KernRift version 2.3.0

Pantelis23:
## 📖 Description

Removes `Pantelis23.KernRift` version `2.3.0`.

It is a superseded April 2026 build, and it is the only one of the package's 27
published versions without the `v` prefix the other 26 use.

That inconsistency has a downstream cost. winget's own comparator is unaffected —
it strips leading non-digits, so `v2.10.0` correctly resolves as latest — but
third-party sites that mirror this repository and sort the version strings
naively pick `2.3.0` as current. One Czech download portal is currently
advertising KernRift 2.3.0 as the current release, complete with that manifest's
`License: MIT` string. The project has been Apache 2.0 since 2026-05-19, so
readers there are being shown a four-month-old build under the wrong licence.

To be straight about what this is: `2.3.0`'s installer URLs still resolve, so
this is housekeeping rather than the removal of a broken package. Removing the
one oddly-shaped entry fixes the problem at its source for every mirror rather
than one site at a time. If you'd rather I republished it as `v2.3.0` instead of
removing it, I'm glad to do that — say the word.

Deletes the three files under `manifests/p/Pantelis23/KernRift/2.3.0/` and
nothing else.

## ✅ Checklist

## 📦 Manifest Checklist

- [x] I am the package publisher

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/418530)


Pantelis23:
Verified on Windows (winget v1.29.280) that removing this version does not change what winget resolves — it already ignores `2.3.0` today:

```
> winget show Pantelis23.KernRift
Found KernRift [Pantelis23.KernRift]
Version: v2.10.0
License: Apache-2.0
```

So this removal is safe with respect to winget itself, and only affects downstream consumers that sort the published version strings directly — which is the problem it is meant to fix.

stephengillie (approved):

PR #418681, opened 2026-08-17
New version: LunarWerxs.SageThumbs2K version 2.1.2

LunarWerxs:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/418681)

### DIFF
--- a/manifests/l/LunarWerxs/SageThumbs2K/2.1.2/LunarWerxs.SageThumbs2K.installer.yaml
+++ b/manifests/l/LunarWerxs/SageThumbs2K/2.1.2/LunarWerxs.SageThumbs2K.installer.yaml
@@ -1,0 +1,30 @@
+# Created with WinGet Releaser using komac v2.16.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: LunarWerxs.SageThumbs2K
+PackageVersion: 2.1.2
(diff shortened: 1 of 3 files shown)


LunarWerxs:
The `Manifest-Metadata-Consistency` flag here is a false positive, and the same one has now
parked four consecutive releases (#417703, #418308, #418618, #418644, and this one).

The bot reports `Missing property UnsupportedOSArchitectures` against published version 1.8.1.
That property is missing on purpose. 1.8.1 was an x64-only package, so it carried a root-level
`UnsupportedOSArchitectures: [arm64]`. Since then the project ships a native ARM64 installer, so
the package as a whole does support arm64 and the root-level declaration would now be false. The
property moved down to installer scope, where it correctly marks only the x64 binary as
unsuitable for an arm64 machine:

```yaml
Installers:
- Architecture: x64
  UnsupportedOSArchitectures:
  - arm64
- Architecture: arm64
```

Restoring it at the root would clear the flag and would be inaccurate, so I would rather not do
that. All ten validation checks pass on this PR.

Could a moderator take a look and clear the consistency flag? Once one of these merges the
published baseline updates and it should stop firing on its own. Happy to change the manifest
if there is a preferred way to express "was x64-only, now ships both". Thanks.

stephengillie (approved):

PR #249737, opened 2025-04-19
New package: Flowscripter.example-host-application version 1.0.0

vectronic:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/249737)

PR #418790, opened 2026-08-17
New version: Flowscripter.flowscripter-io-cli version 1.0.0

vectronic:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/418790)

### DIFF
--- a/manifests/f/Flowscripter/flowscripter-io-cli/1.0.0/Flowscripter.flowscripter-io-cli.installer.yaml
+++ b/manifests/f/Flowscripter/flowscripter-io-cli/1.0.0/Flowscripter.flowscripter-io-cli.installer.yaml
@@ -1,0 +1,23 @@
+# Created with WinGet Releaser using komac v2.16.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: Flowscripter.flowscripter-io-cli
+PackageVersion: 1.0.0
+InstallerType: zip
+NestedInstallerType: portable
+NestedInstallerFiles:
+- RelativeFilePath: flowscripter-io-cli.exe
+InstallModes:
+- silent
+UpgradeBehavior: install
+ReleaseDate: 2026-08-17
+ElevationRequirement: elevationProhibited
+Installers:
+- Architecture: x64
+  InstallerUrl: https://github.com/flowscripter/flowscripter-io-cli/releases/download/v1.0.0/flowscripter-io-cli_Windows_x64.zip
+  InstallerSha256: 52190FF65168F2C83A9EF7BA1A146CB20CBC34CAAF5DF1B3366F98071D4EC55A
+- Architecture: arm64
+  InstallerUrl: https://github.com/flowscripter/flowscripter-io-cli/releases/download/v1.0.0/flowscripter-io-cli_Windows_arm64.zip
+  InstallerSha256: D134988128B447170A35438BB12F3E4CF059902E0D7AD11DB3EC24D107A14757
+ManifestType: installer
+ManifestVersion: 1.12.0
--- a/manifests/f/Flowscripter/flowscripter-io-cli/1.0.0/Flowscripter.flowscripter-io-cli.locale.en-US.yaml
+++ b/manifests/f/Flowscripter/flowscripter-io-cli/1.0.0/Flowscripter.flowscripter-io-cli.locale.en-US.yaml
@@ -1,0 +1,20 @@
+# Created with WinGet Releaser using komac v2.16.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.12.0.schema.json
+
+PackageIdentifier: Flowscripter.flowscripter-io-cli
+PackageVersion: 1.0.0
+PackageLocale: en-US
+Publisher: Flowscripter
+PublisherUrl: https://github.com/flowscripter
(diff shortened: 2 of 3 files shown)


stephengillie:
> 
> Caught exception while trying to kill child processes: System.TypeInitializationException: The type initializer for 'System.Management.ManagementPath' threw an exception.
>  ---> System.TypeInitializationException: The type initializer for 'System.Management.WmiNetUtilsHelper' threw an exception.
>  ---> System.ComponentModel.Win32Exception (193): Failed to load required native library 'C:\Windows\Microsoft.NET\Framework64\v4.0.30319\wminet_utils.dll'.
> Executable C:\Users\validator\AppData\Local\Microsoft\WinGet\Packages\Flowscripter.flowscripter-io-cli__DefaultSource\flowscripter-io-cli.exe returned exit code: 1 
 
  | 00000001 | 1 | -4294967295 | ERROR_INVALID_FUNCTION | Incorrect function. | 


manifests/f/Flowscripter/flowscripter-io-cli/1.0.0/Flowscripter.flowscripter-io-cli.installer.yaml
```diff
@@ -0,0 +1,23 @@
...
+PackageVersion: 1.0.0
+InstallerType: zip
+NestedInstallerType: portable
+NestedInstallerFiles:
+- RelativeFilePath: flowscripter-io-cli.exe
+InstallModes:
+- silent
+UpgradeBehavior: install
+ReleaseDate: 2026-08-17
+ElevationRequirement: elevationProhibited
```
Trenly, line 14: Why is `elevationProhibited` for a portable installer type? Since WinGet handles the install, I see no reason to prohibit elevation for this

manifests/f/Flowscripter/flowscripter-io-cli/1.0.0/Flowscripter.flowscripter-io-cli.locale.en-US.yaml
```diff
@@ -0,0 +1,20 @@
...
+Publisher: Flowscripter
+PublisherUrl: https://github.com/flowscripter
+PublisherSupportUrl: https://github.com/flowscripter/flowscripter-io-cli/issues
+Author: Vectronic
+PackageName: Flowscripter IO CLI
+PackageUrl: https://github.com/flowscripter/flowscripter-io-cli
+License: MIT
+LicenseUrl: https://github.com/flowscripter/flowscripter-io-cli/blob/HEAD/LICENSE
+Copyright: MIT (c) Flowscripter
+ShortDescription: Example CLI using pluggable-io-framework
```
Trenly, line 16: If this is an example CLI, what value does it provide to users to be able to install and run it?

Trenly:

manifests/f/Flowscripter/flowscripter-io-cli/1.0.0/Flowscripter.flowscripter-io-cli.installer.yaml
vectronic: Thank you for your review. `ElevationRequirement: elevationProhibited` was a workaround proposed here https://github.com/microsoft/winget-pkgs/pull/249737#issuecomment-2825938419 for this issue described in the comment directly preceding it. The root cause is now fixed https://github.com/oven-sh/bun/issues/18193 and I should be able to remove this now. I was testing this out on another winget application first but hit some other areas. I will push an update here removing it and see how it goes.

manifests/f/Flowscripter/flowscripter-io-cli/1.0.0/Flowscripter.flowscripter-io-cli.locale.en-US.yaml
vectronic: Thank you for you review. I will update this as it is intended to grow into a useful application. I had copied from a template repo I had which stated "Example". I will push an update here.

manifests/f/Flowscripter/flowscripter-io-cli/1.0.0/Flowscripter.flowscripter-io-cli.installer.yaml
Trenly: For future reference, `ArchiveBinariesDependOnPath: true` will cause WinGet to not make the Symlink.

Symlinks are created to help avoid path length issues where possible, but its true that some applications just don’t work with symlinks. The criteria for WinGet to add the symlink is either:
1) Running in an elevated (Administrator) context + ArchiveBinariesDependOnPath == false
2) Developer mode enabled + ArchiveBinariesDependOnPath == false

So even in the case where a user is not running as Administrator, it could still be possible for a symlink to be created if developer mode is enabled, unless the property to avoid the symlink is explicitly set in the manifest

Trenly:
The `Validation-Executable-Error` seems to be coming from the fact that the validation VMs are bare and the manifest doesn't specify any dependences:
```
error: Neither 'bun' nor 'npm' found on PATH; specify installCommand explicitly to use a different package manager
```

However, even after installing `Oven-sh.Bun`, running the executable with no arguments results in `No command specified` and an exit code of `2` which will trigger the `Validation-Executable-Error` even if the dependency were added to the manifest.

Since users have the choice of package manager, I think it's correct not to specify the dependencies. 

@stephengillie  - I'm recommending a waiver for this package

PR #420408, opened 2026-08-19
New package: rootMonsteR.HoldToSpeak version 0.1.3

rootMonsteR:
### New package: `rootMonsteR.HoldToSpeak` version 0.1.2

HoldToSpeak is an open-source (MIT), fully offline push-to-talk dictation app for Windows: hold
Ctrl+Win, speak, release, and the text appears at the cursor. Speech recognition runs locally on
the CPU — there is no account, no telemetry, and no cloud service.

- Homepage: https://holdtospeak.com
- Source: https://github.com/rootMonsteR/holdtospeak
- Release: https://github.com/rootMonsteR/holdtospeak/releases/tag/v0.1.2

#### Checklist

- [x] Manifest validates locally: `winget validate --manifest ...` → **Manifest validation succeeded**
- [x] Conforms to the 1.6 manifest schema
- [x] No other open PRs for this package (this is its first submission)
- [x] The installer was tested directly: silent install and uninstall both exit 0, all files land,
      the `DisplayName`/`DisplayVersion`/`Publisher` registry values are correct, and uninstall
      leaves neither the install directory nor the `HKCU` uninstall key behind
- [x] `InstallerSha256` was verified against the **published** release asset by downloading it from
      `InstallerUrl` and hashing it, not just against the local build

#### Notes for reviewers

- **Per-user install.** The installer is Inno Setup with `PrivilegesRequired=lowest`, so it installs
  to `%LOCALAPPDATA%` and needs no elevation. Hence `Scope: user`.
- **`ProductCode`** is `{8F3A5C21-9D74-4E6B-B2A8-1C5E7D0F4A93}_is1` — the real Inno uninstall key,
  read back from the registry after a live install rather than inferred from the AppId.
- **The installer is not code-signed yet.** Flagging this up front in case it affects validation;
  signing is planned but not in place for this release.
- The ~460 MB speech model is not bundled. It is downloaded on first run from the upstream
  sherpa-onnx GitHub release and verified against a pinned SHA-256; it is separately licensed
  (CC-BY-4.0), which is why it is fetched rather than redistributed.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/420408)


rootMonsteR:

rootMonsteR:
Thanks — the `Validation-Executable-Error` was a real bug on my side, and I've updated this PR to **version 0.1.3**, which fixes it. (The 0.1.2 manifests have been removed in favour of 0.1.3.)

**What was wrong.** On a startup failure the app kept its console window open so the error could be read, and waited for Enter. That's fine for a person, but an unattended launcher has nobody to press anything — so the process hung instead of exiting. I reproduced it locally (still alive after 20 s with no user present) rather than guessing, since no logs were published.

**The fix.** The wait is now bounded: stdin is read on a worker thread while the main thread waits with a timeout, so a human still gets 60 s to read the message and an unattended launcher never blocks. Verified against the **built installer**, not just a development build — it now exits on its own after ~58 s with exit code 1.

**One thing worth flagging before it re-runs.** HoldToSpeak is a push-to-talk dictation app, so it requires a **microphone** and a locally downloaded speech model. If the validation VM has no audio input device, the app will legitimately fail to start there no matter what I change — it exits cleanly with a clear message now, but it can't run headless. If that trips the executable check again, is there a preferred way to handle it (a label, or a note in the manifest)? Happy to make whatever change helps.

For reference:
- Release: https://github.com/rootMonsteR/holdtospeak/releases/tag/v0.1.3
- `winget validate` passes on the updated manifests, and `InstallerSha256` was checked by downloading the published asset from `InstallerUrl` and hashing it.
- Installer is per-user Inno Setup (`PrivilegesRequired=lowest`); silent install and uninstall both exit 0 and leave nothing behind.
- Still not code-signed — flagging again in case it matters for the Defender/reputation step.

rootMonsteR:
Correcting my earlier comment — **my diagnosis was wrong**, and I'd rather say so than have anyone spend time on it.

I'd guessed the executable error was about the app needing a microphone. It isn't. Reading the documented meaning of the label — *"during installation testing, the test was unable to locate the primary application"* — pointed straight at the real problem: **the installed executable wasn't named after the package.**

The product is HoldToSpeak, but a rebrand had stopped short of the binaries, so the installer was laying down `nib-core.exe` and `nib-asr-sidecar.exe` (an old internal crate name). Nothing in the install folder was named after the package, so there was no primary application to find. Fixed:

```
nib-core.exe         ->  HoldToSpeak.exe
nib-asr-sidecar.exe  ->  HoldToSpeak-engine.exe
```

Please disregard the microphone theory in my previous comment — the app does need a mic to *dictate*, but that was not what this label was reporting.

**This PR still targets 0.1.3**, whose installer has been rebuilt with the renamed binaries. The URL is unchanged, so `InstallerSha256` has been updated in this PR to match the new bytes (`66F1EDFE…`), verified by downloading the published asset and hashing it.

Also verified on a real upgrade rather than a clean install: the previous binaries are removed rather than left beside the renamed ones, the Start-menu shortcut repoints, and the app launches with both processes showing the branded names.

Sorry for the noise on the earlier theory.

stephengillie:
> 
> Executable C:\Users\validator\AppData\Local\HoldToSpeak\HoldToSpeak.exe returned exit code: -1073741515
> Executable C:\Users\validator\AppData\Local\HoldToSpeak\HoldToSpeak-engine.exe returned exit code: -1073741515 
 

PR #420823, opened 2026-08-19
New package: lsuen.sugt version 1.1.1

lsuen:
Adds SUGT (local AI gateway) portable package via x64 zip.

Validity validation metadata:
- Portable zip from GitHub Releases.
- sha256 provided.
- InstallerType: zip, nested portable with alias sugt.


stephengillie:
Close with reason: PR appears malformed.;

PR #421151, opened 2026-08-20
New version: yt-dlp.yt-dlp version 2026.08.19

Jimeszx:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/421151)

### DIFF
--- a/manifests/y/yt-dlp/yt-dlp/2026.08.19/yt-dlp.yt-dlp.installer.yaml
+++ b/manifests/y/yt-dlp/yt-dlp/2026.08.19/yt-dlp.yt-dlp.installer.yaml
@@ -8,7 +8,6 @@ Commands:
 - yt-dlp
 Dependencies:
   PackageDependencies:
-  - PackageIdentifier: DenoLand.Deno
   - PackageIdentifier: yt-dlp.FFmpeg
 ReleaseDate: 2026-08-19
 Installers:
--- a/manifests/y/yt-dlp/yt-dlp/2026.08.19/yt-dlp.yt-dlp.installer.yaml
+++ b/manifests/y/yt-dlp/yt-dlp/2026.08.19/yt-dlp.yt-dlp.installer.yaml
@@ -11,8 +11,15 @@ Dependencies:
   - PackageIdentifier: yt-dlp.FFmpeg
 ReleaseDate: 2026-08-19
 Installers:
+- Architecture: x86
+  InstallerUrl: https://github.com/yt-dlp/yt-dlp/releases/download/2026.08.19/yt-dlp_x86.exe
+  InstallerSha256: A8F91BD41452506BC81EBD2F369B186FEA0EE7075413BA00CEF9FD346A0A5D0C
 - Architecture: x64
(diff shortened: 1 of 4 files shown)


Jimeszx:
need more information

Jimeszx:
### Issue Summary

I've been troubleshooting the validation errors on this PR and would appreciate some guidance.

### What I've Tried

1. **Microsoft Defender False Positive Submission** — I submitted all 3 yt-dlp executables (x86, x64, arm64) to Microsoft for analysis.

   - **Submission ID:** `62744b1e-0575-481f-8ec5-7dc13fc936cb`
   - **Analyst response:**
     > Our scanners show no positive detection, and we have no telemetry indicators for the file(s) submitted either.
     > As such, this submission will be closed with no further action pending.

   Local Microsoft Defender scans also show no threats on my machine.

2. **Local Manifest Validation** — I validated the manifest locally and it passes without issues.

### Request for Help

Could someone from the team provide more details on which specific validation checks are still failing? I'd like to understand what needs to be fixed so I can resolve this properly.

Thanks!

mdanish-kh:
@Jimeszx In case of false positives, it usually clears out within a few days. The maintainers may have more visibility on which AV provider is issuing the blocking detection. 

I can try re-running the pipelines in the meantime to check if it's cleared out

Jimeszx:
### Investigation: Possible Cause of Installation Validation Error

According to yt-dlp's readme, a JavaScript runtime (such as Deno) is **highly recommended** for full YouTube support, but it is not a hard requirement — yt-dlp can install and run without it.

While looking at recent nightly PRs like #425142, I noticed the validation log contains:

> `Installation failed with exit code -1978334972` (`APPINSTALLER_CLI_ERROR_INSTALL_MISSING_DEPENDENCY`)
> `No suitable installer found for manifest DenoLand.Deno with version 2.9.5`

This makes me wonder if the `DenoLand.Deno` dependency might be the issue. Deno only ships x64 and arm64 Windows builds — there is no x86 version. Since yt-dlp includes an x86 variant, winget may be unable to resolve the Deno dependency for the x86 architecture during validation.

### What I've Tried in This PR

To test this hypothesis, I've **temporarily removed `DenoLand.Deno` from the package dependencies** in the latest commit, while keeping `yt-dlp.FFmpeg` (which has x86/x64/arm64 coverage, and is also listed as strongly recommended by yt-dlp). I wanted to see whether the installation validation error goes away without the Deno dependency.

### Request for Help

The `Validation-Installation-Error` label is still present, so there may be other issues I'm not seeing. Could a reviewer share the detailed installation validation log from the latest run? I'd like to understand what's still failing so I can fix it properly.

For a longer-term fix, if the Deno x86 issue is confirmed, I think the right approach would be to add Deno back as a installer dependency for x64 and arm64 only, rather than removing it entirely. But first I'd like to verify whether removing it actually resolves the installation error.

Thanks for your time!

stephengillie (approved):

PR #421552, opened 2026-08-20
New package: FKeysCreative.KeyJ version 1.5.0

zengineco:
### New package: FKeysCreative.KeyJ version 1.4.2

Key-J turns your keyboard into an instrument: every keystroke plays a note, in
any application. It also converts guitar tablature into a sequence you can play
by typing, which is the part it exists for.

- Homepage: https://f-keys.com/keyj/
- Manual: https://f-keys.com/keyj/manual/
- Source: https://github.com/zengineco/f-keys
- Release: https://github.com/zengineco/f-keys/releases/tag/keyj-v1.4.2
- Licence: MIT

Installer is NSIS, x64, from the project's own GitHub release.

#### Checklist

- [x] Have you checked that there aren't other open pull requests for the same manifest update/addition?
- [x] Does your manifest conform to the [1.x schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.12.0)?

`winget validate` passes. The installer itself was installed and uninstalled on
a Windows 11 machine and the result checked in the uninstall registry, so the
install box is ticked on evidence. The CLA box is left unticked because that is
not something the submitter can confirm from the machine the manifest was
prepared on - the bot will say, and it will be signed if it has not been.

#### Why the version moved twice while this PR was open

It was opened as 1.4.0. Two problems were found before it merged, and both are
the kind that cannot be fixed after people have installed:

1. **1.4.0** carried a Windows `appId` left over from before the project was a
   company. Windows writes that into the uninstall registry and every later
   installer must match it to be recognised as an upgrade. **1.4.1** changed it.
2. **1.4.1** revealed the second: installing silently over a running Key-J
   exited 0 and left a mixed install. Windows held the executable and the app
   bundle open, NSIS skipped both, replaced everything else and reported
   success. Since `winget` installs silently by default, that is the upgrade
   path most users would take. **1.4.2** stops a silent install when the app is
   running and returns 1618, `ERROR_INSTALL_ALREADY_RUNNING`. Verified on a
   real machine: with Key-J running the installer exits 1618 and the install
   directory is untouched.

Also corrected earlier in this PR: `InstallerType` was submitted as `nsis`,
which `winget validate` rejected. It is `nullsoft`.


zengineco:

vince-gonzalez:
Superseded by #427477, which is Key-J 1.6.0 filed under the account that owns the repository the manifest points at.

Two things changed beyond the account. It is the current release rather than 1.5.0, and the licence differs: Key-J up to 1.5.0 was MIT, and from 1.6.0 it is proprietary, so this manifest's `License: MIT` would not have been true of any version I could sensibly file today.

Closing this one. The `Validation-Shell-Execute` here was never accompanied by a comment, so it is unaddressed rather than fixed; I have said as much on the new PR.

PR #421610, opened 2026-08-20
New version: Oven-sh.Bun.BaselineProfile version 1.4.0

UnownBot:
Automatically updated by 🌠 Anthelion.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/421610)


UnownPlain:
There are over 2k PRs currently open and pretty much only [1 person that actively reviews PRs](https://github.com/microsoft/winget-pkgs/discussions/333858) so it might take some time. The Manifest-Metadata-Consistency label is currently blocking it from being automerged. Bun 1.4's release notes are over 400k characters long and the limit for the `ReleaseNotes` field is 10k characters which is why I'm not adding them. If you really need this version I could put something in `ReleaseNotes` and remove it in another PR to get this merged quicker.

greythx:
It's fine, I was actually on `Bun Baseline Profile` so didn't get the update. 

But now since `baseline` builds are null (They only retained still to give users on them updates still) 

The `Bun Profile` is identical to this. So I ended up re-installing bun and switching to that version of the package. (Again identical practically)

Just thought to nudge this as some people may be on this and be waiting for an update. It's not that important imo, if anyone absolutely knows about the 1.4 update, he can simply switch.

Take your time.

stephengillie (approved):

PR #421824, opened 2026-08-21
New package: kingkate2009-droid.ai-switch version 2.2.0

SpecterShell:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/421824)

### DIFF
--- a/manifests/k/kingkate2009-droid/ai-switch/2.2.0/kingkate2009-droid.ai-switch.installer.yaml
+++ b/manifests/k/kingkate2009-droid/ai-switch/2.2.0/kingkate2009-droid.ai-switch.installer.yaml
@@ -6,7 +6,7 @@ PackageVersion: 2.2.0
 InstallerType: zip
 NestedInstallerType: portable
 NestedInstallerFiles:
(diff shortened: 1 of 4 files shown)


denelon:
@SpecterShell During manual moderation, I noticed that AI Switch v2.2.0 creates a shared `requests.Session` with TLS certificate verification disabled:

https://github.com/kingkate2009-droid/ai-switch/blob/v2.2.0/core/providers.py#L37-L48

That session is subsequently used for provider requests carrying API credentials in headers such as `Authorization`, `api-key`, and `x-api-key`. Because this application is specifically intended to manage and validate API keys, disabling certificate verification globally could expose those credentials to interception by a malicious proxy or compromised network.

The WinGet validation pipeline has passed, and I am not identifying this as a specific Windows Package Manager policy violation. However, I would like this concern resolved or clarified before applying moderator approval. Could the publisher enable normal TLS certificate verification by default and make any support for self-signed certificates an explicit, narrowly scoped opt-in setting?

Template: msftbot/security/tlsCertificateVerification

PR #422124, opened 2026-08-21
Update: SerZhyAle.StreamsPlayer to 26.0821.1208

SerZhyAle:
## 📖 Description

STREAMS Player (`SerZhyAle.StreamsPlayer`) 26.0821.1208. Portable zip, x64, published as a GitHub Release by the project's own release workflow.

This version exists to fix a total playback failure. 26.0820.1828 - the version currently in this repository - shipped a WPF runtime that breaks `MediaElement` network audio outright, so the application opens, loads its catalog and plays nothing at all, on every machine. The runtime is pinned back in 26.0821.1208 and playback works again. Anyone who installs or upgrades the current manifest today gets a silent application, which is why this is worth a look ahead of a normal feature update. Release notes are in the en-US, ru-RU and uk-UA locale manifests.

Verification on the exact `InstallerUrl` in the installer manifest, beyond the local install below: the asset was downloaded from that URL, its SHA256 recomputed and compared to `InstallerSha256` (`8A9360C1...3CD92`, match), and the archive checked to contain the declared `NestedInstallerFiles` entry `StreamsPlayer.exe` at its root. The installed payload was then launched and played a live stream, confirming the fix is in the artifact this manifest points at and not only in the source.

## ✅ Checklist

  - Resolves #[Issue Number]

## 📦 Manifest Checklist

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/422124)

### DIFF
--- a/manifests/s/SerZhyAle/StreamsPlayer/26.0821.1208/SerZhyAle.StreamsPlayer.installer.yaml
+++ b/manifests/s/SerZhyAle/StreamsPlayer/26.0821.1208/SerZhyAle.StreamsPlayer.installer.yaml
@@ -1,17 +1,14 @@
 ﻿# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
 PackageIdentifier: SerZhyAle.StreamsPlayer
 PackageVersion: 26.0821.1208
-InstallerType: zip
-NestedInstallerType: portable
+InstallerType: inno
 Platform:
 - Windows.Desktop
 MinimumOSVersion: 10.0.17763.0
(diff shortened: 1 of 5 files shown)


SerZhyAle:
Related: #422156 asks for 26.0820.1828 - the version this replaces - to be removed rather than left installable. That build plays no media at all on any machine (root cause reported as dotnet/wpf#11856), so it is not merely superseded, it never worked. This pull request stands on its own; the two are independent.

stephengillie (approved):

PR #422222, opened 2026-08-21
New package: scarlone.WinBoost version 0.1.0

scarlone:
### New package: `scarlone.WinBoost` version `0.1.0`

WinBoost is a portable, single-executable Windows optimizer with a non-destructive
preview and a per-session rollback journal.

- Repository: https://github.com/scarlone/WinBoost
- Release: https://github.com/scarlone/WinBoost/releases/tag/v0.1.0
- License: MIT

#### Notes for reviewers

- `InstallerType: portable`, x64 only. The published installer is the
  **framework-dependent** build (~0.75 MB); the .NET 8 Desktop Runtime is declared as a
  package dependency (`Microsoft.DotNet.DesktopRuntime.8`) rather than bundled.
- **The executable is not code signed yet.** SmartScreen may warn on first launch. This
  is stated in `InstallationNotes` so it is visible to users at install time.
- **The package modifies the system**: registry values, service start types, network
  settings and power plans. It does so only when the user explicitly applies a tweak.
  The application runs unelevated (`asInvoker`) and requests elevation per session, only
  when a selected tweak requires it. Every applied change is journaled with its previous
  value so it can be rolled back from the UI. `InstallationNotes` states this too.

#### Verification performed

- `winget validate --manifest` passes (only the expected warning about the dependency
  not being resolvable locally).
- The `InstallerSha256` was verified by downloading the release asset anonymously and
  recomputing the hash; it matches the `SHA256SUMS.txt` published alongside it.
- The downloaded executable was launched and starts correctly.

#### Not verified

- `winget install --manifest` was **not** run locally: enabling `LocalManifestFiles`
  requires administrator privileges that were unavailable in the environment used to
  prepare this submission. Flagging it rather than silently ticking the box.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/422222)

### DIFF
--- a/manifests/s/scarlone/WinBoost/0.1.0/scarlone.WinBoost.installer.yaml
+++ b/manifests/s/scarlone/WinBoost/0.1.0/scarlone.WinBoost.installer.yaml
@@ -1,0 +1,29 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.28.0.schema.json
+PackageIdentifier: scarlone.WinBoost
+PackageVersion: 0.1.0
+InstallerLocale: en-US
+Platform:
+- Windows.Desktop
+# winget stesso richiede Windows 10 1809: dichiarare meno sarebbe una promessa vuota.
+MinimumOSVersion: 10.0.17763.0
+# "portable": winget copia l'exe in %LOCALAPPDATA%\Microsoft\WinGet\Packages e crea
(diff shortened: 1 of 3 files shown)


Trenly:
Close with reason: Stale; No CLA

PR #422233, opened 2026-08-21
New version: StruisICT.InSearch version 0.7.0

Struis112:
Adds **StruisICT.InSearch** version 0.2.1 (MSI installer).

- [x] Manifest validated with `winget validate` (passed)
- [x] Installation tested with `winget install --manifest`
- [x] Schema 1.12.0; `InstallerSha256` and `ProductCode` from the published GitHub release MSI

This supersedes the 0.2.0 submission — the bundled CLI helper now exits 0 when run with no arguments (it previously returned a non-zero usage code), which should clear the earlier automated executable-validation error.

InSearch is a real-time, content-aware file search app. Source: https://github.com/StruisICT/InSearch
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/422233)

### DIFF
--- a/manifests/s/StruisICT/InSearch/0.2.1/StruisICT.InSearch.installer.yaml
+++ b/manifests/s/StruisICT/InSearch/0.2.1/StruisICT.InSearch.installer.yaml
@@ -1,0 +1,20 @@
+# Created with: scripts/Update-WingetManifest.ps1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: StruisICT.InSearch
+PackageVersion: 0.2.1
+MinimumOSVersion: 10.0.0.0
+InstallerType: msi
+Scope: machine
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+ReleaseDate: 2026-08-21
+Installers:
+- Architecture: x64
+  InstallerUrl: https://github.com/StruisICT/InSearch/releases/download/v0.2.1/InSearch-0.2.1-x86_64.msi
+  InstallerSha256: 39B0310B27E9532DD626E834A8E0CFF2AEA95461C236C6CD6B70FCBBD6CF80E2
+  ProductCode: '{F6C2D17B-66D3-43E4-A832-C76AEC28608A}'
+ManifestType: installer
+ManifestVersion: 1.12.0
--- a/manifests/s/StruisICT/InSearch/0.2.1/StruisICT.InSearch.locale.en-US.yaml
+++ b/manifests/s/StruisICT/InSearch/0.2.1/StruisICT.InSearch.locale.en-US.yaml
@@ -1,0 +1,38 @@
+# Created with: scripts/Update-WingetManifest.ps1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.12.0.schema.json
+
+PackageIdentifier: StruisICT.InSearch
+PackageVersion: 0.2.1
+PackageLocale: en-US
+Publisher: Struis ICT
+PublisherUrl: https://struisict.com
+PublisherSupportUrl: https://github.com/StruisICT/InSearch/issues
+PackageName: InSearch
+PackageUrl: https://github.com/StruisICT/InSearch
+License: MIT
+LicenseUrl: https://github.com/StruisICT/InSearch/blob/main/LICENSE
+Copyright: Copyright (c) 2026 Struis ICT
+ShortDescription: Real-time, content-aware file search â€” find files and search inside them.
+Description: |-
+  InSearch is a real-time file search tool that finds files on disk and searches
+  inside them, streaming matches as it goes. It searches plain text and logs plus
+  PDF, Word (docx) and Excel (xls/xlsx/ods), with matches reported per line or per
+  timestamp-to-timestamp block. Query with substring, regex, all-words (AND),
+  any-words (OR) and exclude (NOT); filter by name, extension, size and date;
+  watch folders for changes (log-tailing); export results; and search a folder
+  straight from the Explorer right-click menu. Everything runs locally.
+Moniker: insearch
+Tags:
+- search
+- file-search
+- content-search
+- grep
+- find
+- logs
+- utility
+- windows
+ReleaseNotes: |-
+  See the full release notes at https://github.com/StruisICT/InSearch/releases/tag/v0.2.1
+ReleaseNotesUrl: https://github.com/StruisICT/InSearch/releases/tag/v0.2.1
+ManifestType: defaultLocale
+ManifestVersion: 1.12.0
--- a/manifests/s/StruisICT/InSearch/0.2.1/StruisICT.InSearch.yaml
+++ b/manifests/s/StruisICT/InSearch/0.2.1/StruisICT.InSearch.yaml
@@ -1,0 +1,8 @@
+# Created with: scripts/Update-WingetManifest.ps1
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.version.1.12.0.schema.json
+
+PackageIdentifier: StruisICT.InSearch
+PackageVersion: 0.2.1
+DefaultLocale: en-US
+ManifestType: version
+ManifestVersion: 1.12.0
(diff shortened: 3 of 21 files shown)


Struis112:
Updated this PR in place from **0.4.0 → 0.6.0** (same branch, to keep its queue position).

0.6.0 adds sortable result columns, a preview pane, an expanded CLI, and an opt-in update check. The manifest passes local `winget validate`; SHA256 + ProductCode are read from the published 0.6.0 MSI, which installs/uninstalls cleanly. Thanks!

stephengillie:
> 
> Executable C:\Program Files\InSearch\insearch-cli.exe returned exit code: -1073741515
> Executable C:\Program Files\InSearch\InSearch.exe returned exit code: -1073741515 
 
Struis112:
Thanks for the validation output — that pinpointed it. The `STATUS_DLL_NOT_FOUND` (0xC0000135) was exactly right: both executables dynamically linked the Visual C++ runtime (`vcruntime140.dll`), so on a clean machine without `Microsoft.VCRedist.2015+` they couldn't start.

**Fixed in 0.6.1**, which this PR now targets: the binaries are built with `-C target-feature=+crt-static` (static CRT), so they no longer import any `vcruntime`/`msvcp`/`ucrtbase` DLL and are fully self-contained — verified against the published 0.6.1 exes. They now launch on a clean install with no VCRedist. The manifest passes local `winget validate`; SHA256 + ProductCode are read from the published 0.6.1 MSI. Would appreciate a re-run when convenient — thank you!

stephengillie:
Manual Validation ended with: 
> PS C:\Users\User\Desktop> &"C:\Program Files\InSearch\insearch-cli.exe"
> InSearch CLI — find files and search inside them.
> 
> Usage: insearch-cli <pattern> [roots...] [options]
>   <pattern>   text to search for (substring by default)
>   roots       one or more files/folders (default: current directory)
> 
> Query:
>   --regex             treat <pattern> as a regular expression
>   --all-words         every space-separated word must appear (AND)
>   --any-words         any of the space-separated words may appear (OR)
>   --exclude "a b"      words that must NOT appear (NOT)
>   --whole-word        match whole words only
>   --case-sensitive    force case-sensitive (default: smart case)
>   --ignore-case       force case-insensitive
>   --block             report one result per timestamp-to-timestamp block
>   --gitignore         honour .gitignore / hidden-file rules (default: search all)
> 
> File filters:
>   --name <glob>       filename glob, e.g. *.log
>   --name-regex        treat --name as a regex
>   --ext <a,b>         only these extensions
>   --exclude-ext <a,b> not these extensions
>   --min-kb <n>        minimum size (KB)
>   --max-kb <n>        maximum size (KB)
>   --days <n>          modified within the last n days
>   --after <when>      file modified at/after <when>
>   --before <when>     file modified at/before <when>
> 
> Entry-time filter (timestamp inside each matching line/block):
>   --entry-after <when>   keep entries at/after <when>
>   --entry-before <when>  keep entries at/before <when>
>   (<when> is YYYY-MM-DD or 'YYYY-MM-DD HH:MM:SS', local time)
> 
> Output:
>   --json              one JSON object per line ({"path","line","text"})
>   --count             print only the number of matches
>   -l, --files-with-matches   print only the matching file paths (unique)
>   -h, --help          show this help
> PS C:\Users\User\Desktop> &"C:\Program Files\InSearch\insearch-cli.exe"  --version
> error: unknown flag: --version
> 
> Run `insearch-cli --help` for usage.
> PS C:\Users\User\Desktop> &"C:\Program Files\InSearch\insearch-cli.exe"  version
> C:\Users\User\Desktop\cmds.ps1:40:              Out-Log 'Updating Defender Security Intelligence version.'
> C:\Users\User\Desktop\cmds.ps1:74:                      $SigVer = (Get-MpComputerStatus).QuickScanSignatureVersion
> C:\Users\User\Desktop\cmds.ps1:80:                      Out-ErrorData $DefenderThreat "Defender (with Security Intelligence version $((Get-MpComputerStatus).QuickScanSignatureVersion))"
> C:\Users\User\Desktop\cmds.ps1:90: ' + 'Defender Security Intelligence version: ' + $SigVer)
> C:\Users\User\Desktop\cmds.ps1:171:                     $SigVer = (Get-MpComputerStatus).QuickScanSignatureVersion
> C:\Users\User\Desktop\cmds.ps1:177:                     Out-ErrorData $DefenderThreat "Defender (with Security Intelligence version $SigVer)"
> C:\Users\User\Desktop\cmds.ps1:184: ' + 'Defender Security Intelligence version: ' + $SigVer)
> C:\Users\User\Desktop\cmds.ps1:209:                             Get-installedVersions
> 8 match(es).
> PS C:\Users\User\Desktop> Get-ARPTable insearch
> 
> 
> DisplayName    : InSearch
> DisplayVersion : 0.6.1
> Publisher      : Struis ICT
> ProductCode    : {E566B6B2-6728-4549-8172-00CC440719CC}
> Scope          : Machine
> 
> 
> 
> PS C:\Users\User\Desktop> # Hi @Struis112,
> PS C:\Users\User\Desktop> #
> PS C:\Users\User\Desktop> # The CLI application seems to work normally. But the GUI application doesn't run when launched. Is this expected?

Struis112:
Yes, that's expected on the validation VM — thanks for checking!

`InSearch.exe` is a graphical (egui / OpenGL via `glow`) desktop app, so it needs a GPU and an interactive display to open its window. The validation environment is headless (no GPU / no desktop session), so the OpenGL context can't be created there.

Rather than crash in that situation, InSearch handles it gracefully: it writes a short diagnostic to stderr explaining that no display/GPU is available and exits with code **0** (success). That's why you see it "not run" — there's simply no window to show on a headless runner, but it doesn't error out. On a normal Windows desktop it launches into the full GUI.

The `insearch-cli.exe` you tested is the headless companion and works anywhere, which is why it behaves as expected in the sandbox.

So: CLI works everywhere; GUI is display-dependent by design and exits cleanly (0) when no display is present. Nothing further needed on my end — happy to answer anything else. Thank you for the manual validation!

Struis112:
Follow-up on the "GUI doesn't run when launched" question — I looked into it properly rather than leaving it at "expected", and I want to correct my earlier answer.

**What was actually happening.** The validation VM isn't headless: it has a desktop session, just no GPU. `InSearch.exe` used an OpenGL renderer (egui's `glow` backend), and a Windows machine without a GPU driver only provides Microsoft's OpenGL 1.1 software implementation, which is too old to create a context. So the GUI printed a diagnostic and exited 0 instead of opening a window. That is the same experience a user would get on a Hyper-V / cloud VM or a server on the Basic Display Adapter — and we agree the app should work there too, not only on machines with a GPU.

**What changed.** The Windows build now renders through DirectX 12 (egui's `wgpu` backend). DirectX 12 exposes WARP, Windows' built-in CPU rasterizer, so on a machine without a GPU the app falls back to software rendering and opens normally; with a GPU it uses the GPU. The OpenGL path is kept as a fallback (`--renderer glow`), and `--software-gpu` forces WARP for diagnostics. Verified locally on both a GPU and a forced-WARP run; the CLI is unchanged.

**What this means for the PR.** The change is merged-in-progress upstream and will ship as the next release. I'll update this PR in place to that version (same branch, as before) once the MSI is published, so manual validation should then show the GUI window opening on the VM as well. Until then the 0.6.1 manifest here remains valid — the GUI exits cleanly (0) on the VM as described. Thanks again for the manual check; it pointed at a real gap.

Struis112:
Updated this PR in place from **0.6.1 → 0.7.0** (same branch, to keep its queue position).

0.7.0 is the release described in my previous comment: the GUI now renders through DirectX 12 with a WARP (software) fallback, so `InSearch.exe` opens a window on machines without a GPU — including the validation VM — instead of exiting with a diagnostic. It also adds a global excluded-file-types default, "skip hidden" / "skip Windows system folders" toggles, and `insearch-cli --version` (the flag tried during the earlier manual validation).

The manifest passes local `winget validate`; SHA256 + ProductCode are read from the published 0.7.0 MSI (hash matches the release's `SHA256SUMS.txt`), which installs/uninstalls cleanly. Thanks!

PR #423242, opened 2026-08-24
New package: Erryb95.VideoDownloaderConverter version 0.3.3

Erryb95:
### New package

`Erryb95.VideoDownloaderConverter` 0.3.3 — a desktop app that downloads video
and audio from the sites yt-dlp supports and converts files you already have.
yt-dlp and FFmpeg ship with it, so it has no external dependencies.

- Repository: https://github.com/Erryb95/video-downloader-converter
- Release: https://github.com/Erryb95/video-downloader-converter/releases/tag/v0.3.3

### Checklist

- [x] Manifest validated locally: `winget validate --manifest .` → *Manifest validation succeeded.*
- [x] Conforms to the 1.6.0 schema.
- [x] No other open pull request for this package — it is a new one.
- [x] `InstallerSha256` verified by downloading the asset and recomputing the
      hash, not by copying it from the release page. It also matches the
      `SHA256SUMS` the release publishes.
- [ ] **Not** tested with `winget install --manifest`: `LocalManifestFiles` is
      disabled on the machine used and enabling it needs elevation. Instead the
      installer itself was run with `/S`, which is what the silent install
      modes resolve to.

### What was verified by hand

The NSIS installer is per-user and installs unattended: `setup.exe /S` returned
exit code 0 and registered under
`HKCU\Software\Microsoft\Windows\CurrentVersion\Uninstall`. The installed copy
was then run and downloaded a video successfully. `uninstall.exe /S` returned 0
and left neither the directory nor the registry entry behind.

That test is also why `AppsAndFeaturesEntries` is present: the bundle registers
its `DisplayName` as **Video Downloader**, shorter than the package name, so
without it winget cannot match an installed copy back to this package.

### Notes for reviewers

- The installer is **not code-signed**. Individual developers cannot currently
  obtain a Microsoft Artifact Signing certificate outside the US and Canada.
- Every release publishes `SHA256SUMS` and a GitHub build attestation, so the
  artifact can be traced to the commit and workflow that produced it:
  `gh attestation verify <file> --repo Erryb95/video-downloader-converter`
- The app bundles FFmpeg under **GPLv3**. The licence text and the written
  offer for corresponding source ship inside the installer, and the FFmpeg
  build is pinned to a specific tagged artifact rather than a rolling one.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/423242)

### DIFF
--- a/manifests/e/Erryb95/VideoDownloaderConverter/0.3.3/Erryb95.VideoDownloaderConverter.installer.yaml
+++ b/manifests/e/Erryb95/VideoDownloaderConverter/0.3.3/Erryb95.VideoDownloaderConverter.installer.yaml
@@ -1,0 +1,27 @@
+# Created with the winget schema 1.6.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+
+PackageIdentifier: Erryb95.VideoDownloaderConverter
+PackageVersion: 0.3.3
+InstallerType: nullsoft
+Scope: user
+InstallModes:
+  - interactive
+  - silent
+  - silentWithProgress
+UpgradeBehavior: install
+ReleaseDate: 2026-08-24
+Installers:
+  - Architecture: x64
+    InstallerUrl: https://github.com/Erryb95/video-downloader-converter/releases/download/v0.3.3/Video.Downloader_0.3.3_x64-setup.exe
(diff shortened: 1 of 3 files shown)


Erryb95:
Withdrawing this — the author would rather not distribute through winget for now. Nothing wrong on your side; closing so it does not sit in the review queue. Apologies for the noise.

PR #423307, opened 2026-08-24
New package: RudySource.Dirgo version 0.7.1

RudySource:
Adds Dirgo 0.7.1 as a portable x64 package from the immutable GitHub release ZIP.

Upstream: https://github.com/RudySource/Dirgo
Release: https://github.com/RudySource/Dirgo/releases/tag/v0.7.1

Windows integration:
- installs dgo.exe with the dgo command alias
- retains the versioned DirgoPredictor DLL and module manifest from the release archive
- records dgo.exe and both predictor files in InstallationMetadata with exact SHA-256 values
- PowerShell 7+ users can run dgo setup after installation

Validation performed:
- YAML parsed successfully
- release ZIP SHA-256 matched published SHA256SUMS
- dgo.exe and predictor paths verified in the archive
- individual dgo.exe, DLL, and PSD1 hashes matched the manifest
- schema headers and package versions agree across all three manifests
- no private paths, credentials, or internal artifacts included


RudySource:

PR #423450, opened 2026-08-24
Remove: Codesomelabs.trau

RomkaLTU:
Trau is no longer publicly distributed; the source repository is private and all release asset URLs referenced by these manifests return 404. Please remove all versions of Codesomelabs.trau from the catalog.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/423450)


Trenly (changes requested):
Each version will have to be removed in a separate PR - only one manifest can be changed per PR

PR #423478, opened 2026-08-24
Update existing agentic workflows to reference github app check instead of old ado validation logs

yao-msft:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/423478)


denelon:
This is a strong architectural improvement over discovering Azure DevOps builds through PR comments and then reading public build logs. Moving the three workflows to WinGetValidator GitHub Check Runs makes the evidence directly associated with GitHub, binds it to the current pull-request head SHA, removes the `dev.azure.com` network dependency, and reduces the amount of URL parsing and external content the agents need to trust.

The safety posture looks good:

- `pull_request_target` remains paired with `checkout: false`.
- The deterministic collector is read-only and requires the exact `wingetvalidator-prod` app slug and current head SHA.
- Check output is still treated as untrusted evidence.
- Missing, stale, ambiguous, or conflicting evidence fails closed with `noop`.
- Existing security-label, no-binary-download, recommend-only, and one-comment-per-head-SHA safeguards remain intact.
- Including `neutral` only for the URL and domain checks matches the validator's manual-review behavior.

I do see one reentrancy/reliability issue that I think should be fixed before merge. `manifest-validation-diagnosis` explicitly serializes per PR with `cancel-in-progress: false` and `queue: max`, but `missing-dependency-assist` and `wingetbot-pr-triage` still compile with the default `cancel-in-progress: true`.

For label-triggered workflows, GitHub places a run in the concurrency group before the generated job-level `if` is evaluated. That means a later, unrelated label event can cancel an active diagnosis even though the later run ultimately skips. Cancelled Missing Dependency Assist and Wingetbot PR Triage jobs are already visible on https://github.com/microsoft/winget-pkgs/pull/424185.

I recommend adding the same explicit per-PR serialization to both workflows:

```yaml
concurrency:
  group: "gh-aw-${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}"
  cancel-in-progress: false
  queue: max
```

Their current-head and comment/SHA gates will then make stale queued runs safely `noop`, giving us reliable sequential reentrancy instead of cancellation during policy-label cascades.

One secondary hardening item: the collector reads a single page of at most 100 Check Runs. Busy PRs can accumulate many gh-aw job checks from label events, so the validator checks could eventually fall outside that page. A bounded pagination strategy or filtering by the validator App ID would preserve the current resource bound without silently losing evidence.

Overall: the migration is materially safer and the evidence model is better, but I would address the concurrency behavior before treating the workflows as fully reentrant.

_(Drafted with GitHub Copilot assistance.)_

yao-msft:
> This is a strong architectural improvement over discovering Azure DevOps builds through PR comments and then reading public build logs. Moving the three workflows to WinGetValidator GitHub Check Runs makes the evidence directly associated with GitHub, binds it to the current pull-request head SHA, removes the `dev.azure.com` network dependency, and reduces the amount of URL parsing and external content the agents need to trust.
> 
> The safety posture looks good:
> 
> * `pull_request_target` remains paired with `checkout: false`.
> * The deterministic collector is read-only and requires the exact `wingetvalidator-prod` app slug and current head SHA.
> * Check output is still treated as untrusted evidence.
> * Missing, stale, ambiguous, or conflicting evidence fails closed with `noop`.
> * Existing security-label, no-binary-download, recommend-only, and one-comment-per-head-SHA safeguards remain intact.
> * Including `neutral` only for the URL and domain checks matches the validator's manual-review behavior.
> 
> I do see one reentrancy/reliability issue that I think should be fixed before merge. `manifest-validation-diagnosis` explicitly serializes per PR with `cancel-in-progress: false` and `queue: max`, but `missing-dependency-assist` and `wingetbot-pr-triage` still compile with the default `cancel-in-progress: true`.
> 
> For label-triggered workflows, GitHub places a run in the concurrency group before the generated job-level `if` is evaluated. That means a later, unrelated label event can cancel an active diagnosis even though the later run ultimately skips. Cancelled Missing Dependency Assist and Wingetbot PR Triage jobs are already visible on #424185.
> 
> I recommend adding the same explicit per-PR serialization to both workflows:
> 
> ```yaml
> concurrency:
>   group: "gh-aw-${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}"
>   cancel-in-progress: false
>   queue: max
> ```
> 
> Their current-head and comment/SHA gates will then make stale queued runs safely `noop`, giving us reliable sequential reentrancy instead of cancellation during policy-label cascades.
> 
> One secondary hardening item: the collector reads a single page of at most 100 Check Runs. Busy PRs can accumulate many gh-aw job checks from label events, so the validator checks could eventually fall outside that page. A bounded pagination strategy or filtering by the validator App ID would preserve the current resource bound without silently losing evidence.
> 
> Overall: the migration is materially safer and the evidence model is better, but I would address the concurrency behavior before treating the workflows as fully reentrant.
> 
> _(Drafted with GitHub Copilot assistance.)_

resolved by adding concurrency group and the github app id

denelon:
This is a good direction overall: moving from the old ADO evidence path to the GitHub App checks improves provenance, removes the `dev.azure.com` dependency, and preserves the important `pull_request_target` + `checkout: false`, security-label abort, per-PR serialization, and per-head idempotency guardrails.

I found two functional gaps that I think should be addressed before merging:

1. **Bind validation evidence to the PR number.** The collector verifies the App ID, slug, and head SHA, but does not verify the validation operation's `PullRequestNumber`. The recorded `output.pullRequestNumber` echoes the target PR rather than extracting it from the check output. Because validator check suites can have an empty `pull_requests` array, a commit SHA reused across PRs could select evidence from another PR. Please extract the authoritative `PullRequestNumber` from the `10. Validation Completed` output, require it to equal the target PR, and fail closed when that binding is absent.

2. **Handle minimum-version dependency failures.** The new mandatory gate accepts `DependenciesNotFound`, but `Validation-Missing-Dependency` also covers `DependenciesMinVersion` / “No Suitable Minimum Version.” As written, those actionable failures will always produce `noop`. Please accept and explicitly classify both result types.

With those two evidence-binding/classification issues fixed, I think this will be a solid improvement.

Template: msftbot/review/agenticWorkflow

denelon (approved):

PR #424771, opened 2026-08-26
New package: MichaelFlynn.StormSewer version 0.10.0

mf4633:
### Pull request title follows the repository convention: `New package: MichaelFlynn.StormSewer version 0.9.1`

- [x] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)? — will sign when the bot prompts.
- [x] Have you checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/change? — this is a new package; no existing manifests for it.
- [x] Have you validated your manifest locally with `winget validate --manifest <path>`? — passes on winget v1.29.290.
- [x] Have you tested your manifest locally with `winget install --manifest <path>`? — not yet; the installer itself has been installed and run from the release, but I have not exercised the install through the manifest.
- [x] Does your manifest conform to the [1.6 schema](https://github.com/microsoft/winget-cli/tree/master/schemas/JSON/manifests/v1.6.0)?

### About this package

StormSewer is a free, GPL-3.0 storm sewer design and analysis application for
gravity pipe networks — Rational method hydrology, Manning capacity, standard-step
HGL/EGL backwater, and HEC-22 inlet interception. I am the author and maintainer.

- Homepage / source: https://github.com/mf4633/stormsewer
- Release: https://github.com/mf4633/stormsewer/releases/tag/v0.9.1
- Installer: `StormSewer-0.9.1-setup.exe` (Inno Setup 6, per-user, `PrivilegesRequired=lowest`)

### Notes for reviewers

- `Scope: user` — the installer is per-user and requires no elevation.
- `ProductCode` is the Inno Setup `AppId` with the `_is1` suffix, matching the
  uninstall registry key the installer writes.
- The installer is **not code-signed**, so expect a SmartScreen prompt outside
  the validation sandbox.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/424771)


mf4633:
Updated this PR from 0.9.1 to **0.9.3**, which fixes the cause of the `Validation-Executable-Error`.

The application was built against OpenGL only, so it failed to start on a machine with no usable GL driver — which includes the validation sandbox. 0.9.3 compiles both renderers and tries Direct3D 12 (with a software adapter as a last resort) before falling back to OpenGL, so it now starts without a GPU. It also shows a dialog explaining the problem instead of exiting silently if no backend is available at all.

Thanks for the automated catch — it found a real bug that would equally have hit users on remote desktop and VDI.

Manifests re-validated locally with `winget validate`.

mf4633:
Updated to **0.9.5** and, I believe, addressed the actual cause of `Validation-Executable-Error`.

I had read the label as "the executable would not run" and fixed that side (0.9.3 added a Direct3D path, 0.9.5 bundles a software OpenGL renderer so the app starts with no GPU at all). But the docs define the label as *the test was unable to locate the primary application after install*. Reproducing locally: the Inno installer registers the Apps & Features entry as **"StormSewer version 0.9.5"** (Inno's default `UninstallDisplayName`), while the manifest's `PackageName` is **"StormSewer"**, so installed-app correlation could not match the entry.

This revision adds an `AppsAndFeaturesEntries` block spelling out the ARP entry exactly (DisplayName, Publisher, DisplayVersion, ProductCode). The installer itself is unchanged; future releases will register under the plain package name. `winget validate` passes locally.

Also: the `Needs-CLA` label is stale — the CLA was signed on this PR on 2026-08-26 (`license/cla` check is green).

stephengillie:
> 
> Executable C:\Users\validator\AppData\Local\Programs\StormSewer\StormSewer.exe returned exit code: -1073741515
> Executable C:\Users\validator\AppData\Local\Programs\StormSewer\mesa\StormSewer.exe returned exit code: -1073741515 
 
mf4633:
Fixed, and the PR now points at 0.9.8 (the 0.9.5 manifests are removed).

The exit code was exactly right. `StormSewer.exe` imported `VCRUNTIME140.dll` — rustc links it by default on `x86_64-pc-windows-msvc` — and that DLL comes with the Visual C++ Redistributable rather than with Windows, so on your clean validation image the loader failed before `main()`. Both reported paths are the same binary: `mesa\StormSewer.exe` is a copy of the app that sits beside a bundled Mesa `opengl32.dll` for machines with no GPU driver, so it failed identically.

Rather than declare a `Microsoft.VCRedist.2015+.x64` dependency in the manifest, 0.9.8 links the CRT statically (`-C target-feature=+crt-static`). The dependency would have satisfied winget alone and left users of the portable zip, Chocolatey and Scoop with an app that still would not start, so this seemed like the better fix — and it keeps the manifest dependency-free.

Verified against the published 0.9.8 installer, not just a local build:

- `StormSewer.exe` and `mesa\StormSewer.exe` after a silent per-user install: 14 imports, all OS DLLs, no VC++ runtime (PE import table, not a launch test — every build machine already has the redist, which is how this shipped in the first place).
- The app starts on hardware wgpu and on the bundled Mesa llvmpipe (`--check-renderer` exits 0 on both).
- `AppsAndFeaturesEntries` matches what the installer actually writes to the uninstall key on a real install: DisplayName `StormSewer`, DisplayVersion `0.9.8`, Publisher `Michael Flynn`, ProductCode `{A7B3C4D5-E6F7-4890-ABCD-EF1234567890}_is1`.

A CI check now reads the import table of every shipped binary and fails the release if a redistributable dependency comes back.

Thanks for the detail in the automated report — the exit code plus the two paths was enough to find this without guessing, and it was breaking bare VMs and VDI sessions too, not just validation.

Trenly:
[Policy] CLA Signed

Trenly:
cc @denelon @stephengillie

### Policy-Test-1.8 Analysis

2. **Flagged Substring:** No banking-related substring is present; the entire storm-sewer engineering description was classified by the diagnostic `All banking applications require review` heuristic.
3. **Recommendation:** waiver should be applied — StormSewer is a civil-engineering and EPA SWMM modeling tool. Its description contains no banking, payment, purchase, or financial-account functionality, so Policy 1.8 does not apply.

---
*Authored by Copilot*

mf4633:
Updated while this was waiting for review: the project was renamed from StormSewer to AquaSWMM in 0.11.0 (released today), so this new-package PR now adds `MichaelFlynn.AquaSWMM` 0.11.0 instead of `MichaelFlynn.StormSewer` 0.10.0. StormSewer was never merged, so no existing identifier is affected. The installer keeps the same Inno AppId/ProductCode. `winget validate` passes locally. The description is unchanged apart from the name, so the Policy-Test-1.8 waiver reasoning above still applies.

Trenly:
@denelon - Policy waiver needs to be re-applied because of changes pushed

PR #425045, opened 2026-08-27
New package: ubranch.Jot version 0.2.2

ubranch:
### New package: ubranch.Jot 0.2.2

Jot is a push-to-talk dictation app for Windows. Hold a key, speak, let go, and the text lands at the cursor in whatever app you were already using. Audio goes from the user's PC straight to the Gemini API with their own key — no middleman server and no account.

- Repository: https://github.com/ubranch/jot-gemini-transcribe-Windows
- Release: https://github.com/ubranch/jot-gemini-transcribe-Windows/releases/tag/v0.2.2
- License: Apache-2.0

#### Checklist

- [x] Manifest passes `winget validate`
- [x] `InstallerSha256` verified against the published asset by downloading it
- [x] Installer URL is a permanent GitHub release asset
- [x] I am the publisher of this package

#### Note for the reviewer

The binary is **not code-signed**. It is an open-source project whose author is
in a country outside the eligibility list for Azure Trusted Signing, and a
traditional OV certificate with the required hardware token is not currently
practical. The project states this plainly in its own README and release notes
rather than hiding it, and `InstallerType: zip` with a nested portable
executable means nothing is written outside winget's own package directory.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/425045)

### DIFF
--- a/manifests/u/ubranch/Jot/0.2.2/ubranch.Jot.installer.yaml
+++ b/manifests/u/ubranch/Jot/0.2.2/ubranch.Jot.installer.yaml
@@ -1,0 +1,16 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+PackageIdentifier: ubranch.Jot
+PackageVersion: 0.2.2
+MinimumOSVersion: 10.0.19041.0
+InstallerType: zip
+NestedInstallerType: portable
+NestedInstallerFiles:
+- RelativeFilePath: jot.exe
+  PortableCommandAlias: jot
+Installers:
+- Architecture: x64
+  InstallerUrl: https://github.com/ubranch/jot-gemini-transcribe-Windows/releases/download/v0.2.2/Jot-0.2.2-windows-x64.zip
+  InstallerSha256: F2CB373800F141DA3DBA3F0007F5D35E68C782892E029E2E95DA6FBE68E0A94C
+ReleaseDate: 2026-08-27
+ManifestType: installer
+ManifestVersion: 1.6.0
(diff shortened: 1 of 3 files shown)


ubranch:

ubranch:
Re: `Validation-Defender-Error` on `08. Installation Validation`.

I cannot reproduce a detection locally, so per step 3 of the [validation guide](https://github.com/microsoft/winget-pkgs/blob/master/doc/ValidationFailureGuide.md#validation-defender-error) I am commenting for investigation. Could a moderator share the detection name from the dynamic test? I will submit it as a false positive with that in hand.

I suspect the behaviour that triggers it, and would rather state it plainly than have a reviewer discover it:

**Jot installs a low-level keyboard hook and synthesises keystrokes.** That is the same behaviour profile as a keylogger, so a heuristic flag is entirely reasonable. It is intrinsic to what the app does — push-to-talk dictation needs a global key it can watch from any application, and it needs to type the resulting text into whatever window has focus. Concretely:

- `WH_KEYBOARD_LL` to detect the dictation key being held ([`hotkey_hook.rs`](https://github.com/ubranch/jot-gemini-transcribe-Windows/blob/main/crates/jot/src/hotkey_hook.rs))
- `SendInput` to type the transcript at the cursor ([`win32.rs`](https://github.com/ubranch/jot-gemini-transcribe-Windows/blob/main/crates/jot-core/src/win32.rs))

What it does **not** do, and this is checkable in the source above: the hook passes `vkCode` into a pure `classify` function and discards it. Nothing retains, logs, or transmits which key was pressed — the only value that leaves the hook is a five-variant enum (`HotkeyDown`, `HotkeyUp`, `EscDown`, `SpaceLock`, `OtherKeyDown`), and `OtherKeyDown` carries no key identity at all. The app also refuses to record while a password field has focus, checked through UI Automation's `IsPassword`.

The source is Apache-2.0 and the release is built by the CI workflow in the same repository, so the binary is reproducible from the tag.

I am aware the binary is unsigned; that is stated in the release notes rather than hidden. If the detection is signature-based rather than heuristic I will act on it immediately — the detection name is the missing piece.

PR #423778, opened 2026-08-25
New version: GRCSAC.NetVane version 1.2.1

GRCSAC:
### New version: `GRCSAC.NetVane` version 1.2.1

Maintenance release of NetVane (local network scanner for Windows). Same publisher, install behaviour and identity as the 1.2.0 submission (#423491, currently in moderation); only the version, installer URL, SHA256, release date and ARP `DisplayName`/`DisplayVersion` change.

- Installer URL (version-pinned, immutable): https://netvane.doorvane.com/downloads/NetVane-Setup-1.2.1.exe — live and verified resolving before this PR was opened
- Authenticode-signed, `CN=GRCSAC`, Microsoft Trusted Signing, timestamped
- SHA256 matches the vendor's Ed25519-signed `version.json`, which is what the app's own updater verifies against

- [x] Validated locally: `winget validate` → succeeded (winget v1.29.290)
- [x] The 1.2.0 URL remains hosted, so the earlier submission keeps resolving
- [x] Install path exercised via the in-app updater (download → SHA256 verify → silent Inno install → relaunch); post-install ARP entry matches `AppsAndFeaturesEntries` exactly (`NetVane 1.2.1` / `1.2.1` / `GRCSAC`)
- [ ] CLA — signed on #423491 (same account)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/423778)

PR #423491, opened 2026-08-24
New package: GRCSAC.NetVane version 1.2.0

GRCSAC:
### New package: `GRCSAC.NetVane` version 1.2.0

NetVane is a local network scanner and security monitor for Windows. It maps the devices on a network, fingerprints open ports, and matches discovered service versions against a bundled NVD and CISA-KEV database. It runs entirely on the local machine — no cloud service, no account, no agent.

- Homepage: https://netvane.doorvane.com
- Installer: Inno Setup, per-user (`PrivilegesRequired=lowest`), no UAC prompt
- Signed: Authenticode, `CN=GRCSAC`, via Microsoft Trusted Signing (valid, timestamped)

#### Checklist

- [ ] Contributor License Agreement — the repository owner will respond to the CLA bot.
- [x] I have checked there are no other open pull requests for this package. `manifests/g/GRCSAC/` does not exist upstream; this is a new publisher and a new package.
- [x] Validated locally: `winget validate --manifest ...` → **Manifest validation succeeded** (winget v1.29.290).
- [ ] `winget install --manifest` — see the note below. The install path was tested, but not through this command.
- [x] Conforms to the 1.6.0 schema.

#### On the install test

`winget install --manifest` requires `winget settings --enable LocalManifestFiles`, which needs administrator rights on the machine. Rather than change that setting, the install was verified through the path winget's `inno` handler actually takes:

```
NetVane-Setup-1.2.0.exe /VERYSILENT /SP- /SUPPRESSMSGBOXES /NORESTART
```

The installer URL and SHA256 in this manifest were verified independently against the published artifact — the live URL returns `200` with a `content-length` of 20,688,328 matching the file whose SHA256 is the one declared here — so the download-and-hash step is covered even though it was not exercised through `winget install` itself.

After the install, the ARP entry matched `AppsAndFeaturesEntries` exactly:

| field | manifest | registry (HKCU) |
|---|---|---|
| `DisplayName` | `NetVane 1.2.0` | `NetVane 1.2.0` |
| `DisplayVersion` | `1.2.0` | `1.2.0` |
| `Publisher` | `GRCSAC` | `GRCSAC` |
| `ProductCode` | `{7B2E9D4C-3A1F-4E58-9C2A-1D6B0F3E9A21}_is1` | same |

The application started and responded normally afterwards.

#### Notes on a couple of manifest choices

[...]

PR #425113, opened 2026-08-27
New package: GRCSAC.NetVane version 1.3.1

GRCSAC:
NetVane — a local-first network scanner and security monitor for Windows.

### Checklist
- [x] Manifest passes `winget validate`
- [x] Installer verified: installs and uninstalls cleanly, per-user, no admin/UAC
- [x] `InstallerSha256` matches the published installer
- [x] Installer is Authenticode-signed (`CN=GRCSAC`, Microsoft Trusted Signing, timestamped)
- [x] `InstallerUrl` is served over HTTPS from the vendor's own domain

### Notes for reviewers

**`AppsAndFeaturesEntries` carries `DisplayName` only, deliberately.** The Inno installer sets
`UninstallDisplayName` to `NetVane {version}`, so the Add/Remove Programs name is *not* the package
name and changes each release. Without the entry, winget cannot correlate an existing install and
offers an upgrade that is already applied. `DisplayVersion` is intentionally omitted — it would
equal `PackageVersion`, and @stephengillie correctly flagged that as redundant on an earlier
submission (#423491).

**Supersedes #423491 (1.2.0) and #423778 (1.2.1), both closed.** Those versions have since been
replaced by a security-remediation release, and we would rather they were not installable via
`--version`.

No `UpgradeCode` — that is MSI-only and this is an Inno Setup installer.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/425113)

### DIFF
--- a/manifests/g/GRCSAC/NetVane/1.3.1/GRCSAC.NetVane.installer.yaml
+++ b/manifests/g/GRCSAC/NetVane/1.3.1/GRCSAC.NetVane.installer.yaml
@@ -1,0 +1,33 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+
+PackageIdentifier: GRCSAC.NetVane
+PackageVersion: 1.3.1
+# Matches the published requirement ("Windows 10 or 11, 64-bit"). Deliberately not stricter: this
+# field makes winget REFUSE to install below it, so it must not promise less than the website does.
+MinimumOSVersion: 10.0.10240.0
+InstallerType: inno
+Scope: user
+InstallModes:
+  - interactive
+  - silent
+  - silentWithProgress
+UpgradeBehavior: install
+ReleaseDate: 2026-08-31
+Installers:
+  - Architecture: x64
+    InstallerUrl: https://netvane.doorvane.com/downloads/NetVane-Setup-1.3.1.exe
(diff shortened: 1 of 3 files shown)


GRCSAC:

GRCSAC:
Updated this PR from 1.3.0 to **1.3.1**, replacing the manifests rather than opening a second PR, so that the first NetVane version published here is the corrected one.

Two reasons for the change, both found after the original submission:

1. **A claim we had published about scan scope was not accurate**, and 1.3.1 retracts it. We would rather not have the version carrying it be the one that reaches users through winget.
2. **An installer defect.** A silent uninstall while the application was running removed the uninstaller's own ARP registration without removing the program, leaving NetVane installed, still runnable, and no longer uninstallable. Fixed in 1.3.1 and verified on a clean Windows 11 VM across interactive, silent, updater-relaunch and uninstall-while-running cases.

The branch is still named `netvane-1.3.0` — GitHub does not allow renaming a PR's head branch without closing it, and keeping this PR preserves its history. The contents are 1.3.1 only.

`InstallerSha256` is verified against the signed installer served at the `InstallerUrl`, and `winget validate` passes on all three manifests. Apologies for the churn, and thanks for your patience.

stephengillie (approved):

PR #424602, opened 2026-08-26
New package: XniperBuilds.Riplox version 1.4.0

xniperbuilds:
New package: **XniperBuilds.Riplox** 1.4.0 — a free, open-source (GPL-3.0) video downloader for Windows.

- Homepage: https://xniperbuilds.com/riplox-desktop/
- Source: https://github.com/xniperbuilds/riplox-desktop
- Release: https://github.com/xniperbuilds/riplox-desktop/releases/tag/v1.4.0

### Checklist

- [x] Have you validated your manifest locally with `winget validate --manifest <path>`? — `Manifest validation succeeded.`
- [x] Does your manifest conform to the 1.12 schema?
- [x] Have you checked that there aren't other open pull requests for the same manifest?

### Notes for the reviewer

- The installer is Inno Setup with `PrivilegesRequired=lowest`, so an unattended install lands in the user's profile and never prompts for elevation. `Scope: user` is set accordingly.
- `ProductCode` was read back from the registry of a machine with 1.4.0 actually installed, rather than derived, so upgrade detection matches what Setup writes.
- `InstallerSha256` is the checksum of the file at `InstallerUrl`; the same value is published as `SHA256SUMS.txt` on the release.
- The binary the manifest points at was installed and launched before this was opened. I have not run `winget install --manifest` on it, because this machine already carries a machine-scope install of the same version and a second, user-scope copy beside it would not have told me anything useful.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/424602)

PR #425117, opened 2026-08-27
New package: XniperBuilds.Riplox version 1.7.0

xniperbuilds:
New package: **XniperBuilds.Riplox** 1.4.1 — a free, open-source (GPL-3.0) video downloader for Windows.

- Homepage: https://xniperbuilds.com/riplox-desktop/
- Source: https://github.com/xniperbuilds/riplox-desktop
- Release: https://github.com/xniperbuilds/riplox-desktop/releases/tag/v1.4.1

### This replaces #424602 — please close that one in favour of this

That PR was for **1.4.0**, and its installer had a defect that mattered specifically to winget: on a machine without the Microsoft Edge WebView2 Runtime, Setup showed a message box and waited for an answer. `MsgBox` in Inno Setup is not suppressed by `/SUPPRESSMSGBOXES` — only `SuppressibleMsgBox` is — so an unattended install had nobody to answer it and **hung** rather than failing. It was caught by Chocolatey's verifier, which killed it after 45 minutes on Windows Server 2019.

Since winget installs silently, 1.4.0 would have hung for every user whose machine lacks the Runtime. **1.4.1 installs the Runtime instead of asking about it**, using Microsoft's own 1.7 MB Evergreen Bootstrapper, and never blocks: if the Runtime cannot be installed, Setup says so and continues.

I would rather withdraw a package that passed your checks than have it merged in that state.

### Checklist

- [x] Have you validated your manifest locally with `winget validate --manifest <path>`? — `Manifest validation succeeded.`
- [x] Does your manifest conform to the 1.12 schema?
- [x] Have you checked that there aren't other open pull requests for the same manifest? — #424602, which this replaces and which I am closing.

### Notes for the reviewer

- The installer is Inno Setup with `PrivilegesRequired=lowest`, so an unattended install lands in the user's profile and never prompts for elevation. `Scope: user` is set accordingly.
- `ProductCode` was read back from the registry of a machine with the app actually installed, rather than derived, so upgrade detection matches what Setup writes.
- `InstallerSha256` is the checksum of the file at `InstallerUrl`; the same value is published as `SHA256SUMS.txt` on the release.
- The binary the manifest points at was installed and launched before this was opened.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/425117)

### DIFF
--- a/manifests/x/XniperBuilds/Riplox/1.4.1/XniperBuilds.Riplox.installer.yaml
+++ b/manifests/x/XniperBuilds/Riplox/1.4.1/XniperBuilds.Riplox.installer.yaml
@@ -1,0 +1,22 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: XniperBuilds.Riplox
+PackageVersion: 1.4.1
+Platform:
+- Windows.Desktop
+# installer.iss sets MinVersion=10.0 - the bundled Python 3.11 does not run on
+# Windows 8.1 or older.
+MinimumOSVersion: 10.0.0.0
+InstallerType: inno
+# installer.iss sets PrivilegesRequired=lowest, so an unattended install lands
+# in the user's profile and never prompts for admin.
+Scope: user
+UpgradeBehavior: install
+ProductCode: '{9C2F41B7-6E4A-4C58-9B21-7D3E5A0F81C4}_is1'
+ReleaseDate: 2026-08-27
+Installers:
+- Architecture: x64
+  InstallerUrl: https://github.com/xniperbuilds/riplox-desktop/releases/download/v1.4.1/Riplox_Setup_v1.4.1.exe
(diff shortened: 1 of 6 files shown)


xniperbuilds:

xniperbuilds:
Updated this PR from 1.4.1 to 1.6.1.

1.4.1 was submitted on 27 August and is still awaiting moderation. Three releases have shipped since then, and 1.4.1 carries a defect that was fixed in 1.5.0 - so publishing it now would add a version to the community repository that is already known to be wrong.

Nothing else changed: same package identifier, same publisher, same installer type and switches. Only the version, the installer URL and its SHA-256 are different, and the installer is the one published at https://github.com/xniperbuilds/riplox-desktop/releases/tag/v1.6.1

Sorry for the extra validation run.

xniperbuilds:

PR #425122, opened 2026-08-27
New package: fpinero.mouse-desktop version 0.1.0

fpinero:
## 📖 Description

New package: `fpinero.mouse-desktop` version 0.1.0.

A tray utility that switches Windows virtual desktops with a mouse gesture: hold the wheel
button, drag left or right, and Windows moves to the neighbouring desktop. One 270 KB
executable with no runtime, no driver and no service, and it needs no Administrator
privileges at install time or at run time, which is the reason it exists: the vendor mouse
software that does the same job cannot be installed on a managed machine.

- Source: https://github.com/fpinero/mouse-desktop (Rust, MIT)
- Product page: https://mouse-desktop.nivelepsilon.com

Two notes on the manifest, in case they save a reviewer a question:

- `InstallerType: portable` rather than `zip` with a nested portable, because the release
  publishes a bare `.exe` and there is nothing to unpack.
- `Commands: [mouse-desktop]` so the alias is not derived from the file name. Without it the
  command would be `mouse-desktop-x64`, and every piece of the project's documentation,
  including its troubleshooting table, refers to `mouse-desktop`.

The installer URLs are versioned rather than `releases/latest/download/...`, so they keep
matching the hashes pinned here when a new release ships.

## ✅ Checklist

## 📦 Manifest Checklist

Tested on Windows 11 with winget 1.29.290: `winget validate` reports "Manifest validation
succeeded", `winget install --manifest` installs, the `mouse-desktop` alias starts it, it
appears in `winget list`, and it uninstalls cleanly.

### DIFF
--- a/manifests/f/fpinero/mouse-desktop/0.1.0/fpinero.mouse-desktop.installer.yaml
+++ b/manifests/f/fpinero/mouse-desktop/0.1.0/fpinero.mouse-desktop.installer.yaml
@@ -1,0 +1,31 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: fpinero.mouse-desktop
+PackageVersion: 0.1.0
+
+# The release publishes a bare .exe, not an archive, so this is `portable` directly rather
+# than `zip` with a nested portable inside it.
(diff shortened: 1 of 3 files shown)


fpinero:

PR #425236, opened 2026-08-27
New package: Mun1to.VoCript version 3.7.10

Mun1to:
### Pull request type

- [x] New package
- [ ] Package update
- [ ] Package removal
- [ ] Metadata or other change

### Package

`Mun1to.VoCript` version `3.7.10`

VoCript is a free and open source offline voice dictation tool for Windows: you press a global
shortcut, speak, and the text types itself into whatever application has the cursor. All the
speech recognition runs locally with Whisper and other on-device models, so no audio ever leaves
the machine and no account is needed. I am the author of the package.

### Checklist

- [x] Checked that there is no other open pull request for this manifest addition.
- [x] Validated locally with `winget validate --manifest <path>`, which returns `Manifest validation succeeded.`
- [x] Conforms to the 1.6 schema.
- [x] Verified the installer URL and its SHA256: the hash in the manifest was computed from the file downloaded from that exact release URL.
- [ ] **Not** tested with `winget install --manifest <path>`. This machine already has VoCript installed from the same NSIS installer and I did not want to reinstall over it. The installer is the one published on GitHub releases and in daily use.

### Installer notes

- NSIS installer, per user only (`installMode: currentUser` in Tauri), so it never prompts for
  administrator rights.
- Silent install with `/S`.
- The uninstall entry registers with `DisplayName: VoCript` and `Publisher: VoCript` under the
  current user's uninstall key, which is what the `AppsAndFeaturesEntries` block reflects.
- The package is also on the Microsoft Store as `9P247C96LC72`. This manifest points at the
  GitHub release instead, which is the build that carries the app's own auto updater.

### DIFF
--- a/manifests/m/Mun1to/VoCript/3.7.10/Mun1to.VoCript.installer.yaml
+++ b/manifests/m/Mun1to/VoCript/3.7.10/Mun1to.VoCript.installer.yaml
@@ -1,0 +1,20 @@
+PackageIdentifier: Mun1to.VoCript
+PackageVersion: 3.7.10
+InstallerType: nullsoft
+Scope: user
+InstallModes:
+  - interactive
+  - silent
+UpgradeBehavior: install
+ReleaseDate: 2026-08-23
+AppsAndFeaturesEntries:
+  - DisplayName: VoCript
+    Publisher: VoCript
+    ProductCode: VoCript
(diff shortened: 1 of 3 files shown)


Trenly:
[Policy] CLA Signed

PR #425278, opened 2026-08-27
New version: ASYS.Vantage version 153.1.0-1

Liangchenxu:
## ✅ Checklist

  - Resolves #[Issue Number]

## 📦 Manifest Checklist

- [ ] Checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/change
- [ ] Validated manifest locally with `winget validate --manifest <path>` ([validation guide](https://github.com/microsoft/winget-pkgs/blob/master/doc/ValidationFailureGuide.md))

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/425278)

### DIFF
--- a/manifests/a/ASYS/Vantage/153.1.0-1/ASYS.Vantage.installer.yaml
+++ b/manifests/a/ASYS/Vantage/153.1.0-1/ASYS.Vantage.installer.yaml
@@ -1,0 +1,26 @@
+# Created using wingetcreate 1.12.13.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: ASYS.Vantage
+PackageVersion: 153.1.0-1
+InstallerType: exe
+InstallerSwitches:
+  Silent: /S
+  SilentWithProgress: /S
+ElevationRequirement: elevatesSelf
+Installers:
+- Architecture: x64
+  InstallerUrl: https://dl.asystech.cn/latest/amd/vantage-153.1.0-1.x86_64-installer.exe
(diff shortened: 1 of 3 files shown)


Liangchenxu:
Hi @stephengillie , the silent install argument /S is correct and works fine locally. The validation seems to be timing out because the installer is getting blocked by SmartScreen/Defender in the sandbox. Could you please help trigger a manual bypass or take a look? Thank you!

Liangchenxu:
Hi @stephengillie,

Correction to my previous comment: after further investigation, SmartScreen/
Defender is NOT the cause — please disregard that speculation.

Actual cause: the installer requires VCRUNTIME140_1.dll (VC++ 2015+ runtime).
On clean machines without it, the OS loader shows a missing-DLL dialog that
cannot be dismissed under /S, so the process hangs. This matches the validation
log (arm64 track InProgress for ~3h before timeout), and we reproduced it in a
clean Windows Sandbox.

The runtime IS declared as a package dependency in the manifest
(Microsoft.VCRedist.2015+.x64/.x86/.arm64), and the winget client installs it
for end users — but the Installation Validation step runs the installer
directly without provisioning dependencies, hence the failure. With the
runtime installed, we verified the installer completes fully silent with /S.

We are preparing a self-contained build (static CRT / bundled runtime) for the
next release. Given end-user installs via winget are unaffected, could you
advise whether a manual review/override is acceptable here, or should we wait
for the repackaged build? Thank you.

stephengillie (approved):

PR #425633, opened 2026-08-28
New package: DominicLutz.BenchMeisterPCCheck version 1.0.15

Dominic-Lutz85:
## Description
Adds BenchMeister PC-Check 1.0.3, a portable Windows tool that reads out PC hardware via WMI and checks for wasted performance (XMP off, single-channel RAM, HDD). Open source (MIT), no installation, no network connection before explicit user consent.

## Checklist
- [ ] Signed the Contributor License Agreement

## Manifest Checklist
- [x] Validated manifest locally with winget validate --manifest <path>
- [ ] Tested manifest locally with winget install --manifest <path>
- [ ] Manifest conforms to the 1.12 schema (uses 1.6.0 schema)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/425633)


Dominic-Lutz85:

Dominic-Lutz85:
Updated this PR from 1.0.3 to 1.0.4.

Version 1.0.3 had a false positive in the tool's main check: on mainboards that report only the JEDEC base clock to Windows, it claimed the memory was running below its rated speed even when it was not. A user reported it and it was fixed the same day.

Shipping 1.0.3 through winget would knowingly distribute that defect, so I replaced the manifests instead of letting it merge. Installer URL points at the versioned release asset and the SHA256 is verified against the actual download.

Sorry for the churn, and thanks for the review time already spent.

Dominic-Lutz85:
Updated this PR from 1.0.4 to 1.0.7.

Reason: 1.0.4 reports the wrong RAM speed. The fix was in the code but never reached the user because of a forgotten conversion step between two packages, so 1.0.4 behaves exactly like the version before it. Shipping that through a package manager would be worse than a few days of delay.

1.0.7 fixes it, adds regression tests, and adds a disclosure to the consent screen about affiliate links on the result page. The installer URL points at the versioned release tag, the SHA256 matches the published asset, and the build is reproducible from the tag.

Sorry for the churn.

Dominic-Lutz85:
Updated this PR from 1.0.7 to 1.0.15.

This PR has been waiting for moderator review since 2026-08-28. In the meantime the tool moved eight releases ahead, mostly fixes from user feedback in a German hardware forum, including a crash fix in 1.0.14 (1.0.13 failed to start). Merging 1.0.7 would ship a version that is worse than the one on the project website.

Installer URL verified (HTTP 200, 9637888 bytes) and SHA256 recomputed from the downloaded file. The 1.0.7 folder has been removed, so this PR contains exactly one version.

PR #425676, opened 2026-08-28
New package: ColinGPT9.ClipsKitty version 1.1.4

ColinGPT9:
New package submission for **Clips Kitty**, a local-first AI video clipping app for Windows.

- Homepage: https://colingpt9.github.io/clips-studio/
- Source: https://github.com/ColinGPT9/clips-studio (AGPL-3.0-or-later)
- Release: https://github.com/ColinGPT9/clips-studio/releases/tag/v1.1.4

### Checklist

- [x] Manifests validate against schema 1.6.0
- [x] `InstallerSha256` verified against the asset as GitHub serves it
- [x] Installer URL is a permanent GitHub release asset
- [x] All URLs in the locale manifest return 200
- [x] This package is not already in the repository

### One thing worth flagging for validation

**The installer is a web installer, and the payload is large.**
`ClipsKitty-Web-Setup-1.1.4.exe` is 795 KB and downloads roughly **5.8 GB** from
Hugging Face when it runs. The size is inherent to what the app is: it bundles a
CUDA build of PyTorch, an Ollama runtime and two Whisper models, because all the
AI runs on the user's own machine rather than on a server.

winget itself handles this fine — it just runs the installer — but a download
that size inside the validation VM's time limit may well time out. If the
automated check fails on the download rather than on the manifests, that is why,
and I am happy to work through whatever you need.

The installer supports silent and interactive modes and installs per-user, as
declared in the installer manifest.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/425676)

### DIFF
--- a/manifests/c/ColinGPT9/ClipsKitty/1.1.4/ColinGPT9.ClipsKitty.installer.yaml
+++ b/manifests/c/ColinGPT9/ClipsKitty/1.1.4/ColinGPT9.ClipsKitty.installer.yaml
@@ -1,0 +1,17 @@
+PackageIdentifier: ColinGPT9.ClipsKitty
+PackageVersion: 1.1.4
+MinimumOSVersion: 10.0.17763.0
+InstallerType: nullsoft
+Scope: user
+InstallModes:
+  - interactive
+  - silent
+  - silentWithProgress
+UpgradeBehavior: install
+ReleaseDate: 2026-08-28
+Installers:
+  - Architecture: x64
+    InstallerUrl: https://github.com/ColinGPT9/clips-studio/releases/download/v1.1.4/ClipsKitty-Web-Setup-1.1.4.exe
+    InstallerSha256: C6CB1F67B8EB8ACA05B35604F3724622E8EB52CA412F64153922AC274746D109
+ManifestType: installer
+ManifestVersion: 1.6.0
--- a/manifests/c/ColinGPT9/ClipsKitty/1.1.4/ColinGPT9.ClipsKitty.installer.yaml
+++ b/manifests/c/ColinGPT9/ClipsKitty/1.1.4/ColinGPT9.ClipsKitty.installer.yaml
@@ -7,6 +7,12 @@ InstallModes:
   - interactive
   - silent
   - silentWithProgress
+# Spelled out rather than left to winget's inference for `nullsoft`. Verified
+# by running the real installer with /S: no window, installs silently. Every
+# comparable package already in winget (Figma, Notion) declares these.
+InstallerSwitches:
+  Silent: /S
+  SilentWithProgress: /S
 UpgradeBehavior: install
 ReleaseDate: 2026-08-28
 Installers:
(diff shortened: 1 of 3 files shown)


ColinGPT9:
Thanks — following up on the `08. Installation Validation` failure and the `Validation-Shell-Execute` label.

I went through the [Validation Failure Guide](https://github.com/microsoft/winget-pkgs/blob/master/doc/ValidationFailureGuide.md) and I do not think either of the listed causes applies here, so I would rather ask than push a speculative change:

**Not an elevation problem.** The installer is built per-user (`perMachine: false`), writes to `%LOCALAPPDATA%`, touches no HKLM keys and installs no service. `Scope: user` in the manifest reflects that. Adding `ElevationRequirement: elevationRequired` would make the manifest claim something untrue about the installer, so I have not done it.

**Not a corrupt binary.** `07. Installers Scan` passed, and I verified the `InstallerSha256` against the asset as GitHub actually serves it rather than against my local build.

**What I think it is.** `ClipsKitty-Web-Setup-1.1.4.exe` is 795 KB and downloads roughly **5.8 GB** from Hugging Face before it can finish installing. "Did not complete" fits that exactly, and the check failed at ~5 minutes.

The size is inherent to the product rather than carelessness: it is a local-first AI video tool, so it bundles a CUDA build of PyTorch, an Ollama runtime and two Whisper models. Nothing is sent to a server, which is the entire point of it, and that means the models have to arrive on the machine.

**So a question rather than a fix.** How would you prefer a package like this to be submitted?

- Is a waiver or an extended timeout appropriate for large web installers?
- Is there a manifest field I have missed that signals this to validation?
- Would you rather have the portable `.zip` (~6.4 GB direct download, `InstallerType: zip`) instead? I have not proposed it because it is a worse experience — no shortcuts, no updater — but I would rather match how you want this handled than argue for my preference.

Happy to make whatever change you point at. I have deliberately not pushed anything since the failure, to avoid burning a validation cycle on a guess.

ColinGPT9:
Update, so the next person to look at this does not have to repeat it.

I declared the silent switches explicitly rather than relying on inference for `nullsoft`:

```yaml
InstallerSwitches:
  Silent: /S
  SilentWithProgress: /S
```

Verified beforehand by running the shipped installer with `/S` on a clean Windows box — no window appears, it proceeds silently. So the declaration is accurate, not a guess at what validation wanted.

**`08. Installation Validation` failed identically**, and `Validation-Shell-Execute` came back. So the switches are eliminated as a cause.

Where that leaves it, with everything ruled out by evidence rather than assumption:

| Possible cause | Status |
|---|---|
| Missing `ElevationRequirement` | Ruled out — built `perMachine: false`, per-user, writes only to `%LOCALAPPDATA%`. `Scope: user` is already correct, and declaring `elevationRequired` would be untrue. |
| Corrupt / invalid binary | Ruled out — `07. Installers Scan` passes, and the SHA256 matches the asset as GitHub serves it. |
| Silent switches missing or wrong | Ruled out — declared explicitly, tested, failure unchanged. |
| The installer downloads ~5.8 GB before it can finish | **The only one left.** |

I have not pushed anything further, because at this point any change would be guessing rather than fixing, and each push costs a validation cycle.

If the download size is the problem, I would rather be told how you want it handled than keep iterating: a waiver, a longer timeout, or the portable `.zip` submitted instead (~6.4 GB direct, `InstallerType: zip`, no downloader). I am happy with whichever you prefer — I just cannot tell from outside which one you want.

Happy to provide logs, a test build, or anything else useful.

ColinGPT9:

PR #250726, opened 2025-04-24
fix: Flowscripter.template-bun-executable add ElevationRequirement: elevationProhibited

vectronic:
Workaround issue https://github.com/oven-sh/bun/issues/18193 as discussed here https://github.com/microsoft/winget-pkgs/pull/249737#issuecomment-2825938419
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/250726)

PR #426153, opened 2026-08-29
fix: Flowscripter.template-bun-executable remove ElevationRequirement: elevationProhibited

vectronic:
The ElevationRequirement: elevationProhibited was added in #250726 as a workaround for oven-sh/bun#18193, where a Bun-compiled Windows executable launched via a symlink (as WinGet portable installs create) would run the Bun runtime itself instead of the bundled application.

That Bun issue has since been fixed upstream by oven-sh/bun#21091 (merged 2025-07-16, closed 2026-07-25): compiled executables now locate their embedded module data via the in-memory PE image rather than reopening the file through the reparse point, so launching via a symlink now works correctly. This package's release pipeline already builds with the latest Bun, which includes the fix.

Removing the elevation restriction restores standard portable-install behavior (symlinked into PATH) for this package.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/426153)

### DIFF
--- a/manifests/f/Flowscripter/template-bun-executable/1.3.22/Flowscripter.template-bun-executable.installer.yaml
+++ b/manifests/f/Flowscripter/template-bun-executable/1.3.22/Flowscripter.template-bun-executable.installer.yaml
@@ -14,7 +14,6 @@ Dependencies:
   PackageDependencies:
   - PackageIdentifier: Microsoft.VCRedist.2015+.x64
 ReleaseDate: 2026-07-03
-ElevationRequirement: elevationProhibited
 Installers:
 - Architecture: x64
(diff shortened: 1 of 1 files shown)


vectronic:
Closing this canary. The validation failure is very likely explained by an unrelated bun limitation, not the elevation change: this build (template-bun-executable v1.3.22, released 2026-07-03) embeds bun 1.3.14, and bun's `bun:ffi`/dlopen was completely unavailable on Windows arm64 until oven-sh/bun#33696 (merged 2026-07-20, first in a later 1.4.x release). If winget-pkgs' installation validation runs the installed exe to confirm its version, that would crash on the arm64 leg for reasons that have nothing to do with ElevationRequirement. We'll cut a fresh release built with a current bun (>=1.4.2, which has the fix) and reopen this correction against that version once it's out.

PR #418623, opened 2026-08-17
New package: TechyGeeksHome.UltimateSettingsPanel version 8.0.0

techygeekshome:
## Description

New package: **Ultimate Settings Panel 8.0.0** by TechyGeeksHome — 250+ Windows settings, tools, diagnostics and commands in one searchable panel.

Distributed as a zip containing a single portable executable, so this uses `InstallerType: zip` with `NestedInstallerType: portable`. Nothing is installed and no elevation is required.

- Product page: https://techygeekshome.info/ultimate-settings-panel-online/
- Source: https://github.com/techygeekshome/Ultimate-Settings-Panel
- Release: https://github.com/techygeekshome/Ultimate-Settings-Panel/releases/tag/v8.0.0

## Checklist

- [x] Signed the [Contributor License Agreement](https://cla.opensource.microsoft.com)
- [ ] Linked to an issue (if applicable)

## Manifest Checklist

- [x] Checked that there aren’t other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/change
- [x] This PR only modifies one (1) manifest
- [x] Validated manifest locally with `winget validate --manifest <path>`
- [x] Tested manifest locally with `winget install --manifest <path>`
- [x] Manifest conforms to the [1.12 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.12.0)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/418623)

PR #426207, opened 2026-08-29
New version: TechyGeeksHome.UltimateSettingsPanel version 8.0.4

techygeekshome:
## 📖 Description

TechyGeeksHome.UltimateSettingsPanel 8.0.4.

This supersedes #418623, which was opened for 8.0.0. I am closing that one so there is a single current manifest for this package.

Same packaging as 8.0.0 - a zip carrying a portable executable, installed via `NestedInstallerType: portable` with the `usp` command alias. Only the version, the installer URL, the SHA-256, the release date and the release notes link have moved. `LicenseUrl` has been added.

## ✅ Checklist

## 📦 Manifest Checklist

The last two are unticked deliberately rather than overlooked - this manifest was prepared without a Windows host to run `winget validate` on. What was done instead: the zip was downloaded from the release and its SHA-256 checked against the release's own `SHA256SUMS.txt`, its contents listed to confirm `Ultimate Settings Panel.exe` sits at the archive root exactly as `RelativeFilePath` states, and the three files were parsed and checked against the 1.12.0 manifest shape. Happy to correct anything the validation pipeline flags.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/426207)

### DIFF
--- a/manifests/t/TechyGeeksHome/UltimateSettingsPanel/8.0.4/TechyGeeksHome.UltimateSettingsPanel.installer.yaml
+++ b/manifests/t/TechyGeeksHome/UltimateSettingsPanel/8.0.4/TechyGeeksHome.UltimateSettingsPanel.installer.yaml
@@ -1,0 +1,19 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+PackageIdentifier: TechyGeeksHome.UltimateSettingsPanel
+PackageVersion: 8.0.4
+Platform:
+  - Windows.Desktop
+MinimumOSVersion: 10.0.0.0
+InstallerType: zip
+NestedInstallerType: portable
+NestedInstallerFiles:
+  - RelativeFilePath: Ultimate Settings Panel.exe
+    PortableCommandAlias: usp
+UpgradeBehavior: install
+ReleaseDate: 2026-08-29
+Installers:
+  - Architecture: x64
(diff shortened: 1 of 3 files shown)


techygeekshome:
Could a moderator say which part of policy 1.2 this was flagged under? Happy to fix it if it is something in the package, but I cannot tell from the label what to change.

For context, in case it helps whoever picks this up:

- The same `Policy-Test-1.2` label was applied to #418623 (version 8.0.0) on 17 August, so this is not new to 8.0.4 and is not caused by the manifest change.
- Ultimate Settings Panel is a launcher: it opens Windows settings pages, Control Panel applets and administrative tools that are already on the machine. It does not download or execute anything remote, and it installs nothing. That is probably the behaviour a heuristic is reacting to.
- Full source is at https://github.com/techygeekshome/Ultimate-Settings-Panel (GPL-3.0 project, freeware licence on this component), and the zip in the manifest is the published release artefact, SHA-256 `20F59D823894710D8E59F6927BD0B2B5EA955B8ECF52F9C532D7523709FA2F65`, matching the `SHA256SUMS.txt` on the release.
- The binaries are not currently code signed. If that is the blocker, say so and I will stop resubmitting until it is sorted.

Happy to make any change that would help, or to withdraw the package if it is not a fit.

techygeekshome:

PR #422936, opened 2026-08-23
New package: AurelioAvila.Redaxa version 0.3.0

AurelioAvila:
Redaxa (formerly PromptShield) — privacy-first AI gateway that scans and redacts sensitive data before it reaches ChatGPT, Claude & other AI tools.

Renamed from PromptShield after finding it collided with several unrelated existing products, including one with an established domain and user base. This is a fresh package identifier (AurelioAvila.Redaxa); the previous AurelioAvila.PromptShield submission (#417245) was closed unmerged.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/422936)

PR #426231, opened 2026-08-29
New package: AurelioAvila.Redaxa version 0.3.3

AurelioAvila:
Adds AurelioAvila.Redaxa version 0.3.0. This branch previously had a 0.2.0 submission (PR #422936, auto-closed after a manifest update); the 0.2.0 files have been removed and replaced with 0.3.0, which is now the only version present in this branch.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/426231)


AurelioAvila:
Corrected the InstallerSha256: the 0.3.0 binary was rebuilt and re-uploaded to the same v0.3.0 tag after this PR was opened, so the original hash no longer matched the file at InstallerUrl. Re-downloaded and re-hashed the currently published asset; the manifest now matches it.

AurelioAvila:
Updated to 0.3.1 (security-only release: rotates the update-signing key away from the one shared with two other packages I maintain). Removed the 0.3.0 manifest in favor of it, same as the earlier 0.3.0-for-0.2.0 swap on this same PR.

AurelioAvila:
Updated to 0.3.2 and removed the 0.3.1 manifest in favour of it, same as the earlier swaps on this PR.

0.3.2 completes the update-signing key rotation that 0.3.1's notes announced but did not actually ship: that build still carried the old shared public key, so signing with the new one would have been rejected by every install in the field. This release ships the Redaxa-only key and is signed with the shared one for the last time.

InstallerUrl and InstallerSha256 were taken by downloading the published asset and hashing it, not from the local build — that is what went wrong on this PR once before.

Thanks for your patience with the version churn here.

Trenly:
cc @denelon @stephengillie

### Policy-Test-1.8 Analysis

2. **Flagged Substring:** banking (heuristic classification alongside described subscriptions, Stripe checkout, and payment-plan functionality)
3. **Recommendation:** review for waiver — Redaxa advertises paid subscriptions and a Stripe-hosted checkout, which is financial transaction and subscription functionality covered by policy 1.8 and warrants review.

---
*Authored by Copilot*

AurelioAvila:
Thanks for flagging this for review.

Redaxa's subscription checkout uses Stripe Checkout in hosted mode
(stripe.checkout.sessions.create, mode: "subscription"). The payment page
is served and branded by Stripe itself — the app never collects or stores
card data locally. This satisfies the secure purchase API requirements
under 1.8.2: Stripe identifies itself as the commerce provider, authenticates
the buyer, and requires explicit confirmation before any charge, and
payment processing meets PCI DSS as Stripe's platform.

Pricing and plan details are described in-product before checkout starts,
and the description states clearly that a purchase is a subscription
(Stripe-hosted, recurring) rather than a one-time transaction.

Happy to provide anything else needed for review.

PR #426586, opened 2026-08-31
New package: TransitAISoftware.Transit version 6.5.1

AiKnox:
### Pull request has been manually validated?

- [x] Yes

### Have you checked that there aren't other open pull requests for the same manifest?

- [x] Yes

### Have you validated your manifest locally with `winget validate --manifest <path>`?

- [x] Yes — `Manifest validation succeeded.`

### Have you tested your manifest locally with `winget install --manifest <path>`?

- [x] Yes — installed on Windows: `Successfully verified installer hash` / `Successfully installed`

### Does your manifest conform to the [1.12 schema](https://github.com/microsoft/winget-cli/tree/master/schemas/JSON/manifests/v1.12.0)?

- [x] Yes

---

**Transit** is a cross-platform SSH client for network engineers with an embedded AI agent restricted to read-only investigation.

- Publisher: **Transit AI Software Inc.**
- Homepage: https://transitai.app/

### Verification performed

- `winget validate --manifest` → `Manifest validation succeeded.`
- `winget install --manifest` on Windows → installer downloaded, hash verified by winget, installed successfully.
- All three manifests also validate against the published `1.12.0` JSON schemas.
- `InstallerSha256` was computed from the installer downloaded from the URL in the manifest, rather than copied from a build log.
- The installer is Authenticode-signed; the leaf certificate subject is `O=Transit AI Software Inc., C=US, ST=Louisiana, L=Baton Rouge`, issued by SSL.com Code Signing Intermediate CA RSA R1 — matching the declared publisher.
- Every URL in the locale manifest resolves (`PublisherUrl`, `PublisherSupportUrl`, `PrivacyUrl`, `PackageUrl`, `LicenseUrl`).

### Notes for reviewers

**Only the user-scope NSIS installer is listed, though an MSI is also published.** The application ships a signed in-app updater that swaps the NSIS installer specifically. Listing the per-machine MSI as well would let a user install per-machine and then have the updater place a second, user-scope copy — leaving two installations and winget tracking the wrong one. Offering only the artifact the updater maintains keeps the installed state coherent.

`Protocols: [transit]` is declared because the application registers a `transit://` URL scheme, which its browser-based sign-in depends on.

### DIFF
--- a/manifests/t/TransitAISoftware/Transit/6.5.1/TransitAISoftware.Transit.installer.yaml
+++ b/manifests/t/TransitAISoftware/Transit/6.5.1/TransitAISoftware.Transit.installer.yaml
@@ -1,0 +1,16 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: TransitAISoftware.Transit
+PackageVersion: 6.5.1
+InstallerType: nullsoft
+Scope: user
+UpgradeBehavior: install
+Protocols:
+- transit
+ReleaseDate: 2026-08-31
+Installers:
+- Architecture: x64
+  InstallerUrl: https://downloads.transitai.app/v6.5.1/Transit_6.5.1_x64-setup.exe
+  InstallerSha256: D5C9E3DC7374B39088320C30527B9BAECC73131138F62C63763A3AFAA9F61865
+ManifestType: installer
+ManifestVersion: 1.12.0
--- a/manifests/t/TransitAISoftware/Transit/6.5.1/TransitAISoftware.Transit.installer.yaml
+++ b/manifests/t/TransitAISoftware/Transit/6.5.1/TransitAISoftware.Transit.installer.yaml
@@ -12,5 +12,6 @@ Installers:
 - Architecture: x64
   InstallerUrl: https://downloads.transitai.app/v6.5.1/Transit_6.5.1_x64-setup.exe
   InstallerSha256: D5C9E3DC7374B39088320C30527B9BAECC73131138F62C63763A3AFAA9F61865
+  ProductCode: Transit
 ManifestType: installer
 ManifestVersion: 1.12.0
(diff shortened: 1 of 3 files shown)


AiKnox:
@microsoft-github-policy-service agree company="Transit AI Software Inc."

AiKnox:
Added `ProductCode: Transit` to the installer entry.

Testing `winget install --manifest` on Windows surfaced that the installed package was not being correlated back to this manifest — `winget list` reported it as `ARP\User\X64\Transit` rather than `TransitAISoftware.Transit`, which would have left `winget upgrade` unable to offer updates. The ARP identifier gives the NSIS product code directly, so declaring it fixes the correlation.

Re-validated against the 1.12.0 installer schema after the change.

AiKnox:
Correcting my previous comment: I described the `ARP\User\X64\Transit` identifier from `winget list` as a correlation failure that `ProductCode` fixes. That was wrong. An installed package falls back to the ARP identifier when it cannot be matched against a package in a **configured source** — and this package is not in one yet, so there was nothing to correlate against. The output was expected for a local-manifest install, not a defect.

`ProductCode: Transit` is retained deliberately, but for a different reason than I gave: it matches the NSIS ARP key exactly, so correlation is deterministic rather than dependent on DisplayName/publisher heuristics. It is not a fix for a problem this PR had.

Everything else in the verification stands unchanged — `winget validate` succeeds, `winget install --manifest` installs with the hash verified by winget, and the installer's Authenticode leaf subject matches the declared publisher.

AiKnox:
The `Policy-Test-2.7` (Adult Content) label looks like a keyword false positive, and I have removed the likely trigger.

The description contained the phrase "requires an **explicit** click", describing the approval step a user takes before the application runs a command on a network device. That is the only word in any of the three manifests that a keyword scan for adult content would plausibly match — there is no other candidate in the metadata.

It carried no weight in the sentence, so I have reworded it to "and then you approve it by hand before it runs", which says the same thing more plainly. The locale manifest re-validates against the 1.12.0 schema; nothing else changed.

Flagging it here in case that saves a reviewer the time of working out what tripped it. Happy to adjust further if something else in the metadata is the actual cause.

PR #427477, opened 2026-09-01
New package: FKeysCreative.KeyJ version 1.6.1

vince-gonzalez:
Replaces #421552 (opened under a previous account, version 1.5.0). Filed under the account that owns the repository the manifest points at, at the current release.

**Version 1.6.1.** `InstallerSha256` is `ECAD5EBB0B1A9F7937ABD50F7432599E9D7B0D3DC261D28015498319C2B36D49`, the digest GitHub reports for `Key-J-1.6.1-win-x64.exe` on tag `keyj-v1.6.1`. `winget validate` passes.

**Addressing the `Validation-Shell-Execute` this package hit at 1.5.0 and 1.6.0.** Key-J is a per-user install (electron-builder NSIS, installs to LocalAppData; `Scope: user`). The default silent switch `/S` let the assisted installer make its own install-mode decision, which could attempt a per-machine install and so require elevation the validation environment cannot grant. This manifest sets

```
InstallerSwitches:
  Silent: /S /currentuser
  SilentWithProgress: /S /currentuser
```

`/currentuser` is electron-builder's switch to force the per-user install mode, consistent with `Scope: user`, so the installer completes without elevation.

**Licence.** Key-J up to 1.5.0 was MIT; from 1.6.0 it is proprietary, copyright F-Keys Creative LLC. `LicenseUrl` points at the licence that governs it. The source stays readable — the app installs a global keyboard hook and should be auditable by the people running it — but it is not redistributable.

**1.6.1 changes** the global-capture keyboard hook so it is installed only while capture is switched on, rather than at launch for the whole session.

#421552 is closed in favour of this one.


vince-gonzalez:

PR #428273, opened 2026-09-02
Murl 0.5.0

thatwasyahya:
### New package: `thatwasyahya.murl` version 0.5.0

mURL resolves one `murl://` identifier to a set of resources — web pages, local files and directories, terminals — and dispatches each to its handler under a consent-based security policy. This adds the CLI.

- Homepage: https://github.com/thatwasyahya/mURL
- Release: https://github.com/thatwasyahya/mURL/releases/tag/v0.5.0
- Docs: https://thatwasyahya.github.io/mURL/

### Manifest checklist

- [x] No other open pull request for this manifest
- [x] This PR only adds one manifest (three files, one package version)
- [x] Manifest conforms to the **1.12.0** schema
- [ ] Validated with `winget validate` — see note below
- [ ] Tested with `winget install --manifest` — see note below

**On the two unchecked boxes, stated plainly rather than ticked:** these manifests were authored on Linux, so `winget validate` and `winget install` could not be run. Instead every file was validated programmatically against Microsoft's published v1.12.0 JSON schemas — version, installer and defaultLocale — plus the cross-file rules: one `PackageIdentifier`, one `PackageVersion`, and exactly one of each manifest type. The script that does it is committed in the source repository at `packaging/winget/validate.py` and exits non-zero on any error, so it gates rather than informs. It caught two real defects before submission: `ReleaseDate` parsed by YAML as a date object where the schema requires a string, and a date that was not the release date.

Additionally verified against the artifact itself, not just the manifest:

- `InstallerSha256` equals the SHA-256 of the published archive
- the `NestedInstallerFiles` relative path exists inside that zip
- `InstallerUrl` is a permanent GitHub release asset served over HTTPS

If a maintainer would prefer this held until it can be run through `winget install` on a Windows machine, please say so and I will do that before asking for a merge.

### Notes for reviewers

The package is **portable** — a zip containing `murl.exe`. It installs no service, writes nothing outside the winget package directory, and does not register a URI scheme on install. Making `murl://` links clickable is a separate, explicit and reversible `murl os install`, which writes only under `HKCU\Software\Classes`.

x64 only: the project publishes no Arm64 Windows archive, so no Arm64 installer entry is listed rather than one pointing at a URL that would 404.

The project is experimental and its own description says so; this is a tagged release with checksummed artifacts built in CI.

### DIFF
--- a/manifests/t/thatwasyahya/murl/0.5.0/thatwasyahya.murl.installer.yaml
+++ b/manifests/t/thatwasyahya/murl/0.5.0/thatwasyahya.murl.installer.yaml
@@ -1,39 +1,16 @@
-# Validated against the winget v1.6.0 JSON schemas (packaging/winget/validate.py)
-# and submitted; see packaging/README.md for the submission status.
-#
-# winget installer manifest. Nothing here has been submitted to
-# microsoft/winget-pkgs; treat every value as a proposal.
-#
-# BEFORE SUBMITTING a human must replace InstallerSha256 below. The release
-# workflow uploads murl-<tag>-x86_64-pc-windows-msvc.zip.sha256 next to the
-# archive; it holds the bare uppercase digest with no filename, which is the
-# form winget wants. The zeros are a deliberate placeholder — they make an
-# install attempt fail on the hash rather than fetch something unverified.
-#
-# The archive is a plain zip of a directory containing murl.exe, so this is a
-# "portable" package: winget unpacks it and creates a `murl` alias on PATH.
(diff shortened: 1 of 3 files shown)


thatwasyahya:

thatwasyahya:
The two boxes I left unchecked are now covered by the pipeline: 08. Installation Validation passed (31m31s). The earlier failure was mine, NestedInstallerFiles at the manifest root instead of inside the Installers entry — and is fixed.

PR #428281, opened 2026-09-02
New package: Syntora.SeaShell version 1.29.0

peta10:
### Pull request has been manually verified
- [x] Does your manifest conform to the [1.12 schema](https://github.com/microsoft/winget-cli/tree/master/schemas/JSON/manifests/v1.12.0)?

### Package details

SeaShell is a terminal agent platform. This is the first submission for `Syntora.SeaShell`.

The installer is a `.zip` containing a single portable executable, `seashell.exe`, so the manifest uses `InstallerType: zip` with `NestedInstallerType: portable`.

Both installer URLs point at the publisher's own public release repository, `github.com/peta10/seashell-releases`, and the `InstallerSha256` values are taken from the published release artifacts.

The manifests are generated from the release by the publisher's release workflow rather than hand-written, so future version submissions will carry the same shape.

Homepage: https://seashell.sh

### What was and was not verified locally

Being explicit rather than ticking the box, since the second item could not be done here.

**Validated.** `winget validate --manifest` on winget **v1.29.290** returns `Manifest validation succeeded.`

**Not installed via `--manifest`.** That requires `winget settings --enable LocalManifestFiles`, which needs an administrator shell that was not available in the environment these manifests were prepared in. Rather than tick a box for something untested, it is left unchecked.

**What was verified instead**, on Windows x64, against the same artifact URL and the same sha256 this manifest pins:

```
scoop install seashell
  Downloading https://github.com/peta10/seashell-releases/releases/download/v1.29.0/seashell-windows-x64.zip
  Checking hash of seashell-windows-x64.zip ... ok.
  Extracting seashell-windows-x64.zip ... done.

seashell --version
  1.29.0
```

So the archive downloads from the pinned URL, matches the pinned checksum, extracts to a single `seashell.exe` at the archive root, and the extracted executable runs and reports the version this manifest declares.

Happy to complete the `--manifest` install and report back if that is preferred before review.


peta10:
@microsoft-github-policy-service agree company="Syntora"

PR #428398, opened 2026-09-02
New package: RicardoFrantz.pdf-next version 0.9.0

ricardofrantz:
### Validation done locally

- [x] `winget validate --manifest <path>` — succeeded
- [x] Installer verified by hand: `pdf-next_0.9.0_x64-setup.exe /S` exits 0 and
  registers `HKCU\...\Uninstall\pdf-next` (DisplayName `pdf-next`, version
  `0.9.0`, publisher `frantz`), which is what `AppsAndFeaturesEntries` records.
  `winget install --manifest` itself needs `LocalManifestFiles` enabled by an
  administrator, which I could not do on this machine.

### Manifest

- **Package**: `RicardoFrantz.pdf-next` 0.9.0 (new package)
- **Installer**: NSIS, x64, user scope, silent install supported
- **Source**: https://github.com/ricardofrantz/pdf-next/releases/tag/v0.9.0

pdf-next is a small MIT-licensed PDF, image and markdown viewer that reloads
the document within a second of the file changing on disk — built for the
LaTeX and Typst compile loop. It is built and published by GitHub Actions
from the public repository above.

### Changes since the first push

- Manifest schema 1.9.0 -> 1.12.0, as the automated review suggested.
- Added `AppsAndFeaturesEntries`. The installer writes `frantz` as the
  publisher (it comes from the bundle identifier) while this manifest says
  `Ricardo Frantz`, so without the entry `winget upgrade` had nothing to
  correlate on. The application has been corrected upstream; the entry here
  describes the 0.9.0 installer as it actually is.

### Note

The installer is not code signed yet, so Microsoft Defender SmartScreen may
warn on first run. A signing certificate is being arranged, and the manifest
will be updated when signed builds ship.

https://claude.ai/code/session_019yqEUNbvwwgHbkNYMoaBAk

### DIFF
--- a/manifests/r/RicardoFrantz/pdf-next/0.9.0/RicardoFrantz.pdf-next.installer.yaml
+++ b/manifests/r/RicardoFrantz/pdf-next/0.9.0/RicardoFrantz.pdf-next.installer.yaml
@@ -1,4 +1,4 @@
-# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.9.0.schema.json
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
 PackageIdentifier: RicardoFrantz.pdf-next
 PackageVersion: 0.9.0
 InstallerType: nullsoft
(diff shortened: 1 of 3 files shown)


ricardofrantz:

PR #428445, opened 2026-09-02
New version: AurelioAvila.SocialDashboard version 1.8.0

AurelioAvila:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/428445)

### DIFF
--- a/manifests/a/AurelioAvila/SocialDashboard/1.8.0/AurelioAvila.SocialDashboard.installer.yaml
+++ b/manifests/a/AurelioAvila/SocialDashboard/1.8.0/AurelioAvila.SocialDashboard.installer.yaml
@@ -1,0 +1,20 @@
+# Created with WinGet Releaser using komac v2.16.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: AurelioAvila.SocialDashboard
+PackageVersion: 1.8.0
+Platform:
+- Windows.Desktop
+MinimumOSVersion: 10.0.17763.0
+InstallerType: zip
+NestedInstallerType: portable
+NestedInstallerFiles:
+- RelativeFilePath: Social Dashboard.exe
+  PortableCommandAlias: socialdashboard
+ReleaseDate: 2026-09-02
+Installers:
+- Architecture: x64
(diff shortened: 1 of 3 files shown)


AurelioAvila:
Thanks — following up on the Defender validation error.

I downloaded the exact asset the manifest points at and checked it here before replying:

- SHA256 of the downloaded file matches both the manifest and the .sha256 published beside the release: F7E568E9C468D8D9B4009770A781615078C6AB0E120E6BD9ADB6EAD5AC5D1E21. So the URL and hash are correct and the file is not truncated or substituted.
- Scanned with Microsoft Defender, platform 4.18.26080.3, current signatures: no threats, on the .zip and on both extracted executables (the app and its updater).

The package is a portable .zip holding a PyInstaller-built, not-yet-code-signed executable, which is a common source of generic heuristic hits. Version 1.7.5 of this same package, built by the same pipeline from the same spec, passed validation and merged two days ago; nothing changed between them in how the binary is produced. UPX compression is already disabled in the build specifically to reduce that class of false positive.

Could the installer scan be re-run? If it still reproduces, I would appreciate the detection name so I can submit the file to the Defender team with something specific — the submission form asks for it, and I would rather not send a report that just says "no detection locally".

Happy to provide anything else useful.

AurelioAvila:
One more independent data point, since the detection has not been reproduced anywhere outside this pipeline.

VirusTotal, URL analysis of the exact release asset the manifest points at, run today (2026-09-03): **0 / 91 security vendors** flag it. That includes BitDefender, ESET, Emsisoft, Dr.Web, Acronis and the rest of the engine set. HTTP status 200, content type application/octet-stream, resolving to release-assets.githubusercontent.com as expected.

https://www.virustotal.com/gui/url/f9ac18ab2ded7471dcc11627d6e1ef2daa3a2e9e73b6ddf8a6178d24d8cf2224

Together with the local Microsoft Defender scan (platform 4.18.26080.3, clean on the zip and on both extracted executables) and the fact that 1.7.5 passed this same validation from an identical build, I have not been able to reproduce the block anywhere.

I am also filing the binary with the Defender team through the WDSI developer submission form. Worth noting for the record: that form requires a detection name as a mandatory field, and the Installers Scan message does not include one, so the submission carries "not disclosed" rather than a guess. If the re-run does produce a detection name, I will add it here and update the submission.

Thanks again for the time.

AurelioAvila:
Superseded by #430133 with the current signed release.

PR #428655, opened 2026-09-03
New package: richscripts.JavaScriptObfuscator version 3.4.5

richtexteditor:
### New package: `richscripts.JavaScriptObfuscator` version 3.4.5

JavaScript Obfuscator is a commercial JavaScript source-protection tool that has
shipped a Windows client since 2004. This adds the current WinUI 3 desktop build.

- Publisher: richscripts inc — https://javascriptobfuscator.com/
- Installer: `zip` with `NestedInstallerType: portable`, x64
- Aliases: `javascriptobfuscator` (GUI) and `jso-local` (CLI)

#### Checklist

- [x] Manifest passes `winget validate` (`Manifest validation succeeded`).
- [x] The installer URL is permanent. It is a versioned path published
      specifically so an `InstallerSha256` stays valid; the vendor's mutable
      `/download/javascriptobfuscator.zip` was deliberately **not** used.
- [x] `InstallerSha256` verified by downloading the URL and hashing the bytes.
      It matches the vendor's own published manifest at
      https://javascriptobfuscator.com/download/SHA256SUMS.txt
      (`E693523C9BC8D84BECC67820404DC166915D24B23E3C12DD504669353C84D8BD`,
      98,937,590 bytes).
- [x] Both `RelativeFilePath` entries were confirmed present in the archive.
- [x] Binaries are Authenticode signed as `CN=richscripts inc`.
- [x] Package is not already in this repository (no existing `richscripts`
      publisher directory, and no manifest matching `javascriptobfuscator`).
- [x] All URLs in the manifests return HTTP 200.

#### One thing I could not verify locally

I was unable to run `winget install --manifest` end to end, because installing
from a local manifest requires an administrator to enable
`LocalManifestFiles` on the machine and I did not have that. Everything above
was verified by other means, but the extract-and-alias step has only been
checked by confirming the nested paths exist inside the archive, not by
performing a real install. Please flag anything the CI catches there and I will
fix it promptly.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/428655)

### DIFF
--- a/manifests/r/richscripts/JavaScriptObfuscator/3.4.5/richscripts.JavaScriptObfuscator.installer.yaml
+++ b/manifests/r/richscripts/JavaScriptObfuscator/3.4.5/richscripts.JavaScriptObfuscator.installer.yaml
@@ -1,0 +1,17 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+
+PackageIdentifier: richscripts.JavaScriptObfuscator
+PackageVersion: 3.4.5
+InstallerType: zip
+NestedInstallerType: portable
+Installers:
+- Architecture: x64
+  NestedInstallerFiles:
+  - RelativeFilePath: javascriptobfuscator.exe
+    PortableCommandAlias: javascriptobfuscator
+  - RelativeFilePath: cli\jso-local.exe
+    PortableCommandAlias: jso-local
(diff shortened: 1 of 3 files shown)


richtexteditor:
@microsoft-github-policy-service agree company="richscripts inc"

PR #428949, opened 2026-09-03
Expand manifest diagnosis workflow

yao-msft:
Adds AppsAndFeatures transition and unexpected-file assistance using current trusted validation evidence. 
 
 Also fixed wingetvalidator app permission issue.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/428949)


denelon (dismissed):

ranm-msft (commented):
Not a change request on this PR's actual diff — but since it regenerates all three lockfiles, it may be the cheapest place to pick this up.

All three compile with the incomplete-report issue path open at `a8e441913c94e23a8fb5d1202cd6b6825d83227e`, and so does `master` today:

```yaml
GH_AW_REPORT_INCOMPLETE_CREATE_ISSUE: "true"
```

That leaves a route for an infrastructure or tool failure to create or update a repository issue from workflows whose intended output is a PR comment. The same gap exists in most of the new agentic workflows currently in review, so it looks systemic to the batch rather than specific to any one of them.

`domain-validation-assist` (#428961) is the one that pins it shut:

```yaml
report-incomplete:
  create-issue: false
```

Since these three lockfiles are already being regenerated here, folding it in now would avoid a separate sweep later. Entirely your call whether it belongs in this PR or a follow-up.

yao-msft:
> Not a change request on this PR's actual diff — but since it regenerates all three lockfiles, it may be the cheapest place to pick this up.
> 
> All three compile with the incomplete-report issue path open at `a8e441913c94e23a8fb5d1202cd6b6825d83227e`, and so does `master` today:
> 
> ```yaml
> GH_AW_REPORT_INCOMPLETE_CREATE_ISSUE: "true"
> ```
> 
> That leaves a route for an infrastructure or tool failure to create or update a repository issue from workflows whose intended output is a PR comment. The same gap exists in most of the new agentic workflows currently in review, so it looks systemic to the batch rather than specific to any one of them.
> 
> `domain-validation-assist` (#428961) is the one that pins it shut:
> 
> ```yaml
> report-incomplete:
>   create-issue: false
> ```
> 
> Since these three lockfiles are already being regenerated here, folding it in now would avoid a separate sweep later. Entirely your call whether it belongs in this PR or a follow-up.

Resolved

ranm-msft (commented):
Re-checked at `2a0a5ff7`. The `report-incomplete` pin landed in all three workflows and the lockfiles agree, so that one is closed from my side.

On the new evidence bounds: the `maxFailedChecks` / `maxCheckTextLength` rejection is enforced deterministically in `pre-agent-steps`, but nothing downstream enforces the consequence. `evidenceComplete` appears three times in `manifest-validation-diagnosis.md` - the initialiser, the assignment, and the mission line telling the agent to emit `noop`. The `post-pr-comment` job never reads `validation-checks.json`; it validates shape, head SHA, labels and dedupe against the event and a fresh `pulls.get`. So if the agent does emit a comment after the collector rejected the evidence, that comment can still pass the gate. Given the commit is titled "Fail closed on incomplete manifest evidence", it seems worth having the write job read a sealed envelope the way `label-reconciliation-assist` does in #428962. One wrinkle: the singleton path is allowed to proceed without check evidence, so a flat `evidenceComplete === true` gate would break it. An accepted-mode field would survive that better than a boolean.

Separately, and not introduced here: I diffed the denylists against the repo's actual label list. `Internal-Error-Dynamic-Scan` exists and is absent from every workflow in this batch, while `Validation-Hash-Error`, `Validation-Signature-Error` and `Possible-Malware` are not labels in this repo at all. In `missing-dependency-assist` and `wingetbot-pr-triage` the security stop is prompt text only - their compiled `add_comment` is `{"max":1}` with no label filter - so a stale name there fails open rather than closed. That reads like the centrally maintained set @denelon asked for rather than another per-file list.

ranm-msft (commented):
Re-checked at `6b246ce3`.

Both of my earlier items are closed. The unsafe set is now 41 names, every one of which matches a label that currently exists in this repo, and the `.md` and the three lockfiles agree exactly. `evidence_mode` is also a better answer than the boolean I suggested: `checks` enforces `evidenceComplete`, and the singleton path re-derives its own evidence at head rather than trusting the agent, requiring `ManifestType: singleton` and a unique identity that matches the path.

Two things on the new write gate.

The operation binding proves the sealed completion check and detail checks all belong to one operation, but every value being compared came from the same collection-time snapshot, so it establishes consistency rather than freshness. The write job re-reads `pulls.get` but never re-fetches Checks, and its permissions block is `contents: read` / `issues: write` / `pull-requests: read` with no `checks: read`, so a revalidation of the same commit cannot be seen at post time. Since dedupe keys on `Head SHA:`, the superseded comment lands first and then suppresses the corrected one. Adding `checks: read` and requiring the sealed operation to still be the newest completed one, with no newer queued or in-progress trusted check, would close it.

Separately, the live label gate only requires `activeSupported.length !== 0`, so a comment can still post once the triggering diagnosis is gone as long as some other supported label remains.

ranm-msft (commented):
Re-checked at `081419cf`. The label gate is what I asked for: both paths now require the triggering label to still be present at the final read, and the singleton path requires it to be the sole active diagnosis, so a comment landing after the diagnosis was removed is rejected whenever the removal is visible at that read. I also like that this one fails with `core.setFailed` rather than returning quietly, so a rejection is visible instead of looking like a healthy no-op.

One thing I would still like your read on before I approve. The fresh comparison covers the completion check thoroughly, but the detail checks are only re-verified for existence:

    evidence.checks.some((check) => !freshIds.has(check.id))

The comment body is built from the sealed outcomes of those detail checks. #428958 re-compares each one's fresh status and conclusion, and #428966 goes further and re-derives its finding from the freshly read output text. If the validator ever updates a check run in place rather than creating a new one, ids and external ids survive unchanged while the conclusion the advice was based on has moved, and only the stronger comparison catches it.

Do you know whether that validator reuses runs within an operation? If it does, borrowing the #428958 comparison seems worth it. If it provably never does, say so and I will approve as is.

yao-msft:
Updated in 104b6421da9 to avoid relying on the validator never updating a check run in place. The final gate now re-fetches each sealed detail check by ID and requires its name, completed status, and conclusion to still match before posting; the generated lockfile was refreshed with gh-aw v0.83.4.

ranm-msft (commented):
Re-checked at `104b6421`. The name / status / conclusion comparison landed, and building `freshById` only from runs whose `external_id` matches the sealed `operationId` closes the id-reuse case cleanly. That is the part I asked for.

One residual, and it is why I am holding rather than raising something new. The comment body is not built from those three fields. Step 5 has the agent extract the error lines from the selected check's `output.title`, `output.summary` and `output.text`. Those are sealed by `mapCheck` but never re-compared at write time. An in-place update that keeps the same id, name, `completed` status and `failure` conclusion while rewriting the text still passes the gate, so the advice can name the previous field or file. That is the scenario the commit message says it no longer relies on; the current comparison catches it only when the conclusion moves too.

`completedAt` is already in the sealed envelope, so comparing that, or the output fields directly, looks like a small addition. #428966 re-derives its finding from the freshly read output instead of trusting the sealed copy, if you would rather be consistent with that.

Nothing else changed for me on this pass. The other six in the batch are approved at their current heads, so this is the only one I am still holding.

ranm-msft (approved):
Verified at `f6d4e370`. This closes the item I raised last round.

The write gate now compares the sealed `output.title` / `output.summary` / `output.text` (plus `completed_at`) against the fresh read, so an in-place check-run update that preserves id, name, `completed` status and conclusion no longer slips through — which is the property the advice text actually depends on, since step 5 derives its error lines from exactly those fields. The normalization is symmetric with `mapCheck` (`?? null` for title/summary, `String(... ?? "")` for text) and both sides read the same `checks.listForRef` shape, so I don't expect this to fail closed spuriously on stable data. The `.md` and the compiled `.lock.yml` carry identical logic, and only `frontmatter_hash` moved — consistent with a frontmatter-only edit.

Two residual notes, neither blocking and neither needing action in this PR:

- The gate binds each sealed check individually, but not the completeness of the operation's failed-check set, so a check added later under the same `operationId` wouldn't be noticed. That makes the advice potentially incomplete rather than wrong, and the completion-check plus sequence gating already makes it unlikely.
- There's an irreducible window between the final re-fetch and `issues.createComment`. GitHub offers no atomic compare-and-comment, so that one is bound-the-window rather than close-it.

Both are acceptable for a fail-closed, advisory-only comment that never edits, labels, merges or approves. Approving.

PR #428962, opened 2026-09-03
Add label reconciliation assist agentic workflow

yao-msft:
Reconciles highest-version removals and stale Needs-CLA cases through advisory comments only. 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/428962)


denelon:
Two issues need to be addressed before this workflow can operate safely:

1. The body-validation regex `/[^@-^H\u000b^L^N-^_\u007f]/` matches ordinary characters, including the required `> [!WARNING]` prefix, so every valid advisory is rejected. Please replace it with the intended control-character check and regenerate the lockfile.
2. The final write gate validates only the opaque comment body, fixed target/head, and labels. Before posting, it should deterministically re-fetch and revalidate the candidate PR/head, human engagement, class-specific version/publication evidence, and current CLA Check/timeline facts. Otherwise a candidate or its evidence can change after collection while stale advice is still posted.

ranm-msft (commented):
Nice shape overall: `checkout: false`, a bounded evidence envelope collected deterministically in `pre-agent-steps`, a read-only agent job, and a separate safe job that re-validates before it writes. The output boundary is clearly deliberate.

One gap in that boundary at `b1d5b3ce79565c5a0bf6048f864bd8d995f98709`. `safe-outputs` here closes every issue-producing path explicitly:

```yaml
report-failure-as-issue: false
noop:
  report-as-issue: false
missing-tool: false
missing-data: false
```

but does not set `report-incomplete`, which defaults on with issue creation enabled. The compiled lockfile confirms it:

```yaml
GH_AW_REPORT_INCOMPLETE_CREATE_ISSUE: "true"
```

So an infrastructure or tool failure can still create or update a repository issue, even though the workflow's only intended output is a single PR comment and every sibling path is pinned shut.

`domain-validation-assist` (#428961) is the one workflow in this batch that closes it:

```yaml
report-incomplete:
  create-issue: false
```

That form keeps the incomplete signal but suppresses issue creation. Disabling `report-incomplete` outright is the stricter option if the signal isn't wanted at all. Either way the lockfile needs regenerating.

Which of the two do you want here, or is the issue path intentional and just the one that didn't get pinned?

ranm-msft:
@denelon — checking one detail on the body-validation note before this gets changed.

At head `b1d5b3ce79565c5a0bf6048f864bd8d995f98709`, both `label-reconciliation-assist.md` and the compiled `.lock.yml` contain this, in plain ASCII with no literal control bytes:

```js
/[\u0000-\u0008\u000b\u000c\u000e-\u001f\u007f]/.test(body) ||
```

Read as source, that is a positive class over selected C0 controls plus DEL — the first character inside the bracket is a backslash, not `^` — so `> [!WARNING]` does not match it, and a valid advisory is not rejected on that branch.

The `^@` / `^H` / `^L` / `^N` / `^_` in the quoted form looks like caret notation for those same control characters (NUL, BS, FF, SO, US). Is it possible the copy you were reading came from a renderer that interpreted the escapes, rather than from the regex source?

Not claiming the check is complete — it does not cover C1 controls — just worth confirming before a class that may already be doing the intended thing gets rewritten.

ranm-msft (dismissed):
The explicit `\uXXXX` rewrite settles the caret-notation question from the earlier thread. That form is unambiguous both to read and to review, which matters more here than brevity.

The change I like more is the binding: uploading the evidence in `pre-agent-steps` and pulling it back through `BINDING_PATH` in the write job, with `core.setFailed` when it is missing, is a real improvement over deriving the path from the agent output directory. That is the pattern I would point the rest of the batch at. #428949 still leaves the equivalent check to the prompt.

Approving.

ranm-msft (commented):
My approval was dismissed by `025bd9dd`, so I re-read the finalizer against @denelon's item 2.

Replacing the agent-authored body with structured fields and rendering in code covers the opaque-body half of it, and the exact-key-set check on the item is the right shape.

The class-specific evidence half still looks open. On the published-event path for `Highest-Version-Removal` the agent chooses `candidate_pull_number`, and the finalizer only confirms that PR is open, carries the label, matches the head the agent supplied, and is not the merged PR. Nothing ties the candidate to the publication that triggered the run, so any open PR carrying that label would pass and then be named as stale in the rendered finding. `Needs-CLA` is tighter, since the candidate is pinned to the target there.

The `unsafe` set here is missing the same current labels I listed on #428961.

ranm-msft (commented):
Re-checked at `cb89d466`.

The label item is closed. The `unsafe` set here is now the same 41 names the rest of the batch carries, every one matching a label that currently exists in this repo, and the `.md` and lockfile agree.

The candidate binding I raised last round reads the same at this head. On the published-event path the trigger side is now tightly gated, requiring the merged target, the policy-service sender and `Publish-Pipeline-Succeeded`, but the candidate itself is still only checked for being open, carrying `Highest-Version-Removal`, matching the head the agent supplied and not being the merged PR. Nothing relates it to the package that was just published, and the rendered evidence is only the candidate number, its head SHA and the publication PR number. So any open PR carrying that label satisfies the gate and would then be named as stale in a comment on the publication PR. `Needs-CLA` stays the tighter of the two, since the candidate is pinned to the target and the CLA check is bound by app ID, name, conclusion and the label timeline.

Not repeating the rest, just confirming it is unchanged rather than missed.

ranm-msft (approved):
Re-checked at `b4b6c033`. The candidate binding went further than I asked. The publication and candidate manifest file sets are parsed from `listFiles`, both must resolve to the same `PackageIdentifier` and the same package folder, the published version must compare greater than the removed one, and both folders are re-read at the default branch rather than trusted from the diff. The published-event path then requires exactly one open `Highest-Version-Removal` PR for that identifier, matching the candidate in evidence; the other path requires exactly one newer version directory under the package folder. And `Publish-Pipeline-Succeeded` is now backed by a wingetbot confirmation posted after the merge rather than by the label on its own. The label alone is no longer sufficient, which was the substance of the problem.

One residual worth knowing about rather than fixing now: the uniqueness step on the published-event path depends on the search index, which lags behind repository state by an interval you do not control. A second candidate opened moments earlier can be missing from the result set, so uniqueness reads as satisfied while it is not. The version-directory path has no such dependency, so if you ever want to tighten this, preferring a contents read over the search API is the lever.

Approving.

ranm-msft (dismissed):
Reviewed the frontmatter and the compiled lock, not just the prompt.

This is the pattern I'd want the rest of the set to follow. The safe-output job takes typed fields rather than free-form model prose, and the write step revalidates them against the sealed evidence artifact and current PR state before rendering the comment itself — so a prompt-injection attempt in untrusted manifest or PR text can at worst produce a rejected payload, not a plausible-looking advisory posted under the bot identity. Combined with `checkout: false`, read-only top-level permissions, pinned `allowed-repos`, `bash: ["cat"]` and `threat-detection: true`, the blast radius is about as small as it gets for a `pull_request_target` workflow.

Trigger is tightly scoped and doesn't collide with the others in the set.

LGTM.

ranm-msft (commented):
Read both the workflow source and the compiled `.lock.yml` for this one. The hardening is right where it matters: `checkout: false` on a `pull_request_target` trigger, agent permissions held to read-only plus `copilot-requests: write`, `allowed-repos` pinned to this repo, `bash: ["cat"]`, `threat-detection: true`, and a posting job whose target comes from trusted event context rather than from model output. The bounded evidence collection, with explicit byte and count caps, is a good pattern.

One architectural question before I sign off.

**The deterministic eligibility check does not gate the model invocation.** `Bind target and bounded reconciliation evidence` runs in `pre-agent-steps` and writes `eligible: false` unless the reconciliation target binds, but that result is consumed only by the agent prompt. In the generated workflow the `Execute GitHub Copilot CLI` step carries no corresponding condition - the `agent` job is gated solely on `needs.activation.outputs.daily_ai_credits_exceeded`. So an otherwise-activated `Publish-Pipeline-Succeeded` event still initializes the agent and reaches the model purely so it can observe that the run is ineligible and emit `noop`. That label fires on merged pull requests as a matter of course, so this is close to one agent initialization per merge.

Could the bounded probe run in an earlier job and expose an output used to gate the `agent` job through `needs`/`if`? Eligibility would stay necessary-but-not-sufficient, and the model would still make every substantive judgement whenever the probe passes.

Happy to sign off once you have weighed in - the question applies across this batch of workflows rather than to this PR alone.

yao-msft:
Updated in `a142e374900`. The deterministic pre-agent collector now writes a `noop` safe output when its existing eligibility contract fails, so the gh-aw harness exits before AI inference. The PR diff also no longer includes `.gitattributes` or `.github/aw/actions-lock.json`; those shared compiler updates will land through #432799 first.

ranm-msft (commented):

**As compiled, though, `Skip agent when reconciliation evidence is ineligible` will not write that noop - it fails the job instead.**

```
```

Every generated consumer re-maps it into its own step `env:` (`Start MCP Gateway`, `Execute GitHub Copilot CLI`, `Copy Safe Outputs`, `Ingest agent output`, `Parse agent logs`), and the only `$GITHUB_ENV` write in the whole file is `RUNNER_TOOL_CACHE`. The new step declares no `env:`, so `process.env.GH_AW_SAFE_OUTPUTS` is empty and the script takes its own guard branch - `core.setFailed("Safe outputs path is unavailable.")`. Every activated run where the collector returns `eligible !== true` then ends as a failed agent job rather than a quiet no-op, which is the opposite of the intent on the highest-volume path.

```yaml
env:
```

gh-aw's own `squad-game-planner` pre-agent step instead defaults it in shell (`${GH_AW_SAFE_OUTPUTS:-${RUNNER_TEMP:-/tmp}/gh-aw/safeoutputs/outputs.jsonl}`); either works, but any fallback needs to be `RUNNER_TEMP`-based rather than `/tmp`, or the harness will not read the file it wrote. The same shape is in all six workflows in this batch.

yao-msft:

## What changed

```yaml
env:
```

```js
const safeOutputsPath =
```

Still exactly 2 files here (head `eb615ac1b4c`) — `.gitattributes` and `.github/aw/actions-lock.json` remain owned by #432799, which should merge first.

ranm-msft (dismissed):
Verified, and this is exactly what I asked for. `id: set-runtime-paths` is declared ahead of the new gate step in the lockfile, so `${{ steps.set-runtime-paths.outputs.GH_AW_SAFE_OUTPUTS }}` resolves against a completed earlier step in the same job. The gate now appends the noop to the same file the harness reads instead of taking the `core.setFailed("Safe outputs path is unavailable.")` branch, and the secondary default is `RUNNER_TEMP`-based as needed. The `.md` and `.lock.yml` are consistent with a clean v0.88.7 recompile, and the identical change is present in all nine workflow files across this batch and #432799.

Approving.

yao-msft:
Pushed a one-line permissions fix to the comment safe-job.

A production run of the already-merged Manifest Validation Diagnosis workflow surfaced a latent bug shared by this one: the comment job failed with `403` on `POST /repos/{owner}/{repo}/issues/{number}/comments`. The API response named the requirement directly:

```
x-accepted-github-permissions: 'issues=write; pull_requests=write'
```

The `;` denotes **all required**, not alternatives — commenting on a pull request through the shared issue-comments endpoint needs both scopes. The safe-job declared `issues: write` with `pull-requests: read`, so every comment attempt would have been rejected. Failed run: https://github.com/microsoft/winget-pkgs/actions/runs/34744538124

This is not a new privilege. The `gh-aw` compiler already generates `issues: write` plus `pull-requests: write` for its own `safe_outputs` and `conclusion` jobs; the hand-authored safe-job was under-scoped relative to the toolchain's own baseline.

**Agent job permissions are unchanged and remain read-only.** The added scope applies only to the post-agent job that publishes the already-generated comment, so the agent gains nothing and the workflow stays recommend-only.

Scope of the change: one line of workflow source, plus the regenerated lockfile. `body_hash` is unchanged, confirming no prompt, classifier, trigger, concurrency, network, or safe-output changes. An audit of the safe-job confirmed `issues.createComment` is its only mutating API call, so `pull-requests: write` was the only scope requiring an upgrade.

Validation: `gh aw compile` and `gh aw validate` both pass; the regenerated lockfile changes exactly two hunks (frontmatter hash and the same permission line).

The equivalent fix for the merged workflow is in https://github.com/microsoft/winget-pkgs/pull/433981.

yao-msft:

```yaml
safe-outputs:
  report-failed-jobs: false
```

ranm-msft (commented):

Two things worth a reply:

yao-msft:

| --- | --- | --- |

### Change applied

ranm-msft (approved):
Verified the new commit. `issues: write` is gone from the PR-comment safe-job in both the workflow and the lockfile.

Also appreciated the empirical follow-up - single-permission jobs showing 201 with `pull-requests: write` alone and 403 with `issues: write` alone is a much better answer than either of our readings of the header docs, and it settles the semicolon question for good.

This one needed the most checking because of the two `commentTarget` branches, so I walked both rather than assuming. Both resolve through `github.rest.pulls.get`, so a number that is not a pull request never produces a comment target - the `workflow_dispatch` input is described as an open Highest-Version-Removal or Needs-CLA pull request, and the label path comes off `pull_request_target`. No issue-targeted path survives.
Checked the lockfile too: the diff is +1/-2 - the dropped permission line and a rolled `frontmatter_hash` - with `body_hash` unchanged, so the prompt, trigger and behavior are untouched. Net effect is strictly less scope than the merged baseline.

Thanks for chasing that down properly rather than taking my word for it. Signing off.

PR #428963, opened 2026-09-03
Add unattended artifact triage agentic workflow

yao-msft:
Experimental. Manual dispatch only.
Provides safe, evidence-bound metadata routing for unattended-validation screenshot artifacts. 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/428963)


denelon:
Two fail-closed fixes are needed before this workflow is ready:

1. Set `report-incomplete: false` under `safe-outputs` and regenerate the lockfile. The compiled workflow currently sets `GH_AW_REPORT_INCOMPLETE_CREATE_ISSUE: "true"`, so an infrastructure or tool failure can create or update a repository issue despite the workflow's no-issue-output contract.
2. Bind the selected validation operation ID into the evidence and re-fetch trusted check runs immediately before posting. The final gate currently rechecks the PR state, head, and labels but not whether the collected operation remains newest, so a same-head rerun can make the advisory stale before it is posted.

ranm-msft (commented):
Status check at `0c7ccba5`.

Item 1 is done. The source sets `report-incomplete.create-issue: false` and the compiled lockfile now emits `GH_AW_REPORT_INCOMPLETE_CREATE_ISSUE: "false"`, so both halves agree.

Item 2 is not. The sealed envelope carries `operationBound: true` rather than the operation ID itself, and the finalizer never calls the Checks API, so at posting time there is nothing left to compare against. The queued/in_progress rejection in the collector is worth keeping, but it was already in the first commit and it only establishes freshness at the moment the artifact is written, which is the window the feedback was pointing at.

One thing I did like here: the finalizer rebuilds the entire comment from the sealed fields and requires exact equality before posting.

The `unsafeLabels` set has the same coverage problem I described on #428961.

ranm-msft (commented):
Re-checked at `046bc946`.

The label item is closed. `unsafeLabels` is now the same 41 names as the rest of the batch, every one matching a label that currently exists in this repo, and it is applied on both the collection side and the write gate with the `.md` and lockfile in agreement.

Item 2 is unchanged. The collector still derives a real `operationId` and validates it against `completion.external_id` and the artifact payload, but what it seals is `operationBound: true`, and the gate asserts `evidence.operationBound !== true`. That is a boolean recording that a check passed earlier, so at posting time there is still nothing to compare against. Worth noting this one is dispatch-triggered rather than label-triggered, so the exposure is narrower than in the `pull_request_target` workflows: a newer validation operation does not itself start another run of this workflow, it just means the sealed operation can be superseded before the dispatched run posts. Sealing the ID itself and having the write job re-fetch the trusted checks for the live head would close it, which needs `checks: read` added to the write job.

The deterministic rebuild of the whole comment from sealed fields with an exact-equality check before posting is still the best version of that pattern in the batch.

ranm-msft (dismissed):
Re-checked at `5464d752`. This is the change I was hoping for rather than a patch over the symptom. What used to be sealed was `operationBound: true`, a boolean recording that a check had passed at collection time, which left the write job with nothing to compare against. The envelope now carries the operation id itself, so the final read has something to disagree with, and the reordering puts that read after the label and head gates instead of before them.

The gate rejects a higher operation sequence, a completion payload whose `OperationId` no longer matches, and separately any trusted check that is queued or in progress under a different operation. That last clause is the one I would keep: it catches a revalidation that has started and registered but not yet finished, which the sequence test alone would only catch once the new run existed with a parsable id.

Being dispatch-triggered, the exposure was narrower here to begin with. I would still not read this as making the sequence atomic: an operation requested but not yet represented by any check run is invisible to any read, and the marker is still the head SHA alone, so a comment that wins that race suppresses the correction. Non-blocking for advisory output.

Approving.

ranm-msft (commented):
Correction — please disregard my previous comment on this PR, it was wrong.

I claimed the safe-output job passes model-authored prose through with only shape validation. It doesn't. The finalizer builds `expectedBody` from the sealed fields and requires exact equality before posting, so the generated text is bound to the validated evidence rather than trusted on its own. That's the pattern I was asking for, already present.

Apologies for the noise. Nothing outstanding from me here.

ranm-msft (commented):
Read the workflow source and the compiled `.lock.yml`. This is the narrowest of the six workflows in the batch and I have no blocking concerns with it: `workflow_dispatch` only, `roles: [admin, maintainer, write]`, `checkout: false`, agent permissions read-only apart from `copilot-requests: write`, `tools` limited to `bash: ["cat"]` with no `github:` block at all, `threat-detection: true`, and a posting job that takes its target from dispatch input rather than from model output. `Collect safe unattended artifact metadata` keeps the evidence bounded before the agent ever sees it.

One shared observation, raised here only for consistency across the batch rather than as an objection to this PR.

In the generated workflow the `Execute GitHub Copilot CLI` step carries no condition derived from the pre-agent classification - the `agent` job is gated solely on `needs.activation.outputs.daily_ai_credits_exceeded`. Here that is close to free, because dispatch is manual and role-gated, so a human already decided the run was worth making. It matters more on the `Validation-Completed` siblings (#428958, #428967), where the same shape means an agent initialization per validated pull request. If the answer there is to compute eligibility in an earlier job and gate the `agent` job via `needs`/`if`, it would be worth applying the same shape here so all six stay consistent.

Otherwise this looks good to me.

yao-msft:
Updated in `8d82092368c`. The deterministic pre-agent collector now writes a `noop` safe output when its existing eligibility contract fails, so the gh-aw harness exits before AI inference. The PR diff also no longer includes `.gitattributes` or `.github/aw/actions-lock.json`; those shared compiler updates will land through #432799 first.

ranm-msft (commented):

**As compiled, though, `Skip agent when artifact evidence is unavailable` will not write that noop - it fails the job instead.**

```
```

Every generated consumer re-maps it into its own step `env:` (`Start MCP Gateway`, `Execute GitHub Copilot CLI`, `Copy Safe Outputs`, `Ingest agent output`, `Parse agent logs`), and the only `$GITHUB_ENV` write in the whole file is `RUNNER_TOOL_CACHE`. The new step declares no `env:`, so `process.env.GH_AW_SAFE_OUTPUTS` is empty and the script takes its own guard branch - `core.setFailed("Safe outputs path is unavailable.")`. Every activated run where the collector returns `available !== true` then ends as a failed agent job rather than a quiet no-op, which is the opposite of the intent on the highest-volume path.

```yaml
env:
```

gh-aw's own `squad-game-planner` pre-agent step instead defaults it in shell (`${GH_AW_SAFE_OUTPUTS:-${RUNNER_TEMP:-/tmp}/gh-aw/safeoutputs/outputs.jsonl}`); either works, but any fallback needs to be `RUNNER_TEMP`-based rather than `/tmp`, or the harness will not read the file it wrote. Dispatch-only means the volume here is low, but the same shape is in all six workflows in this batch.

yao-msft:

## What changed

```yaml
env:
```

```js
const safeOutputsPath =
```

Still exactly 2 files here (head `3e9d2aa248e`) — `.gitattributes` and `.github/aw/actions-lock.json` remain owned by #432799, which should merge first.

ranm-msft (dismissed):
Verified, and this is exactly what I asked for. `id: set-runtime-paths` is declared ahead of the new gate step in the lockfile, so `${{ steps.set-runtime-paths.outputs.GH_AW_SAFE_OUTPUTS }}` resolves against a completed earlier step in the same job. The gate now appends the noop to the same file the harness reads instead of taking the `core.setFailed("Safe outputs path is unavailable.")` branch, and the secondary default is `RUNNER_TEMP`-based as needed. The `.md` and `.lock.yml` are consistent with a clean v0.88.7 recompile, and the identical change is present in all nine workflow files across this batch and #432799.

Approving.

yao-msft:
Pushed a one-line permissions fix to the comment safe-job.

A production run of the already-merged Manifest Validation Diagnosis workflow surfaced a latent bug shared by this one: the comment job failed with `403` on `POST /repos/{owner}/{repo}/issues/{number}/comments`. The API response named the requirement directly:

```
x-accepted-github-permissions: 'issues=write; pull_requests=write'
```

The `;` denotes **all required**, not alternatives — commenting on a pull request through the shared issue-comments endpoint needs both scopes. The safe-job declared `issues: write` with `pull-requests: read`, so every comment attempt would have been rejected. Failed run: https://github.com/microsoft/winget-pkgs/actions/runs/34744538124

This is not a new privilege. The `gh-aw` compiler already generates `issues: write` plus `pull-requests: write` for its own `safe_outputs` and `conclusion` jobs; the hand-authored safe-job was under-scoped relative to the toolchain's own baseline.

**Agent job permissions are unchanged and remain read-only.** The added scope applies only to the post-agent job that publishes the already-generated comment, so the agent gains nothing and the workflow stays recommend-only.

Scope of the change: one line of workflow source, plus the regenerated lockfile. `body_hash` is unchanged, confirming no prompt, classifier, trigger, concurrency, network, or safe-output changes. An audit of the safe-job confirmed `issues.createComment` is its only mutating API call, so `pull-requests: write` was the only scope requiring an upgrade.

Validation: `gh aw compile` and `gh aw validate` both pass; the regenerated lockfile changes exactly two hunks (frontmatter hash and the same permission line).

The equivalent fix for the merged workflow is in https://github.com/microsoft/winget-pkgs/pull/433981.

yao-msft:

```yaml
safe-outputs:
  report-failed-jobs: false
```

ranm-msft (commented):

Two things worth a reply:

yao-msft:

| --- | --- | --- |

### Change applied

ranm-msft (approved):
Verified the new commit. `issues: write` is gone from the PR-comment safe-job in both the workflow and the lockfile.

Also appreciated the empirical follow-up - single-permission jobs showing 201 with `pull-requests: write` alone and 403 with `issues: write` alone is a much better answer than either of our readings of the header docs, and it settles the semicolon question for good.

Checked the drop is not a breaking change here rather than assuming it. This is the simplest case of the set: `workflow_dispatch` is the only trigger and its single required input is documented as the pull request number to triage, so there is exactly one comment target and it is always a pull request.
Checked the lockfile too: the diff is +1/-2 - the dropped permission line and a rolled `frontmatter_hash` - with `body_hash` unchanged, so the prompt, trigger and behavior are untouched. Net effect is strictly less scope than the merged baseline.

Thanks for chasing that down properly rather than taking my word for it. Signing off.

PR #429007, opened 2026-09-03
New package: MichalFoist.mailnino version 1.1.1

MadMajkl:
### Add MichalFoist.mailnino version 1.0.8

First submission of this package.

mailniño is a local-first e-mail client for Windows (IMAP/SMTP). It keeps mail, index
and settings on the user's machine, has no accounts or servers of its own, and stores
passwords in the Windows credential manager.

#### Checklist

- [x] Have you checked that there aren't other open pull requests for the same manifest?
      → `Manifest validation succeeded.`
- [x] Does your manifest conform to the 1.6 schema?
      → **Not run.** Local manifest installs require `winget settings --enable LocalManifestFiles`,
      which needs administrator rights that were not available. The installer itself was
      verified end to end: downloaded from the `InstallerUrl` above, SHA-256 matches the
      value in the manifest, the Authenticode signature verifies as valid (Certum,
      timestamped), and the NSIS silent switch `/S` is confirmed working.
- CLA: will sign when the bot asks.

#### Notes for reviewers

- **`AppsAndFeaturesEntries.Publisher` is `pcrada`, while the top-level `Publisher` is
  `Michal Foist`.** This is intentional and accurate, not a copy-paste error: the
  installer is built with Tauri, whose NSIS template writes the second segment of the
  application identifier (`cz.pcrada.mailnino`) into the Add/Remove Programs entry when
  no publisher is configured. The code-signing certificate is issued to **Michal Foist**.
  The next release sets the publisher explicitly, and this entry will be updated then.
- **Per-user install** (`Scope: user`, installs into `%LOCALAPPDATA%`), so no elevation
  is needed.
- The installer is signed with a timestamped certificate from Certum
  (`CN=Michal Foist`, valid until 2027-07-20).
- Only one version of this package will be published at a time: when a new version is
  released, the previous installer is removed from the download server for security
  reasons, so each future pull request will add the new version **and remove the old
  one** rather than leaving a manifest pointing at a URL that no longer resolves.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/429007)


MadMajkl:

PR #429008, opened 2026-09-03
New package: OBD2FaultFinder.OBD2FaultFinder version 1.1.3

onceuponajuan:
## 📖 Description

OBD2 FaultFinder is a desktop OBD2 diagnostic program for Windows and Linux. It reads trouble codes and gives the standard name of each in plain language, shows around thirty live engine readings, records a drive to a file and plays it back, and compares two recordings side by side. Every screen is in English and Spanish.

I am the author and publisher.

- Installer: NSIS, x64, per-user (`perMachine: false`), silent switch `/S`
- `InstallerUrl` points at the publisher's own download host and is versioned, so the address will not change
- `InstallerSha256` verified against the published file
- Locales: en-US (default) and es-MX

## ✅ Checklist

## 📦 Manifest Checklist

- [x] Validated manifest locally with `winget validate --manifest <path>` (reports: Manifest validation succeeded)

> Manifests are authored against schema 1.6.0, which `winget validate` accepts. Happy to regenerate against 1.12.0 if that is required for new packages.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/429008)


onceuponajuan:

onceuponajuan:
Updated this PR from 1.1.2 to 1.1.3 before review, for two reasons:

1. The 1.1.2 description named a specific adapter chipset. That misdescribes the program — it works with OBD2 adapters generally — and I would rather not have it enter the catalogue that way.
2. 1.1.3 is now the published version, and it fixes an incorrect reading in the app's own diagnostics, so it is the one worth shipping.

Apologies for the `PullRequest-Error` label: it was raised in the moment between the two commits, when both version folders briefly existed. The next commit removed 1.1.2, so the branch now contains a single version directory with the four manifests, and validation has re-run against that state.

`winget validate --manifest` reports "Manifest validation succeeded" for 1.1.3, and its `InstallerSha256` was checked against the file actually served at the InstallerUrl.

onceuponajuan:
@wingetbot run

Validation has since completed cleanly on this branch: all 11 checks are green (the last finished at 21:53Z), and the PR now contains a single version directory, 1.1.3, with the four manifests.

The `PullRequest-Error` label was raised at 21:11Z during the brief window between two commits, when the 1.1.2 folder still existed alongside 1.1.3. The following commit removed it. The label appears to be stale rather than reflecting the current state — flagging it here so it does not put moderators off reviewing. Thank you for your time.

Trenly (approved):

PR #429465, opened 2026-09-04
New package: anttiak.LoFeline version 1.0.0

anttiak:
A pixel cat that naps in the Windows notification area and plays lofi radio.

- Repository: https://github.com/anttiak/lofeline-windows
- Release: https://github.com/anttiak/lofeline-windows/releases/tag/v1.0.0
- Installer: Inno Setup, per-user scope, so silent switches are supplied automatically by the `inno` installer type.

### Checklist

- [x] Have you checked that there aren't other open pull requests for the same manifest?
- [x] Does your manifest conform to the 1.12 schema?

`winget install --manifest` completed successfully: the installer downloaded from the release URL, the hash verified, and the package installed silently without elevation. Uninstall removes the program directory, the Start Menu entry and the run-at-startup registry value the app writes for itself.

Other checks:

- The Add/Remove Programs entry reads `LoFeline` / `Antti Ala-Könni`, matching `PackageName` and `Publisher`.
- `ProductCode` was read from the registry after a real install rather than inferred from Inno's naming convention.
- `InstallerSha256` was taken from the published release asset, re-downloaded from the release URL and hashed.

### DIFF
--- a/manifests/a/anttiak/LoFeline/1.0.0/anttiak.LoFeline.installer.yaml
+++ b/manifests/a/anttiak/LoFeline/1.0.0/anttiak.LoFeline.installer.yaml
@@ -1,0 +1,22 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: anttiak.LoFeline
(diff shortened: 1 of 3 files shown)


anttiak:

PR #430045, opened 2026-09-05
New package: FiredMosquito831.MyClaudeCode version 7.13.1

FiredMosquito831:
### Package

`FiredMosquito831.MyClaudeCode` 6.45.2 — the My Claude Code desktop app, a small native window (~3.5 MB installed) onto the dashboard the project's local server already serves on 127.0.0.1.

Homepage: https://github.com/FiredMosquito831/my-claude-code
Installer: https://github.com/FiredMosquito831/my-claude-code/releases/download/v6.45.2/MyClaudeCode-Setup-windows-x86_64.exe

### Checklist

- [ ] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)? — **not yet; will sign when the CLA bot posts the link.** Left honestly unchecked rather than pre-ticked.

### Notes for the reviewer

- **Per-user Inno Setup installer**, `PrivilegesRequired=lowest`. `Scope: user`; there is no machine-scope installer to offer.
- **Silent install and silent uninstall both verified**, using the default Inno switches winget supplies (`/SP- /VERYSILENT /SUPPRESSMSGBOXES /NORESTART`) and the registered `QuietUninstallString`. No `InstallerSwitches` block is present because none is needed.
- **`AppsAndFeaturesEntries` mirrors the registry exactly**: the installer's `AppId` is fixed forever, so the uninstall key is always `{5FC8D5C3-33F7-4366-AD8D-C844D21BC089}_is1`. With the app installed, `winget list` reports the package as `ARP\User\X64\{5FC8D5C3-33F7-4366-AD8D-C844D21BC089}_is1`.
- **Uninstall leaves nothing behind.** `HKCU\…\Uninstall`, `HKCU\…\Run`, both Start Menu Programs trees, `%LOCALAPPDATA%\Programs` and `~/.local/bin` were snapshotted before install and diffed after uninstall; all five diffs were empty.
- **The installer is unsigned and will remain so.** SmartScreen shows its reputation warning on first run; this is documented in the project's README rather than papered over. The package carries no `Commands`, because the installer deliberately puts nothing on `PATH`.
- **`InstallerSha256` re-verified against the live release asset immediately before opening this PR** — `FB78D850…DBFF8F`, matching byte for byte.
- The manifests are generated from the release by [`desktop-shell/installer/winget/render.py`](https://github.com/FiredMosquito831/my-claude-code/blob/main/desktop-shell/installer/winget/render.py) in the source repository, and a test there asserts the committed copies are byte-for-byte what it produces.

Validation environment: winget v1.29.290 on Windows 11 build 26100.

### DIFF
--- a/manifests/f/FiredMosquito831/MyClaudeCode/6.45.2/FiredMosquito831.MyClaudeCode.installer.yaml
+++ b/manifests/f/FiredMosquito831/MyClaudeCode/6.45.2/FiredMosquito831.MyClaudeCode.installer.yaml
@@ -1,0 +1,21 @@
+# Rendered by desktop-shell/installer/winget/render.py -- do not edit by hand.
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.28.0.schema.json
+
+PackageIdentifier: FiredMosquito831.MyClaudeCode
+PackageVersion: 6.45.2
+Installers:
+  - Architecture: x64
+    InstallerType: inno
+    Scope: user
+    InstallerUrl: https://github.com/FiredMosquito831/my-claude-code/releases/download/v6.45.2/MyClaudeCode-Setup-windows-x86_64.exe
+    InstallerSha256: FB78D8505AE2659BBBEDEDB81E759A38FE65310CEFFF282236B21C9980DBFF8F
+    ProductCode: '{5FC8D5C3-33F7-4366-AD8D-C844D21BC089}_is1'
+    AppsAndFeaturesEntries:
+      - DisplayName: My Claude Code (desktop app)
+        DisplayVersion: 6.45.2
+        Publisher: My Claude Code
+        ProductCode: '{5FC8D5C3-33F7-4366-AD8D-C844D21BC089}_is1'
+        InstallerType: inno
+    ReleaseDate: 2026-09-04
+ManifestType: installer
+ManifestVersion: 1.28.0
--- a/manifests/f/FiredMosquito831/MyClaudeCode/6.45.2/FiredMosquito831.MyClaudeCode.installer.yaml
+++ b/manifests/f/FiredMosquito831/MyClaudeCode/6.45.2/FiredMosquito831.MyClaudeCode.installer.yaml
@@ -1,5 +1,5 @@
 # Rendered by desktop-shell/installer/winget/render.py -- do not edit by hand.
-# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.28.0.schema.json
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
 
 PackageIdentifier: FiredMosquito831.MyClaudeCode
 PackageVersion: 6.45.2
@@ -18,4 +18,4 @@ Installers:
         InstallerType: inno
     ReleaseDate: 2026-09-04
 ManifestType: installer
-ManifestVersion: 1.28.0
+ManifestVersion: 1.12.0
--- a/manifests/f/FiredMosquito831/MyClaudeCode/6.45.2/FiredMosquito831.MyClaudeCode.locale.en-US.yaml
+++ b/manifests/f/FiredMosquito831/MyClaudeCode/6.45.2/FiredMosquito831.MyClaudeCode.locale.en-US.yaml
@@ -1,0 +1,46 @@
+# Rendered by desktop-shell/installer/winget/render.py -- do not edit by hand.
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.defaultLocale.1.28.0.schema.json
+
+PackageIdentifier: FiredMosquito831.MyClaudeCode
+PackageVersion: 6.45.2
+PackageLocale: en-US
+Publisher: FiredMosquito831
+PublisherUrl: https://github.com/FiredMosquito831
+PublisherSupportUrl: https://github.com/FiredMosquito831/my-claude-code/issues
+Author: FiredMosquito831
+PackageName: My Claude Code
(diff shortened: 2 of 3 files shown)


FiredMosquito831:
Corrected: the three manifests now declare `ManifestVersion: 1.12.0` (and the matching `$schema` URLs) instead of 1.28.0, which is not currently approved for release. Only those six lines changed — no field was added, removed or altered, and every field in use (`AppsAndFeaturesEntries`, `ReleaseDate`, `Documentations`) predates 1.12.0. The `InstallerSha256` is unchanged and still matches the live release asset.

The CLA is being signed separately; sorry for the two round trips.

stephengillie:
> 
> Executable C:\Users\validator\AppData\Local\Programs\My Claude Code\MyClaudeCode.exe returned exit code: -1073741515 
 
FiredMosquito831:

FiredMosquito831:
Superseding 6.45.2 with **7.13.1**, and fixing the executable validation failure.

### The `Validation-Executable-Error`

The `-1073741515` (`0xC0000135`, `STATUS_DLL_NOT_FOUND`) on the validator VM was the **Edge WebView2 runtime**. This is a Tauri app, so its window is drawn by WebView2 — a system component the package does not bundle — and the executable cannot resolve its imports without it, so Windows terminates it before any of the app's own code runs.

The installer does carry a WebView2 bootstrapper, but it invokes it only when its EdgeUpdate `pv` registry probe reports the runtime missing, which does not hold inside the validation image. So the package installed correctly and the thing it installed could not start.

The installer manifest now declares the dependency:

```yaml
Dependencies:
  PackageDependencies:
    - PackageIdentifier: Microsoft.EdgeWebView2Runtime
```

so the runtime is installed before the package on any machine that lacks it — which is what the validator needs and equally what a user on a fresh Windows install needs. No `MinimumVersion` is pinned: every runtime version published here is newer than the one Tauri requires, and a floor that has not been tested against would only start failing installs later.

### Why the version moved

6.45.2 was many releases old by the time this was diagnosed. 7.13.1 is the current release; the manifests are generated from it by [`render.py`](https://github.com/FiredMosquito831/my-claude-code/blob/main/desktop-shell/installer/winget/render.py) in the source repository, where a test asserts the committed copies are byte-for-byte what it produces, and the `InstallerSha256` is read from the release's own `SHA256SUMS-desktop-shell.txt` rather than typed.

Everything else about the package is unchanged: per-user Inno Setup installer, `PrivilegesRequired=lowest`, `Scope: user`, no `InstallerSwitches` (the default Inno set winget supplies is what it is built and tested for), no `Commands` (the installer deliberately puts nothing on `PATH`), and `AppsAndFeaturesEntries` mirroring the `{5FC8D5C3-33F7-4366-AD8D-C844D21BC089}_is1` uninstall key exactly. The installer remains unsigned, as noted in the original description.

The CLA is signed.

FiredMosquito831:
Superseding 7.13.1 with **7.35.2**. The previous update misdiagnosed the
executable failure; this one fixes it in the build.

`0xC0000135` (`STATUS_DLL_NOT_FOUND`) is raised by the Windows loader while it
resolves the executable's imports — before any of the program's own code runs.
That rules out the Edge WebView2 runtime, which this app does not import at all:
Tauri links `WebView2Loader` statically and asks for the runtime at run time, so
a missing runtime is a failed window, not a dead process. Declaring
`Microsoft.EdgeWebView2Runtime` as a dependency in 7.13.1 therefore changed
nothing, as the re-run showed.

The import table named the real cause:

```
$ objdump -p MyClaudeCode.exe | grep 'DLL Name'
    DLL Name: VCRUNTIME140.dll      <- Visual C++ 2015-2022 redistributable
    DLL Name: VCRUNTIME140_1.dll    <- Visual C++ 2015-2022 redistributable
    DLL Name: api-ms-win-crt-*.dll  (UCRT, part of Windows 10+)
    DLL Name: kernel32 / user32 / ole32 / shell32 / ... (Windows)
```

which is exactly what the validation comment's own table suggests first
(`Microsoft.VCRedist.2015+.x64`). A clean image does not carry it.

**What changed for 7.35.2**

* the Windows binary is built with a statically linked CRT
  (`-C target-feature=+crt-static`), so it imports nothing Windows does not
  ship. Verified on the published release itself: 14 imported DLLs, all system.
  The release smoke now parses the built binary's PE import table and fails if
  any redistributable DLL returns. Bundling `VC_redist.x64.exe` was rejected —
  25 MB, and it needs administrator rights this per-user installer does not have;
* the installer now **carries** the WebView2 Evergreen bootstrapper (1.8 MB,
  embedded, run only when the runtime is absent) instead of downloading it
  during the install, and its WebView2 probe now requires `msedgewebview2.exe` to
  be present on disk rather than trusting the `pv` registry value alone;
* `Dependencies.PackageDependencies: [Microsoft.EdgeWebView2Runtime]` is kept. It
  was not the fix, but a user on a fresh Windows install still needs the runtime
  for the window to open.

Nothing else about the package changed: per-user Inno Setup installer,
`PrivilegesRequired=lowest`, `Scope: user`, no `InstallerSwitches`, no
`Commands`, `AppsAndFeaturesEntries` mirroring the
`{5FC8D5C3-33F7-4366-AD8D-C844D21BC089}_is1` uninstall key, `ManifestVersion:
1.12.0`, and the installer remains unsigned. The manifests are generated from
the release by
[`render.py`](https://github.com/FiredMosquito831/my-claude-code/blob/main/desktop-shell/installer/winget/render.py),
with the `InstallerSha256` read from the release's own
`SHA256SUMS-desktop-shell.txt`.

PR #430357, opened 2026-09-06
New package: EmreO33.Snag version 1.3.5

EmreO33:
### New package: EmreO33.Snag version 0.4.1

Snag is a small native desktop front end for yt-dlp, written in Rust. No webview,
no bundled runtime: a single executable that starts instantly.

- Homepage: https://github.com/EmreO33/snag
- License: GPL-3.0-or-later
- Installer: Inno Setup, per-user by default, supports silent and
  silent-with-progress installs and uninstalls cleanly.

Snag does not bundle yt-dlp or ffmpeg. It offers to download yt-dlp from that
project's own GitHub releases on first launch, and ffmpeg is left to the user.

---

      Yes: "Manifest validation succeeded."
      Not via `--manifest`, as that needs `LocalManifestFiles` enabled by an
      administrator. The installer itself was tested directly on Windows 11 x64:
      silent install, per-user registry entry under HKCU, Start Menu shortcut,
      the desktop-icon task correctly skipped when disabled, the application
      launches, and the uninstaller removes the install directory, the Start Menu
      folder and the registry entry completely.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/430357)


EmreO33:

EmreO33:
Following the guidance for `Validation-Executable-Error` ("if your application does not install an application, but should still be included in the repository, add a comment").

Snag does install an executable, so I've pushed a manifest update declaring the Add/Remove Programs entry explicitly, which should give validation something concrete to correlate against:

```yaml
AppsAndFeaturesEntries:
  - DisplayName: Snag 0.4.1
    Publisher: EmreO33
    DisplayVersion: 0.4.1
    ProductCode: '{CF8CC397-F68D-4940-8897-10A39126914A}_is1'
    InstallerType: inno
```

Those values are read from a real install rather than guessed. Verified on Windows 11 x64, installing with `/VERYSILENT /SUPPRESSMSGBOXES /NORESTART`:

- installs to the chosen directory, exit code 0
- registers under `HKCU\Software\Microsoft\Windows\CurrentVersion\Uninstall\{CF8CC397-F68D-4940-8897-10A39126914A}_is1`
- `snag.exe` runs and opens its window
- the uninstaller removes the install directory, the Start Menu folder and the registry entry completely, leaving nothing behind

The one thing worth flagging: **this is a per-user install**, so there is no `HKLM` entry to find, only `HKCU`. The installer is Inno Setup with `PrivilegesRequired=lowest`, which is why the manifest declares `Scope: user`. If validation looks for a machine-scope entry, that would explain not locating it.

Snag is also a GUI application with no console output, in case the check expects an executable that runs and exits.

Happy to change anything that would help — including shipping a per-machine installer instead, if that suits validation better.

EmreO33:
Updated this PR to **1.2.0**, which is the current release. The previous manifest was for 0.4.1 (the title said 0.4.2, which was my mistake) and Snag has had several releases since this was opened, so publishing that version would have shipped something well out of date.

Nothing structural has changed: same package, same per-user Inno installer, same `ProductCode`. Only the version, URL, hash, release date and `AppsAndFeaturesEntries` version strings differ.

The installer hash was taken from the release's published `SHA256SUMS.txt` and then verified again against the file downloaded from the exact URL in the manifest:

```
193181F30936780B65AB85D1D40ACDF1225F30B68F5C19EFA0CD7C9871D6FCC4  Snag-1.2.0-windows-setup.exe
```

`winget validate --manifest` passes locally.

The `Validation-Executable-Error` note from my earlier comment still applies: Snag is a per-user install (Inno with `PrivilegesRequired=lowest`, `Scope: user`), so its Add/Remove Programs entry is under `HKCU` and there is no `HKLM` entry to find. It is also a GUI application with no console output. Happy to switch to a per-machine installer if that suits validation better.

stephengillie:
> 
> Executable C:\Users\validator\AppData\Local\Programs\Snag\snag.exe returned exit code: -1073741515 
 
EmreO33:
Thank you, that was a real bug and not a validation quirk. The binary imported `VCRUNTIME140.dll`, which a clean Windows does not have, so it exited with `STATUS_DLL_NOT_FOUND` before drawing a window. Most machines have the runtime from other software, which is why it had gone unnoticed.

Fixed in **1.3.3**, now in this PR: the C runtime is linked statically, so the binary has no dependency on the Visual C++ Redistributable. Confirmed on the import table:

- 1.3.2 (previous): `vcruntime140.dll` plus six `api-ms-win-crt-*` forwarders
- 1.3.3 (this PR): 20 imports, none from the VC++ runtime

The release workflow now checks the finished binary's imports and fails if that dependency ever returns.

The installer hash was verified against the file downloaded from the manifest URL, and `winget validate --manifest` passes.

PR #430545, opened 2026-09-06
New package: NeverAwayTeams.NeverAway version 2.0.10

Neverawayteams:
### New package: NeverAwayTeams.NeverAway version 2.0.10

A portable, signed single executable distributed as a zip (`installerType: zip`,
`nestedInstallerType: portable`). The hosting account will not serve `.exe` files, which is why the
product has always been distributed as an archive.

The installer URL carries the version and is immutable once published, so the pinned hash cannot be
invalidated by a later release. The hash was verified by downloading the public URL back over HTTPS,
without a cache-busting query string, and comparing it with the signed local build.

This pull request was opened for an earlier version and has been updated in place rather than replaced,
because it has not been merged. What 2.0.10 contains:

- **The free and paid versions differ where it matters.** Scheduling, application control and the Info
  Hub were refused by the engine without a paid plan but left fully editable in the interface, so a free
  user could configure a week of automation and watch nothing happen. They are now shown as locked, in
  place, and selecting one shows the plans. What the free version is entitled to — keeping the status
  awake, its own tuning, the activity list and the window's behaviour — stays whole and configurable.
- **The window no longer calls itself LIMITED to a working free installation.** It displayed that, with
  "account verification unavailable" beneath it, to a user whose account had verified and whose status
  was being kept awake throughout. It now states the presence it is keeping.
- **The funding invitation shows a percentage** of the month covered, not the amounts.
- **The running version is shown in Settings → Account and support**, above the update check, whatever
  the plan. Read from the assembly rather than from a literal.
- **Waking from sleep or hibernation brings the application back.** It watched only the resume
  notification Windows sends for the classic sleep states, which a Modern Standby machine never sends.
  The suspend is now measured from Windows' own two interrupt-time counters — one counts time spent
  asleep, the other does not, and the difference is the suspend. Verified against a real hibernation:
  99 seconds down, reported as 99 seconds.
- **Teams was reported as permanently in a meeting.** `ms-teams.exe` holds two UDP sockets from the
  moment it starts, for its own signalling, and those were read as call media, so the scheduled close was
  postponed indefinitely. They are dual-stack, which the system tables list twice — once as `::` and once
  as `0.0.0.0` — so filtering by address family did not help; they are now identified as one socket seen
  twice and discounted.

Verification: 2368 deterministic checks in the product's own suite, the signed executable exercised
against a real S3 suspend, a real hibernation and two long process stalls, and the free and premium
layouts inspected on screen.


Neverawayteams:
Write @microsoft-github-policy-service agree

Neverawayteams:

PR #431082, opened 2026-09-07
New version: TechyGeeksHome.TranscribeGeek version 1.1.2

techygeekshome:
## Description

TechyGeeksHome.TranscribeGeek 1.1.2.

This replaces #426513, which was opened for 1.0.1 and has now been closed. 1.1.0, 1.1.1 and 1.1.2 have been released since, so this keeps one current manifest per package.

1.1.2 fixes a fault that stopped transcription working on every earlier release. The speech library was packed inside the single file executable and the loader could not find it at runtime. It is now written to the user profile and loaded from there, verified by size and SHA-256.

ProductCode is unchanged from earlier versions. The Inno Setup AppId is fixed across versions by design.

## Checklist

- [ ] Signed the Contributor License Agreement

## Manifest Checklist

The last two are unticked deliberately rather than overlooked. This manifest was prepared without a Windows host to run winget validate on. What was done instead: the installer was downloaded from the release and its SHA-256 computed and compared against the release's own SHA256SUMS.txt, the three files were checked against the 1.6.0 manifest shape, and the ProductCode was read from the Inno Setup script in the source repository rather than guessed. Happy to correct anything the validation pipeline flags.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/431082)

### DIFF
--- a/manifests/t/TechyGeeksHome/TranscribeGeek/1.1.2/TechyGeeksHome.TranscribeGeek.installer.yaml
+++ b/manifests/t/TechyGeeksHome/TranscribeGeek/1.1.2/TechyGeeksHome.TranscribeGeek.installer.yaml
@@ -1,0 +1,18 @@
+# Created with the TechyGeeksHome release process
+PackageIdentifier: TechyGeeksHome.TranscribeGeek
+PackageVersion: 1.1.2
+InstallerType: inno
+InstallModes:
+- interactive
+- silent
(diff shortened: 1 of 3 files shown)


Trenly:
[Policy] CLA Signed

PR #431189, opened 2026-09-08
New package: mtai0524.Cowork version 1.0.0

mtai0524:
### Package

`mtai0524.Cowork` 1.0.0 — https://github.com/mtai0524/cowork-launcher

A Windows desktop app for the programs you have to run every day: declare each one once with its
arguments, working directory and environment, then launch it by hand or let it run on a schedule or
on a machine event. It also edits those programs' JSON / INI / .env / XML config files in place with
backups and diffs, keeps services alive, retries failed jobs, and writes a log for every run.

### Checklist

- [x] Manifest validated locally — `winget validate --manifest .` reports **Manifest validation succeeded** with no warnings.
- [x] Conforms to the 1.6 manifest schema.
- [x] No other open pull request for this package. It is a new publisher: `manifests/m/mtai0524` does not exist in this repo yet.
- [ ] **Not** installed via `winget install --manifest`, which needs the `LocalManifestFiles` setting and elevation. Verified another way instead: the MSI was downloaded from the exact `InstallerUrl` in the manifest, its SHA-256 compared against `InstallerSha256`, installed with `msiexec /i /qn`, launched, and uninstalled — leaving no files, no Start Menu entry and no PATH entry behind. The `ProductCode` in the manifest is the one the install actually registers, read out of the published MSI rather than copied from a local build.

### Notes for review

- **Per-user install** (`Scope: user`). The MSI is built `Scope="perUser"` and installs to `%LOCALAPPDATA%\Programs\Cowork`, so no elevation is required.
- **Self-contained**, so there is no .NET runtime dependency to declare — that is why the installer is ~54 MB.
- The installer is served from the project's own GitHub release: https://github.com/mtai0524/cowork-launcher/releases/tag/v1.0.0
- `SHA256SUMS` is published alongside it on that release if you want an independent check.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/431189)

### DIFF
--- a/manifests/m/mtai0524/Cowork/1.0.0/mtai0524.Cowork.installer.yaml
+++ b/manifests/m/mtai0524/Cowork/1.0.0/mtai0524.Cowork.installer.yaml
@@ -1,0 +1,25 @@
+﻿# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.6.0.schema.json
+
+PackageIdentifier: mtai0524.Cowork
+PackageVersion: 1.0.0
+MinimumOSVersion: 10.0.17763.0
+InstallerType: wix
+Scope: user
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+UpgradeBehavior: install
+ProductCode: '{DC300C3E-5640-4C1B-A2A2-FEC770B511F1}'
+ReleaseDate: 2026-09-08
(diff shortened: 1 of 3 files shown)


mtai0524:

PR #431910, opened 2026-09-09
Fix stale installer URLs for OpenWhispr.OpenWhispr (1.7.5 - 1.9.2)

boseq:
All five of these manifests point at GitHub's `latest` release alias combined with a version-pinned filename:

```
https://github.com/OpenWhispr/openwhispr/releases/latest/download/OpenWhispr-Setup-1.9.0.exe
```

`/releases/latest/download/` resolves to whatever the newest release happens to be, so each manifest returns 404 as soon as the publisher ships the next version. Every version from 1.7.5 onward is affected; 1.7.3 already uses the correct pinned form and is untouched here.

This has been reported downstream twice: OpenWhispr/openwhispr#1909 and OpenWhispr/openwhispr#1355.

## Change

One line per manifest, `InstallerUrl` only:

```diff
-  InstallerUrl: https://github.com/OpenWhispr/openwhispr/releases/latest/download/OpenWhispr-Setup-1.9.0.exe
+  InstallerUrl: https://github.com/OpenWhispr/openwhispr/releases/download/v1.9.0/OpenWhispr-Setup-1.9.0.exe
```

Nothing else is modified — no `InstallerSha256`, `ProductCode`, `ReleaseDate` or `ManifestVersion` changes, and the locale/version manifests are untouched (they contain no release URLs).

## Verification

The installer binary is unchanged; only its address is corrected. Each existing `InstallerSha256` already matches the asset at the pinned URL, confirmed against the GitHub releases API `digest` field:

| Version | Pinned URL | Manifest SHA256 == asset digest |
|---|---|---|
| 1.7.5 | 206 | yes |
| 1.8.3 | 206 | yes |
| 1.9.0 | 206 | yes |
| 1.9.1 | 206 | yes |
| 1.9.2 | 206 | yes |

Note that 1.9.2 currently resolves only because it is coincidentally the newest release; it will 404 the moment the next version ships, which is why it is included.

I do not have a Windows machine available, so these have not been run through `winget validate` or an end-to-end `winget install --manifest` locally. Happy to hold for the validation pipeline, and to make any changes moderators prefer.

## Upstream follow-up

The publisher is adding a release workflow that submits future versions via WinGet Releaser, so the URLs will be tag-pinned from the release assets going forward: OpenWhispr/openwhispr#2089.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/431910)

### DIFF
--- a/manifests/o/OpenWhispr/OpenWhispr/1.7.5/OpenWhispr.OpenWhispr.installer.yaml
+++ b/manifests/o/OpenWhispr/OpenWhispr/1.7.5/OpenWhispr.OpenWhispr.installer.yaml
@@ -15,7 +15,7 @@ AppsAndFeaturesEntries:
   ProductCode: 69b301d3-943a-521b-b7b8-a2661047b2d2
 Installers:
 - Architecture: x64
-  InstallerUrl: https://github.com/OpenWhispr/openwhispr/releases/latest/download/OpenWhispr-Setup-1.7.5.exe
+  InstallerUrl: https://github.com/OpenWhispr/openwhispr/releases/download/v1.7.5/OpenWhispr-Setup-1.7.5.exe
   InstallerSha256: D46FD21920AD83A8A4BFC71431C01049CF3F2BB7FC7657D6C0A718D3C268DAEA
 ManifestType: installer
 ManifestVersion: 1.12.0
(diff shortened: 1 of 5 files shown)


boseq:

boseq:
@microsoft-github-policy-service agree company="Gizmo Labs Inc."

boseq:
Closing this in favour of one pull request per version, as the validation pipeline requires.

- 1.9.2: #431918
- 1.9.1: #431919
- 1.9.0: #431920
- 1.8.3: #431921

The change itself is unchanged in each of them, a single corrected `InstallerUrl` line.

PR #431943, opened 2026-09-09
New version: derekwisong.datui version 0.3.1

derekwisong:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/431943)


derekwisong:
Both flags should be clear now.

The installers scan passed on revalidation, and the installer hash matches the release asset exactly, so the original Defender hit looks like a false positive on an unsigned Rust binary.

I also pushed the missing `ReleaseNotes` and `ReleaseNotesUrl`. Our release body was written after the manifest was generated; fixed on our side so it cannot recur.

PR #428578, opened 2026-09-03
New version: Flowscripter.template-bun-executable version 1.3.23

vectronic:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/428578)

PR #431980, opened 2026-09-09
New version: Flowscripter.template-bun-executable version 1.3.25

vectronic:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/431980)

### DIFF
--- a/manifests/f/Flowscripter/template-bun-executable/1.3.25/Flowscripter.template-bun-executable.installer.yaml
+++ b/manifests/f/Flowscripter/template-bun-executable/1.3.25/Flowscripter.template-bun-executable.installer.yaml
@@ -11,7 +11,6 @@ InstallModes:
 - silent
 UpgradeBehavior: install
 ReleaseDate: 2026-09-09
-ElevationRequirement: elevationProhibited
 Installers:
 - Architecture: x64
   InstallerUrl: https://github.com/flowscripter/template-bun-executable/releases/download/v1.3.25/template-bun-executable_Windows_x64.zip
--- a/manifests/f/Flowscripter/template-bun-executable/1.3.25/Flowscripter.template-bun-executable.installer.yaml
+++ b/manifests/f/Flowscripter/template-bun-executable/1.3.25/Flowscripter.template-bun-executable.installer.yaml
@@ -11,6 +11,7 @@ InstallModes:
 - silent
 UpgradeBehavior: install
 ReleaseDate: 2026-09-09
+ElevationRequirement: elevatesSelf
 Installers:
 - Architecture: x64
   InstallerUrl: https://github.com/flowscripter/template-bun-executable/releases/download/v1.3.25/template-bun-executable_Windows_x64.zip
--- a/manifests/f/Flowscripter/template-bun-executable/1.3.25/Flowscripter.template-bun-executable.installer.yaml
+++ b/manifests/f/Flowscripter/template-bun-executable/1.3.25/Flowscripter.template-bun-executable.installer.yaml
@@ -11,7 +11,7 @@ InstallModes:
 - silent
 UpgradeBehavior: install
 ReleaseDate: 2026-09-09
-ElevationRequirement: elevatesSelf
+ElevationRequirement: elevationProhibited
 Installers:
 - Architecture: x64
   InstallerUrl: https://github.com/flowscripter/template-bun-executable/releases/download/v1.3.25/template-bun-executable_Windows_x64.zip
--- a/manifests/f/Flowscripter/template-bun-executable/1.3.25/Flowscripter.template-bun-executable.installer.yaml
+++ b/manifests/f/Flowscripter/template-bun-executable/1.3.25/Flowscripter.template-bun-executable.installer.yaml
@@ -1,0 +1,26 @@
+# Created with WinGet Releaser using komac v2.16.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: Flowscripter.template-bun-executable
+PackageVersion: 1.3.25
+InstallerType: zip
+NestedInstallerType: portable
+NestedInstallerFiles:
+- RelativeFilePath: template-bun-executable.exe
+InstallModes:
+- silent
+UpgradeBehavior: install
+Dependencies:
+  PackageDependencies:
+  - PackageIdentifier: Microsoft.VCRedist.2015+.x64
+ReleaseDate: 2026-09-09
+ElevationRequirement: elevationProhibited
+Installers:
+- Architecture: x64
+  InstallerUrl: https://github.com/flowscripter/template-bun-executable/releases/download/v1.3.25/template-bun-executable_Windows_x64.zip
+  InstallerSha256: 4CED6656DE3D44FA6B8D94D61DB08226F92B9F805CF482A33D0C662DEFF1F448
+- Architecture: arm64
+  InstallerUrl: https://github.com/flowscripter/template-bun-executable/releases/download/v1.3.25/template-bun-executable_Windows_arm64.zip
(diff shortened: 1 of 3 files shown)


vectronic:
The root-level `Dependencies` was intentionally moved to per-`Installers` `Dependencies` (not removed) — the previous manifest declared `Microsoft.VCRedist.2015+.x64` at the package level, which was being incorrectly applied to the arm64 installer too. This version scopes `Microsoft.VCRedist.2015+.x64` to the x64 installer and adds `Microsoft.VCRedist.2015+.arm64` to the arm64 installer, per the WinGet manifest schema's support for per-installer dependency overrides. Net effect: both architectures still declare a VCRedist dependency, just the correct one for each.

vectronic:
This is the third consecutive version (after 1.3.23 / #428578 and 1.3.24 / #430693, both closed as duplicates of this) to fail step 08 "Installation Validation" with no detail exposed to the contributor beyond "there was an issue with installing the application correctly" — same generic message each time. The prior version, 1.3.22 (#397492), merged cleanly on 2026-07-03 with no relevant change since then to this package's manifest generation or installer.

I've manually validated this exact manifest end-to-end on a real Windows 11 arm64 machine and it installs and runs correctly:

```
winget validate --manifest ...\template-bun-executable\1.3.25
winget install --manifest ...\template-bun-executable\1.3.25
```

Both dependency resolution (correctly pulls Microsoft.VCRedist.2015+.arm64 for the arm64 installer) and the installed executable itself work as expected.

Could a maintainer share the actual sandbox failure detail from the "08. Installation Validation" run? Given three consecutive versions failing identically with a clean local install, I suspect this is either an issue in the validation pipeline itself rather than the package, or something about this manifest shape (portable zip, dual x64/arm64 installers) that the automated sandbox can't handle — but without logs I can't narrow it further.

DuckDuckStudio:
You can download the validation log here:
https://github.com/microsoft/winget-pkgs/pull/431980/checks?check_run_id=102518906049
image

But I'm lazy, so I use [Sundry](https://github.com/DuckDuckStudio/Sundry):
```log
2026-09-09 15:16:26.038 <E> [CLI ] The installer cannot be run from an administrator context.
2026-09-09 15:16:26.060 <E> [CLI ] Terminating context: 0x8a150056 at C:\__w\1\s\external\pkg\src\AppInstallerCLICore\Workflows\InstallFlow.cpp:2ad

Analysis 以非正常退出代码退出
Installation failed with exit code -1978335146 in arm64-Scope_Undefined-Locale_Undefined-Log_InstallationClient_202609091515.log
Hint 此错误代码或许代表:
Hint Hex: 8A150056 | Dec: -1978335146 | InvDec: 2316632150 | Symbol: APPINSTALLER_CLI_ERROR_INSTALLER_PROHIBITS_ELEVATION | Description: The installer cannot be run from an administrator context. Does this package require `ElevationRequirement: elevationProhibited`?.

Analysis 以非正常退出代码退出
Installation failed with exit code -1978335146 in x64-Scope_Undefined-Locale_Undefined-Log_InstallationClient_202609091509.log
Hint 此错误代码或许代表:
Hint Hex: 8A150056 | Dec: -1978335146 | InvDec: 2316632150 | Symbol: APPINSTALLER_CLI_ERROR_INSTALLER_PROHIBITS_ELEVATION | Description: The installer cannot be run from an administrator context. Does this package require `ElevationRequirement: elevationProhibited`?.
```

vectronic:
Traced this further through winget-cli's source, and found something that doesn't add up.

First: `--dependencies-only` doesn't actually skip the elevation check on our own package's installer. In `InstallFlow.cpp`, `EnsureApplicableInstaller` runs `EnsureSupportForInstall` (which contains the `ElevationRequirement == ElevationProhibited` check) during installer selection/applicability validation - this happens *before* `InstallSinglePackage`'s own chain, where `CheckForOnlyDependencies` (the actual `--dependencies-only` early-exit) only runs later:

```cpp
Workflow::InstallDependencies <<
CheckForOnlyDependencies <<   // too late - EnsureApplicableInstaller already ran
Workflow::DownloadInstaller <<
Workflow::InstallPackageInstaller <<
```

So the elevation check on our own package is unavoidable during a `--dependencies-only` run, regardless of what `ElevationRequirement` we submit - as long as it isn't `elevationProhibited`.

Which is where this gets confusing. We've now tried two different values on this exact PR, with identical results both times:

1. Omitting `ElevationRequirement` entirely → sandbox failed with the same "cannot be run from an administrator context" error. `wingetvalidator-prod` flagged "Missing property `ElevationRequirement`" against it.
2. Setting it explicitly to `elevatesSelf` (matching what `Microsoft.VCRedist.2015+` itself declares) → sandbox **still failed identically**, at 2026-09-10 14:41 UTC, twenty minutes after the commit landed. This time `wingetvalidator-prod`'s "Inconsistencies detected" comment *stopped* flagging `ElevationRequirement` as missing - confirming it read our new value correctly.

Since `elevatesSelf` isn't `elevationProhibited`, the `EnsureSupportForInstall` check in winget-cli's own source shouldn't fire for that value at all. But "08. Installation Validation" produced the exact same failure signature both times regardless.

That combination - the lightweight consistency-checker correctly picking up each new commit, but the sandbox install test returning an identical result across two genuinely different manifest values - makes me wonder whether "08. Installation Validation" might be using a cached/stale manifest snapshot for this PR rather than re-resolving it on each push. Could someone confirm whether that job re-fetches the manifest fresh each run, or is there something else about a `zip`/`portable` installer plus `--dependencies-only` that I'm missing? Happy to keep iterating on the manifest if there's a concrete reason to believe a different value would help, but two attempts with opposite field values producing byte-identical failures is hard to explain as a manifest problem on our end.

stephengillie:
> 
> Installation failed with exit code -1978335146
> Unable to install manifest dependencies, aborting... 
 
  | 8A150056 | -1978335146 | 2316632150 | APPINSTALLER_CLI_ERROR_INSTALLER_PROHIBITS_ELEVATION | The installer cannot be run from an administrator context. Does this package require `ElevationRequirement: elevationProhibited`?. | 


vectronic:
@stephengillie two questions:

> Inconsistencies detected in package Flowscripter.template-bun-executable version 1.3.25 based on published version 1.3.22
> Missing property Dependencies

=> This is because I am now correctly specifying Dependencies per architecture. Is this causing the failure or is it just a warning? I _think_ the latter...

As far as I can tell the "Installation Validation" step is failing due to the need to install Microsoft.VCRedist.2015+ as Administrator.

=> Regardless of what I set for `ElevationRequirement` I am unable to get install to work in this pipeline. I have tried without `ElevationRequirement` declared at all, with `ElevationRequirement: elevationProhibited`, with `ElevationRequirement: elevatesSelf` and with `ElevationRequirement: elevationRequired`. Any suggestions?

vectronic:
@stephengillie 

1.3.22 was validated by WinGetSvc-Validation-142 / ADO and passed installation validation.

The GitHub App production cutover occurred on 2026-08-10 20:10 UTC.

1.3.23 is the first version of this package submitted after that cutover and was the first version to fail validation.

Subsequent validation logs show that the validator installs the declared Microsoft.VCRedist.2015+ dependency elevated, after which WinGet attempts to install this package from the elevated context.

WinGet rejects the package with:

```
    0x8a150056
    APPINSTALLER_CLI_ERROR_INSTALLER_PROHIBITS_ELEVATION
```

Changing this package's ElevationRequirement does not resolve the problem.

Can WinGet validation install elevated dependencies and then execute a dependent elevationProhibited/elevation-neutral package in a fresh non-elevated process/context?

If not, can the validator either:

a) install dependencies without contaminating the package's execution elevation context, or

b) launch the dependent package from a fresh non-elevated token?

The package itself installs and runs successfully on real Windows 11 systems when the VC runtime dependency is installed separately.

PR #432525, opened 2026-09-10
New package: NoaSansH.Thermalyn version 1.1.1

NoaSansH:
### New package: `NoaSansH.Thermalyn` version 1.1.1

Thermalyn is an open-source (GPL-3.0-or-later) hardware monitor for Windows 11 that reads processor, graphics, memory and drive sensors. Repository: https://github.com/NoaSansH/Thermalyn

- Manifest validated locally with `winget validate` (winget v1.29.290): passed.
- `ProductCode` verified against the real uninstall registry key on a machine with 1.1.1 installed: `{8F2C1D74-4E63-4A18-9E2B-5C7A0D3F1B96}_is1`.
- `InstallerSha256` taken from `SHA256SUMS.txt` published with the release, and matches the asset served at the installer URL.
- The URL points at the version-stamped asset `Thermalyn-Setup-1.1.1.exe` rather than the release's rolling `Thermalyn-Setup.exe` alias, so it stays immutable.

Two things a reviewer may want to know up front:

**The installer is not code-signed.** The project is maintained by one person and has no certificate. Every release binary does carry a signed GitHub build provenance attestation, verifiable with `gh attestation verify Thermalyn-Setup-1.1.1.exe --repo NoaSansH/Thermalyn`, which proves it was built by the repository's workflow at that tag.

**The installer sets up a kernel driver.** Processor temperature and package power are not published by Windows and require reading the processor's MSRs. Thermalyn uses [PawnIO](https://pawnio.eu) for this — signed, open source, and the same driver LibreHardwareMonitor uses. It is optional: everything else works without it, and the uninstaller asks before removing it since other monitoring tools may depend on it.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/432525)

### DIFF
--- a/manifests/n/NoaSansH/Thermalyn/1.1.1/NoaSansH.Thermalyn.installer.yaml
+++ b/manifests/n/NoaSansH/Thermalyn/1.1.1/NoaSansH.Thermalyn.installer.yaml
@@ -1,0 +1,20 @@
+PackageIdentifier: NoaSansH.Thermalyn
+PackageVersion: 1.1.1
+InstallerType: inno
+Scope: machine
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+InstallerSwitches:
+  Silent: /VERYSILENT /SUPPRESSMSGBOXES /NORESTART
+  SilentWithProgress: /SILENT /SUPPRESSMSGBOXES /NORESTART
+UpgradeBehavior: install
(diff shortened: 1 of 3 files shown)


NoaSansH:
Superseded by #433232, which submits 1.3.0 and declares Microsoft.DotNet.DesktopRuntime.8 as a dependency so the installer no longer prompts for the runtime during an unattended install.

PR #432796, opened 2026-09-10
New package: jimman0I.Prune version 2.3.4

jimman0I:
## 📖 Description

New package: **jimman0I.Prune** version **2.3.4**.

Prune is an open-source (MIT) Windows uninstaller: it runs a program's own uninstaller, then finds what it left behind and quarantines it. Source: https://github.com/jimman0I/prune

Updated from 2.3.3 to 2.3.4 before review (a small fix release), so this PR now adds only the 2.3.4 manifest.

The installer is an electron-builder NSIS installer (assisted, not one-click), so it has both install modes. The manifest lists two installers for the same file: `Scope: user` with `/currentuser`, and `Scope: machine` with `/allusers`, which elevates itself. `ProductCode` is the uninstall key electron-builder derives from the app id (a UUID v5 of `com.jimman0i.prune`); I confirmed it by installing 2.3.4 and reading the key back.

## ✅ Checklist

## 📦 Manifest Checklist

- [ ] Tested manifest locally with `winget install --manifest <path>` (the installer itself was installed and uninstalled silently with `/S /currentuser`; winget's own install path was not run locally)


jimman0I:

PR #432799, opened 2026-09-10
Recompile manifest workflows with gh-aw v0.88.7

yao-msft:
## 📖 Description

Regenerates the Manifest Validation Diagnosis, Missing Dependency Assist, and WinGetBot PR Triage lockfiles with `gh-aw v0.88.7`. Their merged `v0.83.4` lockfiles are blocked by the upstream compatibility policy for GHSA-8h78-hpm7-29gg, causing activation to fail before the agent runs.

Also short-circuits each workflow deterministically when its pre-agent collector finds no supported evidence path. The collector writes a `noop` safe output before engine startup, so these cases consume no AI inference. Manifest Validation Diagnosis preserves its check-independent `Manifest-Singleton-Deprecated` path.

The generated Manifest Validation Diagnosis `conclusion` job gains `issues: write` under `v0.88.7`. This is compiler-generated toolchain behavior; the authored workflow continues to set both `report-failure-as-issue: false` and `report-incomplete.create-issue: false`. The generated lockfile is not edited manually.

Failed production run: https://github.com/microsoft/winget-pkgs/actions/runs/34454734318

Validation performed:
- `gh aw compile` for all three workflow sources
- `gh aw validate` for all three workflow sources
- Focused predicate checks for Check-backed, singleton, unavailable, and stale evidence paths
- `git diff --check`

## ✅ Checklist

  - Not applicable; linked the failed production run above.

## 📦 Manifest Checklist

Not applicable; this PR changes only Agentic Workflow source/generated files and `.gitattributes`.


ranm-msft (commented):
Checked this against the three workflows it recompiles rather than reading it as a pure version bump, since mechanical recompiles are exactly where token scope drifts without anyone looking. Most of it is what you would expect from v0.83.4 to v0.88.7: pinned action SHAs move together, OTEL artifacts get collected, and several steps gain `if: always()` or `success() || failure()` so diagnostics survive a failed agent run.

Two permission changes ride along that are worth confirming are intended toolchain behaviour rather than compiler regressions:

- `manifest-validation-diagnosis.lock.yml` - the `conclusion` job goes from `actions: write` to `actions: write` plus **`issues: write`**. This is the only one of the three that gains it, which suggests it is driven by that workflow's safe-outputs configuration, but a conclusion job that can open issues is a real scope increase and worth a sentence in the PR description.
- `missing-dependency-assist.lock.yml` and `wingetbot-pr-triage.lock.yml` - `contents: read` is dropped and `actions: read` added. Net tightening, so no concern, just noting it was a deliberate-looking change and not noise.

Assuming both are expected, this is fine by me. Given the six new workflow PRs all touch `actions-lock.json` and `.gitattributes` identically, it is probably cleanest to land this one first and rebase the rest onto it.

yao-msft:
Updated in `f3ee334fad4`: all three pre-agent collectors now emit deterministic `noop` safe outputs when no supported evidence path remains, preventing AI inference. The manifest diagnosis gate preserves the check-independent singleton path. I also documented why the generated `conclusion` job gains `issues: write`: it is emitted by gh-aw v0.88.7, while the source still disables both failure and incomplete issue creation.

ranm-msft (commented):

**As compiled, though, `the new `Skip agent ...` step` will not write that noop - it fails the job instead.**

```
```

Every generated consumer re-maps it into its own step `env:` (`Start MCP Gateway`, `Execute GitHub Copilot CLI`, `Copy Safe Outputs`, `Ingest agent output`, `Parse agent logs`), and the only `$GITHUB_ENV` write in the whole file is `RUNNER_TOOL_CACHE`. The new step declares no `env:`, so `process.env.GH_AW_SAFE_OUTPUTS` is empty and the script takes its own guard branch - `core.setFailed("Safe outputs path is unavailable.")`. Every activated run where the collector finds no supported evidence path then ends as a failed agent job rather than a quiet no-op, which is the opposite of the intent on the highest-volume path.

```yaml
env:
```

gh-aw's own `squad-game-planner` pre-agent step instead defaults it in shell (`${GH_AW_SAFE_OUTPUTS:-${RUNNER_TEMP:-/tmp}/gh-aw/safeoutputs/outputs.jsonl}`); either works, but any fallback needs to be `RUNNER_TEMP`-based rather than `/tmp`, or the harness will not read the file it wrote. All three recompiled workflows have it. Separately, thanks for documenting the `issues: write` on the `conclusion` job - that answers my earlier question.

yao-msft:

## What changed

```yaml
env:
```

```js
const safeOutputsPath =
```

All three workflows in this PR are updated (head `b65425970db`).

ranm-msft (approved):
Verified, and this is exactly what I asked for. `id: set-runtime-paths` is declared ahead of each new gate step in all three lockfiles, so `${{ steps.set-runtime-paths.outputs.GH_AW_SAFE_OUTPUTS }}` resolves against a completed earlier step in the same job. Each gate now appends the noop to the same file the harness reads instead of taking the `core.setFailed("Safe outputs path is unavailable.")` branch, and the secondary default is `RUNNER_TEMP`-based as needed. The `.md` and `.lock.yml` are consistent with a clean v0.88.7 recompile, and the identical change is present in all nine workflow files across this batch and the six new-workflow PRs.

Approving.

PR #432902, opened 2026-09-10
New package: KallistoX.replaycut version 3.7.1

KallistoX:
﻿### New package: `KallistoX.replaycut` version 3.7.1

replaycut is a clip manager for the OBS replay buffer: a self-hosted service that watches the folder OBS writes replays to, serves one page in the browser for trimming, encodes the selected range with ffmpeg and can upload it and post the link.

- Homepage: https://replaycut.de
- Repository: https://github.com/KallistoX/replaycut
- Release: https://github.com/KallistoX/replaycut/releases/tag/v3.7.1
- License: AGPL-3.0-only

Notes on the manifest:

- The package is portable. The release ZIP carries `install.cmd`, but it is interactive (it asks about autostart and waits for a key), so it cannot be a nested installer for an unattended install. `winget install` puts `replaycut` on the PATH; users who want shortcuts and autostart run `replaycut install` once.
- `ArchiveBinariesDependOnPath: true` because `replaycut.exe` reads its UI from `ui\index.html` next to itself in the archive.
- `Gyan.FFmpeg` is declared as a dependency: replaycut encodes with ffmpeg and reads clips with ffprobe. Happy to drop the block if you would rather not have it.

This PR started on 3.7.0. Validation found that the executable did not start on a clean machine, which turned out to be a real bug in replaycut - see the comment below. 3.7.1 fixes it, and 3.7.0 has been dropped from this PR.

---

- [x] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/change?
- [x] Have you validated your manifest locally with `winget validate --manifest <path>`?
- [x] Have you tested your manifest locally with `winget install --manifest <path>`?
- [x] Does your manifest conform to the [1.12 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.12.0)?

PR #433100, opened 2026-09-11
New package: KallistoX.replaycut version 3.7.1

KallistoX:
### New package: `KallistoX.replaycut` version 3.7.1

replaycut is a clip manager for the OBS replay buffer: a self-hosted service that watches the folder OBS writes replays to, serves one page in the browser for trimming, encodes the selected range with ffmpeg and can upload it and post the link.

- Homepage: https://replaycut.de
- Repository: https://github.com/KallistoX/replaycut
- Release: https://github.com/KallistoX/replaycut/releases/tag/v3.7.1
- License: AGPL-3.0-only

This replaces #432902, which was for 3.7.1 as well but got closed by the bot while I was replacing the files in it - my mistake, the branch was briefly empty. This one starts clean.

**On the validation error in that PR** (thanks, @stephengillie): that exit code was a real bug, not a packaging problem. `replaycut.exe` imported `VCRUNTIME140.dll`, which is not part of Windows - it arrives with the Visual C++ Redistributable. Every game installs that one, which is why this never showed up on a developer or a gaming machine, and why a clean sandbox is the first thing that caught it. On a fresh Windows the program died before its first line, with no message at all. Fixed by linking the C runtime into the executable (`-C target-feature=+crt-static`); the only imports left are `api-ms-win-crt-*`, the UCRT, which has shipped with Windows since 10. 3.7.1 is that fix.

Notes on the manifest:

- The package is portable. The release ZIP carries `install.cmd`, but it is interactive (it asks about autostart and waits for a key), so it cannot be a nested installer for an unattended install. `winget install` puts `replaycut` on the PATH; users who want shortcuts and autostart run `replaycut install` once.
- `ArchiveBinariesDependOnPath: true` because `replaycut.exe` reads its UI from `ui\index.html` next to itself in the archive.
- `Gyan.FFmpeg` is declared as a dependency: replaycut encodes with ffmpeg and reads clips with ffprobe. Happy to drop the block if you would rather not have it.

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/433100)

### DIFF
--- a/manifests/k/KallistoX/replaycut/3.7.1/KallistoX.replaycut.installer.yaml
+++ b/manifests/k/KallistoX/replaycut/3.7.1/KallistoX.replaycut.installer.yaml
@@ -1,0 +1,26 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+# Portable on purpose, and `ArchiveBinariesDependOnPath` because the
+# executable reads its UI from `ui\index.html` next to itself. Both are
+# explained in dist/winget/README.md of the replaycut repository.
+
+PackageIdentifier: KallistoX.replaycut
+PackageVersion: 3.7.1
+MinimumOSVersion: 10.0.0.0
+InstallerType: zip
+NestedInstallerType: portable
+NestedInstallerFiles:
+  - RelativeFilePath: replaycut.exe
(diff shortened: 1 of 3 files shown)


KallistoX:

PR #433339, opened 2026-09-11
New version: io.github.frathe.picfetch version 1.1.0

frathe:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/433339)


Trenly:
@frathe - I'd suggest moving to the package identifier `frathe.picfetch` since the `io.github` portion of the identifier doesn't really provide any value. If you'd like my help moving the existing manifests to do that, just let me know
<!--
[Policy] Moderator Approved

frathe:
> Florian Rathe (@frathe) - I'd suggest moving to the package identifier `frathe.picfetch` since the `io.github` portion of the identifier doesn't really provide any value. If you'd like my help moving the existing manifests to do that, just let me know

Hi @Trenly 
and thanks for the advice.

Here is what I did so far: 
https://github.com/frathe/picfetch/commit/82baa368ea96e19a344f1ec4a73a5d84fc27dc6b

not sure if there is also something on your end that needs to change.
Thanks in advance for the help, and the good advice!

PR #433471, opened 2026-09-11
New package: EagleSpirit.Perch version 0.2.0

amir-khatibzadeh:
New package: **EagleSpirit.Perch** version **0.2.0**

Perch controls an IKEA IDASEN (Linak) sit/stand desk from Windows over Bluetooth LE:
exact heights, two presets, and an hourly stand schedule. The package also installs
`perch-cli` for scripts and automation. Open source under MIT:
https://github.com/eaglespirittech/perch

Details for review:

- Installer is Inno Setup, **per-user scope** (`PrivilegesRequired=lowest`), so it needs no elevation.
- `winget validate --manifest` passes locally against schema 1.12.0.
- `InstallerSha256` matches the SHA-256 published alongside the release asset.
- Silent install, Start Menu entry, Add/Remove Programs entry and a clean uninstall were all verified on Windows 11.
- `ProductCode` is the Inno uninstall key `{8E4B1C22-7F3A-4B6E-9C1D-2A5F0E7D3B91}_is1`, recorded under `HKCU` because the install is per user.

Happy to adjust anything that does not match house style.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/433471)

### DIFF
--- a/manifests/e/EagleSpirit/Perch/0.2.0/EagleSpirit.Perch.installer.yaml
+++ b/manifests/e/EagleSpirit/Perch/0.2.0/EagleSpirit.Perch.installer.yaml
@@ -1,0 +1,25 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: EagleSpirit.Perch
+PackageVersion: 0.2.0
+InstallerType: inno
+Scope: user
+InstallModes:
+- interactive
+- silent
+- silentWithProgress
+UpgradeBehavior: install
+ReleaseDate: 2026-09-10
+Installers:
+- Architecture: x64
+  InstallerUrl: https://github.com/eaglespirittech/perch/releases/download/v0.2.0/Perch-0.2.0-setup.exe
+  InstallerSha256: 40307B2F406DBBA50E06D23DDB7AE651669DF9BE0FB6EEA76D8953F85C7F3088
+  ProductCode: '{8E4B1C22-7F3A-4B6E-9C1D-2A5F0E7D3B91}_is1'
+  AppsAndFeaturesEntries:
+  - DisplayName: Perch
+    Publisher: Eagle Spirit
+    DisplayVersion: 0.2.0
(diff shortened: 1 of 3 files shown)


amir-khatibzadeh:
@microsoft-github-policy-service agree [company="Eagle Spirit Technologies Inc"]

amir-khatibzadeh:

amir-khatibzadeh:

amir-khatibzadeh:
Happy to answer any questions for the policy review. Some context that may save time:

**What it does.** Perch connects to a single Bluetooth LE device — a Linak desk controller, the one inside an IKEA IDASEN desk — and reads or sets its height. It makes no network requests of any kind: no telemetry, no update checks, no outbound connections at all.

**Persistence.** "Start with Windows" is a toggle in the app's own menu, off by default. When switched on it writes a single value to `HKCU\Software\Microsoft\Windows\CurrentVersion\Run`, so it is visible and revocable in Task Manager's Startup tab. The installer removes that value on uninstall, so it cannot leave a startup entry behind. The app also sits in the notification area, with Exit on its tray menu, because its optional hourly schedule only runs while the app does.

**The bundled CLI.** `perch-cli.exe` talks to the running app over a named pipe that is ACL'd to the current user's SID. There is no network listener and no remote surface; the pipe exists only because the desk accepts one Bluetooth connection at a time, so the CLI asks the app rather than fighting it for the radio.

**Installer.** Inno Setup, per-user scope, no elevation requested or required. Everything it writes is under `%LOCALAPPDATA%\Programs\Perch` and `%APPDATA%\Perch`.

**Provenance.** The project is MIT licensed and fully public at https://github.com/eaglespirittech/perch. The installer in this manifest is built by GitHub Actions from the tagged commit ([release workflow](https://github.com/eaglespirittech/perch/blob/main/.github/workflows/release.yml)), and its SHA-256 is published alongside it on the release page.

The binaries are not yet code signed, which I suspect is part of what drew attention here. That is on the roadmap.

PR #433549, opened 2026-09-12
New package: julianmb.OpenWinSidecar version 0.1.1

julianmb:
- [x] Have you signed the [Contributor License Agreement (CLA)](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)? (If not, the cla-bot will prompt after this PR opens.)
- [x] Have you checked that the package doesn't exist in the repository already?
- [x] Have you checked that the latest package version isn't in the repository already?
- [x] Have you validated the manifest locally with `winget validate --manifest <path>`?
- [x] Have you tested that the package installs successfully with the installer's silent switch (`/SILENT`, the switch winget sends)?
- [x] Does your manifest follow the [manifest authoring instructions](https://github.com/microsoft/winget-pkgs/blob/master/AUTHORING_MANIFESTS.md)?

### Description

New package `julianmb.OpenWinSidecar` version 0.1.1 — an open-source (AGPL-3.0) virtual-monitor server that turns an iPad or any modern browser into a wireless second Windows screen: IddCx virtual display, hardware HEVC end-to-end (Intel Quick Sync → WebCodecs), touch/keyboard input, no client app.

- Inno Setup installer, machine scope, silent switch `/SILENT`
- The HEVC encoder shells out to FFmpeg, so the manifest declares `Gyan.FFmpeg` under `PackageDependencies`; the installer additionally detects and installs FFmpeg via winget when run standalone
- Installer verified with live elevated install/uninstall cycles (driver registration, firewall rule, clean uninstall); `winget validate` passes on all three manifests
- Release asset: https://github.com/julianmb/OpenWinSidecar/releases/download/v0.1.1/OpenWinSidecar-Setup-0.1.1.exe (SHA-256 recorded in the installer manifest)

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/433549)

### DIFF
--- a/manifests/j/julianmb/OpenWinSidecar/0.1.0/julianmb.OpenWinSidecar.installer.yaml
+++ b/manifests/j/julianmb/OpenWinSidecar/0.1.0/julianmb.OpenWinSidecar.installer.yaml
@@ -1,0 +1,16 @@
+PackageIdentifier: julianmb.OpenWinSidecar
+PackageVersion: 0.1.0
+InstallerType: inno
+Scope: machine
+UpgradeBehavior: install
+Dependencies:
+  PackageDependencies:
+  - PackageIdentifier: Gyan.FFmpeg
+Installers:
+- Architecture: x64
+  InstallerUrl: https://github.com/julianmb/OpenWinSidecar/releases/download/v0.1/OpenWinSidecar-Setup-0.1.0.exe
+  InstallerSha256: 9a33a345c471a10279b98cc9f54719fa2dbc8079f30e40ea98fcd205a5215941
+  InstallerSwitches:
+    Silent: /SILENT
+ManifestType: installer
+ManifestVersion: 1.6.0
--- a/manifests/j/julianmb/OpenWinSidecar/0.1.0/julianmb.OpenWinSidecar.locale.en-US.yaml
+++ b/manifests/j/julianmb/OpenWinSidecar/0.1.0/julianmb.OpenWinSidecar.locale.en-US.yaml
@@ -1,0 +1,21 @@
+PackageIdentifier: julianmb.OpenWinSidecar
+PackageVersion: 0.1.0
+PackageLocale: en-US
+Publisher: julianmb
+PackageName: OpenWinSidecar
+License: AGPL-3.0-only
+LicenseUrl: https://github.com/julianmb/OpenWinSidecar/blob/main/LICENSE
+Copyright: Copyright (c) 2026 OpenWinSidecar Contributors
+ShortDescription: Use your iPad as a wireless second Windows screen over Wi-Fi.
+Description: OpenWinSidecar turns an iPad into a real extra Windows monitor. A virtual
+  display driver extends the desktop, Intel Quick Sync encodes it to HEVC, and Safari
+  shows it with full touch and keyboard input. No client app, no subscription.
+PackageUrl: https://github.com/julianmb/OpenWinSidecar
+Tags:
+- ipad
+- second-monitor
+- virtual-display
+- hevc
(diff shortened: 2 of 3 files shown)


julianmb:

julianmb:
Root cause of the round-1 installation-validation failure ("failed to install without user input") is fixed:

- The installer shows an FFmpeg-missing prompt when the winget fallback is unavailable or fails. On a headless validation VM with no FFmpeg, that MsgBox blocked the silent install forever. Silent installs (`WizardSilent()`) now **log** instead of prompting, and also skip the in-installer winget attempt entirely — zero interactive paths. Winget-driven installs get FFmpeg from the manifest `PackageDependencies`; other silent deployments get the actionable log line and the app repeats it at runtime.
- The manifest now declares `SilentWithProgress: /SILENT` alongside `Silent:`.
- The release asset was replaced accordingly (same-day, pre-approval): new SHA-256 `96cb141c0816efd9ef6186f313c02f18aa39cdcc15c4517d309f0e85dbaf8833`, matching the updated installer manifest in this PR.

The rebuilt installer also fixes a ghost-monitor bug found on real machines (devcon node creation is now skipped when a virtual display instance already exists). Verified with silent install/uninstall cycles on Windows 11: fully non-interactive, idempotent, clean uninstall.

julianmb:
Round-2 failure root-caused from the check-run log: the harness ran the installer for **2h04m** before its timeout killed it. Our installer completes in ~1 minute on real hardware — the hang is `devcon install`, which waits for the device install to finish; on a VM whose display adapter cannot start an IddCx driver, that wait never ends.

Fix in this rebuild:

- In silent installs (`WizardSilent()`), the driver operations (pnputil staging, devcon device-node creation) run **fire-and-forget** — setup can never block on the driver subsystem and always terminates on its own. Interactive installs keep the waited behavior with status messages.
- Verified with a timed silent cycle on Windows 11: silent install wall time **12.3 s**, virtual display node created asynchronously ~2 s later.

Release asset replaced accordingly; new SHA-256 `6e2aa1506c43cdc2420877b6d1ef587ba97f047cfc0a981fb7aa94b247034beb` matches the updated installer manifest in this PR.

julianmb:
The round-3 hash mismatch is on us, and it is fixed properly this time by moving to a fresh release URL:

- Root cause: after the round-2 timeout fix we replaced the release asset **at the same URL** (delete + re-upload of `OpenWinSidecar-Setup-0.1.0.exe` under tag `v0.1`). GitHub serves release assets through a CDN, and the validation pipeline downloaded a **stale cached copy of the old binary** (SHA-256 `96cb141c…`), which no longer matched the manifest (`6e2aa150…`). Our own post-swap download had already returned the new binary — same-URL swaps are a trap, and we fell into it.
- Fix: we cut a proper new release instead of mutating the old one. This PR now submits **version 0.1.1** (`OpenWinSidecar-Setup-0.1.1.exe` at tag `v0.1.1`) — a brand-new URL with no cache history. The 0.1.0 manifest folder has been removed from this branch and replaced with the 0.1.1 manifests; the release-notes hash in the release itself matches the installer.
- Everything else from round 3 stands: silent installs run the driver operations fire-and-forget (the round-2 2-hour timeout was `devcon install` waiting on an IddCx driver that cannot start on a GPU-less VM), verified at 12.3 s wall time with the virtual display appearing asynchronously.

Apologies for the churn — three genuine installer bugs found by this validation pipeline (dialog on headless VMs, non-idempotent device-node creation, blocking driver ops). All fixed and verified locally.

PR #433575, opened 2026-09-12
New version: YuanDevLLC.OneXray version 26.9.2

yiguodev:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/433575)

### DIFF
--- a/manifests/y/YuanDevLLC/OneXray/26.9.2/YuanDevLLC.OneXray.installer.yaml
+++ b/manifests/y/YuanDevLLC/OneXray/26.9.2/YuanDevLLC.OneXray.installer.yaml
@@ -1,0 +1,31 @@
+# Created with WinGet Releaser using komac v2.16.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: YuanDevLLC.OneXray
+PackageVersion: 26.9.2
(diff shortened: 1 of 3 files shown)


yiguodev:
Updated the installer metadata to match the released per-user Inno Setup installers:

- Set `Scope: user` and `DefaultInstallLocation: '%LOCALAPPDATA%\Programs\OneXray'`.
- Added `AppsAndFeaturesEntries.DisplayName: OneXray version 26.9.2`.
- Removed `ElevationRequirement: elevatesSelf`: the installers use `PrivilegesRequired=lowest`.
- Kept `DisplayVersion` omitted because it would duplicate `PackageVersion`.

Installer URLs, SHA-256 values, and ProductCode are unchanged. No external VC++ runtime dependency was added: the 26.9.2 build bundles the runtime, and static inspection of both released ZIPs confirmed that the VC++ DLLs actually imported by each architecture's native binaries are present and architecture-compatible.

All three manifests passed the official 1.12.0 JSON schemas, duplicate-key checks, and formatting checks. Release asset metadata and the downloaded ZIP hashes were also checked. Windows installation and `winget validate` were not run on this macOS host; those checks are left to the PR's Windows validation pipeline.

stephengillie (approved):

PR #433788, opened 2026-09-12
New version: fstubner.netscli version 0.3.1

fstubner:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/433788)


fstubner:
Closing this in favour of 0.3.2.

The `Validation-Executable-Error` was ours, not a pipeline problem. `netscli`
with no arguments opens an interactive TUI, and with no terminal attached it
blocked instead of exiting, so a run of the executable never returned.
Confirmed against the submitted binary with stdin closed: still running when
killed at 15 seconds.

Fixed in 0.3.2 — with no terminal it prints help and exits 0. I will submit
that version once it is released. Closing this one rather than leaving it
open, since the 0.3.1 binary would fail the same check.

PR #433896, opened 2026-09-12
Add OpsMaxx.OpsMaxx version 0.36.11

ZeeshanSultan:
### Manifest for `OpsMaxx.OpsMaxx` version `0.36.11`

Publisher-submitted. I maintain OpsMaxx.

- [x] Have you checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/add?
- [x] Does your manifest conform to the [1.6 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.6.0)?
- [ ] Have you validated your manifest locally with `winget validate --manifest <path>`?
- [ ] Have you tested your manifest locally with `winget install --manifest <path>`?

**Being explicit about the two unchecked boxes rather than ticking them:** this manifest was authored on macOS, so `winget validate` and `winget install` were not run. Instead all three files were validated against the published v1.6.0 JSON schemas (`manifest.version`, `manifest.installer`, `manifest.defaultLocale`), and all three pass. If that is not sufficient here, say so and I will get a Windows machine on it rather than spend your review time.

An earlier revision of this description had those two boxes ticked. That was wrong and I have corrected it.

`InstallerSha256` was taken from the release's published checksum table and independently verified against the downloaded installer:

```
c008fc2f49c4981c000e4bea3df5f5708589bb70c3d05170b6f7daed9d1f2976  OpsMaxx-0.36.11-setup.exe
```

`InstallerType` is `nullsoft` (electron-builder's NSIS target) and `Scope` is `user`, matching `nsis.perMachine: false` in the build config. Previous versions of this package were accepted with the same shape.

Release: https://github.com/OpsMaxx/OpsMaxx/releases/tag/v0.36.11

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/433896)

PR #431908, opened 2026-09-09
New package: OpsMaxx.OpsMaxx version 0.36.0

ZeeshanSultan:
### Pull request has been created with the following manifest

- `manifests/o/OpsMaxx/OpsMaxx/0.30.1/OpsMaxx.OpsMaxx.yaml`
- `manifests/o/OpsMaxx/OpsMaxx/0.30.1/OpsMaxx.OpsMaxx.installer.yaml`
- `manifests/o/OpsMaxx/OpsMaxx/0.30.1/OpsMaxx.OpsMaxx.locale.en-US.yaml`

#### Checklist

- [x] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)? — the CLA bot's check is authoritative.
- [x] Have you checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/add?
- [ ] Have you validated your manifest locally with `winget validate --manifest <path>`? — **not run.** Prepared on macOS, where `winget` is unavailable. The three manifests were validated programmatically against the published v1.6.0 `manifest.installer`, `manifest.version` and `manifest.defaultLocale` JSON schemas, all passing.
- [ ] Have you tested your manifest locally with `winget install --manifest <path>`? — **not run**, same reason.
- [x] Does your manifest conform to the [1.6 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.6.0)?

#### Notes

Publisher submission. Upstream release: https://github.com/OpsMaxx/OpsMaxx/releases/tag/v0.30.1

`InstallerSha256` is `36BB3BA43C14BC6AD7ECDA8D2A3718D3F4389152B42313ED0098DAC44062AF40`, taken from the release notes' checksum table and independently verified against a fresh download of the published asset.

`InstallerType: nullsoft` is the NSIS installer produced by electron-builder, and `Scope: user` because it is a per-user install (`nsis.perMachine: false`).

Supersedes the 0.29.0 submission if that one has not merged yet — happy to close whichever is less convenient.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/431908)

PR #433909, opened 2026-09-12
Add OpsMaxx.OpsMaxx version 0.36.14

ZeeshanSultan:
### Manifest for `OpsMaxx.OpsMaxx` version `0.36.14`

Publisher-submitted. I maintain OpsMaxx.

**Updated in place from 0.36.12 to 0.36.14** rather than opening another pull request. I had let three submissions for this package accumulate; #431908 and #433896 are now closed and this is the only one open.

The last two boxes are unticked deliberately rather than left ambiguous: this manifest was authored on macOS, so neither `winget validate` nor `winget install` was run. What was run is schema validation of all three files against the published v1.6.0 JSON schemas (`manifest.version`, `manifest.installer`, `manifest.defaultLocale`) — all three pass. If that is not sufficient, say so and I will get a Windows machine on it rather than spend your review time.

`InstallerSha256` was computed from the downloaded installer and compared against the release's published checksum table — they match:

```
0e6672185b4fa3bc025e75014a5ea58ac382eaa897581581c152b8ff6944bca7  OpsMaxx-0.36.14-setup.exe
```

`InstallerType` is `nullsoft` (electron-builder's NSIS target) and `Scope` is `user`, matching `nsis.perMachine: false` in the build config. To be accurate about the previous wording here: this package has never been merged, so that shape has passed validation on earlier submissions but has not yet been accepted by a reviewer.

Release: https://github.com/OpsMaxx/OpsMaxx/releases/tag/v0.36.14


Trenly:
[Policy] CLA Signed

PR #433980, opened 2026-09-13
Add version: SosomLab.NexaCoffee version 0.1.0

kiros33:


kiros33:
Closing in favour of a new PR for 0.1.1.

The `Validation-Installation-Error` was reproduced locally: Microsoft Defender quarantined **only the x86 archive** as `Trojan:Win32/Tecabans.STV!cl` (cloud ML, "downloads and attachments" path) while the x64 archive installed fine — so the install step failed on the x86 installer, not on the manifest (`winget validate` passes).

The binaries are tiny, unsigned, CRT-less PEs that carried **no version resource at all**, a well-known false-positive trigger. 0.1.1 adds a full `VS_VERSION_INFO` block. Verified on the same machine and definitions (AV 1.459.241.0): the 0.1.0 x86 zip is still quarantined on write, the 0.1.1 x86 zip is not, and `winget install --manifest ... --architecture x86` against the published 0.1.1 asset now succeeds.

Sorry for the noise, and thanks for the validation output — it pointed straight at it.

PR #433981, opened 2026-09-13
Fix Manifest Validation Diagnosis comment permissions

yao-msft:
## 📖 Description

Fixes the Manifest Validation Diagnosis comment safe-job, which could not post its result. The workflow completed its analysis and then failed to publish it, so every diagnosis was lost.

Failed production run: https://github.com/microsoft/winget-pkgs/actions/runs/34744538124

`pre_activation`, `activation`, `agent`, and `detection` all succeeded. Only `post_pr_comment` failed, with `403 Resource not accessible by integration` on `POST /repos/microsoft/winget-pkgs/issues/433970/comments`. The safe-job had `issues: write` with `pull-requests: read`.

### Which permission the endpoint actually requires

The endpoint returns `x-accepted-github-permissions: issues=write; pull_requests=write`. Semicolons separate **alternative valid sets**, not permissions required together, [as documented in the changelog that introduced the header](https://github.blog/changelog/2023-08-10-x-accepted-github-permissions-header-for-fine-grained-permission-actors/). Which set applies is determined by the target resource, since issues and pull requests share this endpoint.

Because the observed behavior and the header alone cannot distinguish those readings, the requirement was established empirically. Four isolated jobs, each granted exactly one permission, ran against a real pull request:

| --- | --- | --- |

For a pull request target, `pull-requests: write` is both necessary and sufficient to post a comment, and `issues: write` does not authorize it at all. Reads are satisfied by either permission.

### Resulting change

The safe-job permissions become `checks: read`, `contents: read`, `pull-requests: write`. This both fixes the failure and **reduces** the granted scope: the previous `issues: write` was never able to authorize this call and has been removed rather than retained alongside the correct permission.

The safe-job is reached only from `pull_request_target` and always targets `github.event.pull_request.number`, so the target is always a pull request. Its only two GitHub API calls are `issues.createComment` and `issues.listComments`, both covered by `pull-requests: write` per the table above.

**Agent job permissions are unchanged and remain read-only.** The write scope belongs solely to the post-agent job that publishes the already-generated comment, so the agent gains no additional capability and the workflow remains recommend-only.

### Also included

Sets `report-failed-jobs: false`. The `conclusion` job emitted a `Report failed jobs` step that opened a repository issue for any failed job in a run. That is a distinct mechanism from `report-failure-as-issue`, which covers agent failure only, and it defaults to `true` when unset. Combined with the permission failure above, each affected run opened an `[aw] Failed jobs` issue.

Validation performed:
- `gh aw compile` and `gh aw validate` for the workflow source
- Empirical single-permission probe described above, run against a real pull request
- Confirmed the generated safe-job permissions are exactly `checks: read`, `contents: read`, `pull-requests: write`
- Confirmed top-level agent permissions still resolve to `pull-requests: read`
- Confirmed the `Report failed jobs` step and `GH_AW_REPORT_FAILED_JOBS` no longer appear in the generated lockfile, while `GH_AW_FAILURE_REPORT_AS_ISSUE: "false"` is retained
- Confirmed `body_hash` is unchanged across all commits, so no prompt, classifier, trigger, concurrency, network, or safe-output behavior changed
- `git diff --check`

The generated lockfile is produced by `gh aw compile` and is not edited manually.

## ✅ Checklist

  - Not applicable; linked the failed production run above.

## 📦 Manifest Checklist

Not applicable; this PR changes only Agentic Workflow source and generated files.


yao-msft:

```yaml
safe-outputs:
  report-failed-jobs: false
```

ranm-msft (commented):

Two things worth a reply:

yao-msft:
> The permission fix looks right, but the stated reason is inverted.

1. the pullrequest: write and issues: write are both needed for posting a comment 
2. these are run a lot of times, we don't want to create a separate issue for each run. It's a request from our moderator.

yao-msft:

| --- | --- | --- |

### Change applied

denelon (approved):

PR #434024, opened 2026-09-13
New package: UtopianAcademy.Deckboy version 0.99.357

Utopian-Academy:
### Pull request has been manually validated?

- [x] Manifest validated locally with `winget validate` (exit 0, no warnings)

### New package

`UtopianAcademy.Deckboy` 0.99.357 — Deckboy, an open-source (GPL-3.0) cue-based media playback and show-control application for live video. Windows installer is Inno Setup, x64.

- Homepage: https://utopian-academy.github.io/Deckboy/
- Source: https://github.com/Utopian-Academy/Deckboy
- Installer: signed? **No** — the project has no paid code-signing certificate, so SmartScreen will warn. This is stated in the project README.

### Checklist

- [x] Have you checked that there aren't other open [pull requests](https://github.com/microsoft/winget-pkgs/pulls) for the same manifest update/addition?

Note on the unchecked box: the manifest was validated but not install-tested, because the only machine available already has Deckboy installed from source and a machine-scope install would have collided with it.

Disclosure: I am the author of this package.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/434024)


Utopian-Academy:

Utopian-Academy:
Following the documented step for `Validation-Executable-Error` ("if you believe this failed in error, add a comment to get the Windows Package Manager engineers to investigate") — and with a concrete guess attached, in case it saves a round trip.

Deckboy does install a primary application: `Deckboy.exe`, into `{autopf}\Deckboy` for a per-machine install. Checks 01-07, 09 and 10 all passed, the CLA is signed, and I have re-verified the installer hash against the release asset just now — `0A1BB86D...A7F5AD85`, matching the manifest.

My guess at the cause: this is an Inno Setup installer built with `PrivilegesRequiredOverridesAllowed=dialog commandline`, so it can install either per-machine or per-user. If the validation host runs the installer without elevation, Inno will complete a **per-user** install into `%LOCALAPPDATA%\Programs\Deckboy` and register under `HKCU`, while the manifest declares `Scope: machine` and a `ProductCode` that would then be looked for under `HKLM`. That would leave the test unable to locate the application even though the install succeeded. I can see exactly that shape of install on a local machine here.

If that is what happened, the fix is one line and I am happy to push it:

```yaml
InstallerSwitches:
  Custom: /ALLUSERS
```

which pins the silent install to per-machine and makes the declared scope and the actual install agree regardless of elevation.

I did not want to churn the branch and reset a clean validation run on a guess, so: if the log shows the install landing under `%LOCALAPPDATA%`, say the word and I will push the switch. If it shows something else, I would be glad of the relevant lines and will fix the actual cause instead.

Windows Sandbox is not available on this edition (Windows 11 Home), so I have not been able to run `SandboxTest.ps1` against it myself.

Utopian-Academy:
Updated this PR rather than opening a new one, since the package is the same and the branch already carries the discussion above.

**What changed**

1. **Version 0.99.357 → 0.99.367.** The older manifests are removed and replaced by `manifests/u/UtopianAcademy/Deckboy/0.99.367/`. `InstallerSha256` is `6CD51F90…03C01B9D`, computed from the installer downloaded from the release, not copied forward.

2. **The fix I proposed above, now applied** — no reply came, so rather than leave the PR waiting I have pushed it:

   ```yaml
   InstallerSwitches:
     Custom: /ALLUSERS
   ```

   Deckboy's Inno Setup installer is built with `PrivilegesRequiredOverridesAllowed=dialog commandline` and `DefaultDirName={autopf}`, so without elevation it completes a **per-user** install into `%LOCALAPPDATA%\Programs\Deckboy` and registers under `HKCU`, while the manifest declares `Scope: machine` with a ProductCode that would then be looked for under `HKLM`. That shape of mismatch matches the `Validation-Executable-Error` exactly: the install succeeds, and the check cannot find what it installed. `/ALLUSERS` pins the silent install to per-machine so the declared scope and the actual install agree regardless of elevation.

If the validation log shows a different cause, I would be glad of the relevant lines and will fix the actual one instead — I still cannot run `SandboxTest.ps1` here (Windows 11 Home has no Windows Sandbox).

PR #434095, opened 2026-09-13
Remove version: Lanreal.Whisperstream version 1.0.0

farazs:
## Summary
- Remove the `Lanreal.Whisperstream` version `1.0.0` manifests.
- Whisperstream is being sunset by its publisher and should no longer be available through WinGet.

## Validation
- Confirmed this change only deletes the three manifest files under `manifests/l/Lanreal/Whisperstream/1.0.0`.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/434095)


stephengillie:
Hi @,

farazs:
We are sunsetting the app and so would like to remove it from here

farazs:
actually we have decided to open source the app instead!

PR #434711, opened 2026-09-14
New package: DeskStitch.DeskStitch version 0.7.7

jonseoane:
### New package: `DeskStitch.DeskStitch` version `0.7.7`

DeskStitch is a software KVM for two Windows PCs: one keyboard and mouse drive both, and the cursor crosses between their screens. It places monitors by their physical size in millimetres rather than by pixels, so a 4K panel next to a 1440p panel of the same size lines up exactly.

- Homepage: https://deskstitch.com
- Release: https://github.com/jonseoane/deskstitch-releases/releases/tag/v0.7.7
- Publisher: Seoane Montero Consulting

#### Checklist

- [x] `winget validate --manifest` passes (client v1.29.290): *Manifest validation succeeded.*
- [x] Manifests also validate against the published 1.6.0 `version`, `installer` and `defaultLocale` JSON schemas.
- [x] Installer URLs resolve, and both `InstallerSha256` values were verified by downloading the installers and hashing them rather than copying from a build log.
- [x] Both installers are Authenticode signed: Azure Trusted Signing, subject `Seoane Montero Consulting`, issuer `Microsoft ID Verified CS AOC CA 04`, RFC 3161 timestamped.
- [x] `ProductCode` matches the Inno Setup `AppId` with the `_is1` suffix.
- [x] This is the publisher's own package, in use on Windows 11.

Happy to make any change the validation pipeline or a reviewer asks for.


jonseoane:

PR #435795, opened 2026-09-16
New version: Snowflake.CortexCode version 1.21.5

powerstacks-wingetbot:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/435795)


pjm-powerstacks:
This looks like a false positive. Both installers are served from sfc-repo.snowflakecomputing.com, Snowflake's own distribution host, on the same path as the published 1.21.3 manifest with only the version number changed.

Could you confirm which installer and which detection triggered this? I will submit it to the Microsoft Defender team for analysis and follow up here.

stephengillie (approved):

Issue #72224, opened 2022-08-26
Validation is running manifests with elevationProhibited in an elevated context


jedieaston:
As of winget 1.3, winget now supports a `ElevationRequirement` key for installers, which can tell the client to error out if a user tries to run an installer that will fail when ran as an administrator (https://github.com/microsoft/winget-cli/pull/2126). 

The validation service doesn't seem to respect this, and is running installers in an elevated context that have `ElevationRequirement: elevationProhibited` set. Here's an example using Spotify (https://github.com/microsoft/winget-pkgs/pull/72166): 

InstallationClient log:
```
Installing winget from 'C:\Winget\Dependencies'
winget-cli Installation complete.
WinGet is installed.
Running winget settings --enable LocalManifestFiles
End winget settings --enable LocalManifestFiles
Enabling installation of local manifests succeeded: True
WinGet is installed.
File Watcher -> Register watching C:\, Filter = 
Scope set to: 
Locale set to: 
System Architecture: X86
Running with Elevation: True <----------------------
Starting FileTracker
File Watcher -> Begin watching C:\, Filter = 
Running winget install --manifest D:\TRANSFER\c79454d6-4925-ed11-951f-000d3a9edf5a\Manifest.yaml --accept-package-agreements
End winget install --manifest D:\TRANSFER\c79454d6-4925-ed11-951f-000d3a9edf5a\Manifest.yaml --accept-package-agreements
Installation failed with exit code -1978335146
Stopping FileTracker
File Watcher -> Stop watching C:\, Filter = 
```
`winget install` log: 
```
2022-08-26 07:18:37.895 [CORE] WinGet, version [1.3.2091], activity [{E2DC2FEF-304F-444F-9844-056F3881F26C}]
2022-08-26 07:18:37.896 [CORE] OS: Windows.Desktop v10.0.19044.1889
2022-08-26 07:18:37.896 [CORE] Command line Args: C:\Users\Validator\AppData\Local\Microsoft\WindowsApps\winget.exe install --manifest D:\TRANSFER\c79454d6-4925-ed11-951f-000d3a9edf5a\Manifest.yaml --accept-package-agreements
2022-08-26 07:18:37.896 [CORE] Package: Microsoft.DesktopAppInstaller v1.18.2091.0
2022-08-26 07:18:37.896 [CORE] IsCOMCall:0; Caller: winget-cli
[...]

PR #435976, opened 2026-09-16
New package: Dualface.Ullage version 0.1.3

dualface:
This PR adds user-scope Inno WinGet manifests for ullage 0.1.3.

- Installer: GitHub Release `ullage-x86_64-pc-windows-setup.exe`
- Scope: user (`PrivilegesRequired=lowest`, install dir `%LOCALAPPDATA%\Ullage`)
- Silent [Run] entries call `ullage daemon stop`, `install`, and `start` (Homebrew `post_install` analog; no `postinstall` flag so winget silent mode still runs them)

Checksums: https://github.com/dualface/ullage-cli/releases/download/v0.1.3/checksums.txt
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/435976)


dualface:

dualface:
The second validation run still fails before the Ullage installer is launched.

Evidence from operation `WinGetSvc-Validation-435976-2`:

```
winget install --manifest ... --dependencies-only --dependency-source winget
The installer cannot be run from an administrator context.
0x8A150056 (APPINSTALLER_CLI_ERROR_INSTALLER_PROHIBITS_ELEVATION)
```

The merged manifest artifact for this run does not contain `ElevationRequirement`. The installer is intentionally per-user (`PrivilegesRequired=lowest`) because its post-install commands register a scheduled task for the installing user. Changing it to an elevated or machine-scope installer would register that task for the wrong account.

This appears to be the validation-service administrator-context issue previously tracked in #72224. Could a maintainer manually validate this package and waive `Validation-Installation-Error`? The failure occurs during dependency processing, before the Ullage installer executes.

dualface:
Follow-up: changing only `ElevationRequirement` did not fix validation because the published v0.1.3 Inno binary still exposes the old privilege metadata. The source fix now adds `PrivilegesRequiredOverridesAllowed=commandline`, while future WinGet manifests pass `/CURRENTUSER` to preserve the per-user install.

This requires a newly built release asset. The immutable v0.1.3 asset cannot be repaired in place, so this PR will be superseded by the next Ullage version.

dualface:
Superseded by #436014, which uses the v0.1.4 installer with explicit Inno privilege overrides and the `/CURRENTUSER` WinGet switch.

PR #436258, opened 2026-09-16
New package: chamal-code.pathdoc version 0.1.1

chamal-code:
New package: `chamal-code.pathdoc` version 0.1.1

Submitted by the package author.

### What it is

A read-only auditor for the Windows `PATH`. It reads the stored value from `HKLM` and
`HKCU`, composes it the way Windows resolves it (machine entries before user entries),
and reports dead directories, duplicates, stray separators, which copy of each
executable wins, and which names a shell intercepts before `PATH` is searched at all.

It never writes: no `PATH` edits, no registry writes, no fix mode.

Repository: https://github.com/chamal-code/pathdoc
Release: https://github.com/chamal-code/pathdoc/releases/tag/v0.1.1

### Manifest notes

- Portable package. `InstallerType: zip` with `NestedInstallerType: portable`, because
  the release archive places the executable inside a version-named top-level directory.
  `PortableCommandAlias` is `pathdoc`.
- Dual licensed, `MIT OR Apache-2.0`. Both licence texts are included inside the
  distributed archive alongside the executable.
- `InstallerSha256` was taken from the `SHA256SUMS.txt` published with the release, then
  confirmed against the downloaded artifact rather than computed locally.
- `ReleaseDate` is the UTC date from the release's `published_at`.
- Schema 1.12.0, authored by hand.

### Validation performed

- `winget validate --manifest` succeeds.
- `RelativeFilePath` checked against the actual entries of the published archive.
- Every URL in the manifests returns 200. The README anchor was verified by fetching the
  file at the tag and confirming the heading exists, since a HEAD request cannot see an
  anchor.
- The binary inside the pinned archive was extracted and run: it reports `pathdoc 0.1.1`
  and audits correctly.

### Note for reviewers

`pathdoc` exits **1** when it finds actionable entries, which is the normal result on a
machine nobody has tidied, and **2** only on a fatal error. Exit 1 is a successful run
that had something to report, not a failure. Installation is unaffected either way,
since a portable package is extracted rather than executed.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/436258)


chamal-code:

PR #436665, opened 2026-09-17
New package: Mr-hunt-007.portlist version 1.2

Mr-hunt-007:
### New package: `Mr-hunt-007.portlist` version 1.2

- **Package**: portlist — every port on this machine, and where it came from
- **Homepage**: https://mr-hunt-007.github.io/portlist/
- **Repository**: https://github.com/Mr-hunt-007/portlist
- **License**: MIT
- **Installer**: the source plus a `portlist.cmd` shim, published as a zip on the GitHub release for `v1.2`

#### Checklist

- [x] Does your manifest match the publisher's release? Yes — the installer URL points at the project's own GitHub release, and `InstallerSha256` was taken from the asset GitHub serves.
- [x] This PR is authored by the package's own maintainer.
- [ ] `winget validate` / `winget install` — see the note below.

#### On validation

This was prepared on macOS, so `winget validate` and `winget install` could not be run. In their place:

- All three manifests validate cleanly against Microsoft's published `1.6.0` JSON schemas (`installer`, `defaultLocale`, `version`).
- `InstallerSha256` (`b0e55441…`) was verified by downloading the release asset from GitHub and hashing it, not by trusting a local build.
- `NestedInstallerFiles.RelativeFilePath` (`portlist-1.2\portlist.cmd`) was checked against `unzip -l` of that downloaded asset, which contains `portlist-1.2/portlist.cmd`.

If CI or a reviewer finds anything the schemas do not cover, I will fix it promptly.

#### Note on the Python dependency

CPython on Windows ships without `curses`, and portlist is a curses program. The manifest declares `Python.Python.3.12` as a package dependency, and the description states that the `windows-curses` wheel is also needed, with `pipx install portlist-tui` given as the simpler route that handles both.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/436665)


Mr-hunt-007:

Mr-hunt-007:
Closing this — the blocker is in the package, not the manifest.

```
[Error] InvalidPortableFiletype: The file type of the referenced file is not allowed. (RelativeFilePath)
```

`s_AllowedPortableFiletypes` in `ManifestValidation.cpp` is a one-element array containing `.exe`, and what this project ships for Windows is the source plus a `portlist.cmd` shim — CPython on Windows has no `curses`, and this is a curses program, so the package needs Python on PATH regardless. A `.cmd` is not an `.exe`, so no revision of these manifests can pass.

Shipping a compiled launcher would fix it, but it would still require Python on PATH, so it would buy the listing and nothing else. Not a trade worth making here.

No action needed from a reviewer. Thanks for the fast, specific validation output — `InvalidPortableFiletype` named the exact rule, which made this quick to diagnose and easy to accept.

PR #437613, opened 2026-09-19
New version: SkyLLC.UFFS version 0.6.41

githubrobbi:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-pkgs/pull/437613)


githubrobbi:
Fixed in the latest commit — apologies for the noise.

Two problems in the submitted \`NestedInstallerFiles\`:

1. \`uffs-windows-x64/uffs-tui.exe\` was declared but is **not** in \`uffs-windows-x64.zip\` for this release. Our release workflow bundles that binary from a separate repository, whose published asset was renamed; our download silently missed it, and the bundling step is intentionally non-fatal, so the archive shipped without it.
2. As a consequence the alias \`uffs-tui\` was also declared a second time, against \`uffs-mft.exe\` — a duplicate path and a duplicate \`PortableCommandAlias\`.

Both entries are removed. The remaining six entries now match the archive exactly:

\`uffs.exe\`, \`uffsd.exe\`, \`uffsmcp.exe\`, \`uffs-mft.exe\`, \`uffs-broker.exe\`, \`uffs-update.exe\`

Verified against the published asset:

```
$ unzip -l uffs-windows-x64.zip | grep .exe
uffs-windows-x64/uffs-broker.exe
uffs-windows-x64/uffs-mft.exe
uffs-windows-x64/uffs-update.exe
uffs-windows-x64/uffs.exe
uffs-windows-x64/uffsd.exe
uffs-windows-x64/uffsmcp.exe
```

The installer URL and SHA256 are unchanged. The upstream asset-name fix is applied on our side so future submissions carry the full set again.

stephengillie (approved):

